PR #105, opened 2026-02-25
feat: build metadata tracking and excludeTests config shorthand

carlos-alm:
## Summary

- **Build metadata tracking (9.3):** New `build_meta` table (migration v7) stores engine name, version, codegraph version, and build timestamp. Incremental builds now warn when the engine or codegraph version has changed since the last build, suggesting `--no-incremental`.
- **Build metadata in `info` command (9.4):** `codegraph info` displays build metadata from the DB and warns on version/engine mismatches.
- **Top-level `excludeTests` config shorthand (9.2):** `{ "excludeTests": true }` at the top level of `.codegraphrc.json` is now hoisted into `query.excludeTests`. Nested form takes precedence.

## Test plan

- [x] 3 new unit tests for `excludeTests` hoisting (top-level true, nested precedence, top-level false)
- [x] 4 new unit tests for `build_meta` helpers (table creation, null on missing table, read/write, upsert)
- [x] All 249 unit tests pass
- [x] Lint clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,9 +532,14 @@ Create a `.codegraphrc.json` in your project root to customize behavior:
   },
   "build": {
     "incremental": true
+  },
+  "query": {
+    "excludeTests": true
   }
 }
 ```
+
+> **Tip:** `excludeTests` can also be set at the top level as a shorthand — `{ "excludeTests": true }` is equivalent to nesting it under `query`. If both are present, the nested `query.excludeTests` takes precedence.
 
 ### LLM credentials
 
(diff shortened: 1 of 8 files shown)


carlos-alm:
Addressed the review feedback in 8e08725:

**`setBuildMeta` failure logging (line 958):** Upgraded from `debug` to `warn` — metadata persistence failures may indicate disk issues and should be visible to users, not hidden behind debug-level logging.

All 565 tests pass.

@greptileai

PR #136, opened 2026-02-26
fix: sanitize threshold values in complexity SQL queries

carlos-alm:
## Summary
- Coerce complexity threshold `warn` values through `Number()` and guard with `Number.isNaN()` before interpolating into SQL `HAVING` clauses
- Prevents malformed queries when non-numeric values are provided in `.codegraphrc.json` config

## Test plan
- [ ] Run `node src/cli.js complexity` with default thresholds — verify normal behavior
- [ ] Set a non-numeric threshold in config (e.g. `"warn": "abc"`) — verify no SQL error


carlos-alm:
@greptileai

PR #173, opened 2026-02-28
fix(ci): use commit-count versioning for dev builds and benchmarks

carlos-alm:
## Summary

- Dev builds always produced `2.4.1-dev.<sha>` because patch was hardcoded to `PATCH + 1` from package.json. Now uses `git rev-list` to count commits since the release tag, giving monotonically increasing versions (`2.4.1`, `2.4.5`, `2.4.15`, etc.)
- Benchmark scripts now use `getBenchmarkVersion()` (shared `scripts/bench-version.js`) which uses `git describe` to detect dev vs release state, preventing dev/dogfood benchmark runs from overwriting release data in historical reports
- Added `fetch-depth: 0` to the `compute-version` checkout so git history and tags are available

## Changes

| File | Change |
|------|--------|
| `.github/workflows/publish.yml` | Replace `PATCH+1` with `git rev-list` commit counting, add `fetch-depth: 0` |
| `scripts/bench-version.js` | New shared helper — `getBenchmarkVersion(pkgVersion, cwd)` |
| `scripts/benchmark.js` | Use `benchVersion` instead of `pkg.version` |
| `scripts/incremental-benchmark.js` | Same |
| `scripts/query-benchmark.js` | Same |
| `scripts/embedding-benchmark.js` | Same |
| `scripts/token-benchmark.js` | Same |

## Test plan

- [ ] Verify `node -e "import { getBenchmarkVersion } from './scripts/bench-version.js'; console.log(getBenchmarkVersion('2.4.0', '.'))"` outputs `2.4.0-dev.N+hash` (not `2.4.0`)
- [ ] Verify on a tagged commit it returns the clean version
- [ ] Trigger a dev build via push to main and confirm version increments past `.1`

### DIFF
--- a/scripts/bench-version.js
+++ b/scripts/bench-version.js
@@ -46,11 +46,16 @@ export function getBenchmarkVersion(pkgVersion, cwd) {
 		/* git not available or no tags */
 	}
 
-	// Fallback: no git or no tags — match publish.yml's no-tags behavior (PATCH+1)
+	// Fallback: no git or no tags — match publish.yml's no-tags behavior (PATCH+1-dev.SHA)
 	const parts = pkgVersion.split('.');
 	if (parts.length === 3) {
 		const [major, minor, patch] = parts;
-		return `${major}.${minor}.${Number(patch) + 1}-dev`;
+		try {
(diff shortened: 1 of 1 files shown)


carlos-alm:
@greptileai

PR #222, opened 2026-03-02
fix: add CODEOWNERS parse cache and tighten email validation

carlos-alm:
## Summary
- Add module-level mtime-based cache to `parseCodeowners` — avoids re-reading/re-parsing the CODEOWNERS file on repeated calls with the same `rootDir`
- Replace permissive `p.includes('@')` email filter with `/^[^@\s]+@[^@\s]+$/` regex — rejects malformed tokens like `@`, `foo@`, `@bar` while keeping `startsWith('@')` for GitHub handles
- Add unit tests for both cache behavior (hit + invalidation) and malformed email rejection

## Test plan
- [x] `npx vitest run tests/unit/owners.test.js` — 25 tests pass (including 4 new)
- [x] `npx vitest run tests/integration/owners.test.js` — 12 tests pass
- [ ] Verify Greptile review feedback is addressed


carlos-alm:
Fixed the BACKLOG.md regression in b8df62b — restored from origin/main. The stale changes (date rollback + 6 reverted DONE markers for IDs 5, 15, 16, 18, 20, 31) came from an outdated local branch base. The PR diff now only touches src/owners.js and tests/unit/owners.test.js.

PR #311, opened 2026-03-04
fix: benchmark scripts handle missing WASM grammars gracefully

carlos-alm:
## Summary
- Add `isWasmAvailable()` to `parser.js` — checks if required WASM grammar files (JS, TS, TSX) exist on disk
- All 4 benchmark scripts (`benchmark.js`, `query-benchmark.js`, `incremental-benchmark.js`, `token-benchmark.js`) now check WASM availability before attempting WASM engine benchmarks
- When neither engine is available, scripts print a helpful error and exit cleanly instead of crashing
- Mirrors the existing `isNativeAvailable()` pattern for symmetric engine availability handling

## Test plan
- [x] `isWasmAvailable()` returns `false` when grammars directory is empty
- [x] `isWasmAvailable()` returns `true` when grammars are built
- [x] All 476 integration tests pass
- [x] Biome lint/format passes on all changed files
- [ ] Run `benchmark.js` without WASM grammars — should skip WASM and run native-only
- [ ] Run `benchmark.js` with WASM grammars — should run both engines as before

### DIFF
--- a/scripts/query-benchmark.js
+++ b/scripts/query-benchmark.js
@@ -167,12 +167,14 @@ if (!hasWasm && !hasNative) {
 	process.exit(1);
 }
 
+// Build with first available engine to select targets, then reuse for both
+let targets = null;
 let wasm = null;
 if (hasWasm) {
 	if (fs.existsSync(dbPath)) fs.unlinkSync(dbPath);
 	await buildGraph(root, { engine: 'wasm', incremental: false });
 
-	const targets = selectTargets();
+	targets = selectTargets();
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed review feedback in 1ac5d28:

**Greptile — targets selected separately for each engine (query-benchmark.js):**
Fixed. Targets are now selected once from whichever engine builds first, then reused for both. The native block only calls `selectTargets()` if WASM was unavailable (so no targets exist yet).

**Claude — missing test coverage for isWasmAvailable():**
Added `tests/unit/parser.test.js` with 6 tests covering: returns boolean, all grammars present, some missing, all missing, only checks required grammars (JS/TS/TSX), and verifies paths point to grammars/ directory.

@greptileai

PR #312, opened 2026-03-04
feat: enable --cfg and --dataflow by default on builds

carlos-alm:
## Summary

- CFG and dataflow analyses now run by default during `codegraph build`, eliminating the surprise of empty results when users try `cfg` or `dataflow` commands
- Users who want faster builds can opt out with `--no-cfg` / `--no-dataflow` (same Commander `--no-*` convention as `--no-incremental`)
- Added `dataflowMs` to the build return object (was tracked internally but not returned)
- Updated MCP tool descriptions and fallback warnings to reflect the new defaults

Closes #292

## Test plan

- [x] `codegraph build . --no-incremental` — CFG + dataflow run by default (timing visible in output)
- [x] `codegraph build . --no-cfg --no-dataflow --no-incremental` — both skipped cleanly
- [x] `codegraph cfg buildGraph` / `codegraph dataflow buildGraph` — return data without explicit flags
- [x] `npm test` — all tests pass (only pre-existing ruby/php WASM failures)

### DIFF
--- a/src/builder.js
+++ b/src/builder.js
@@ -536,13 +536,7 @@ export async function buildGraph(rootDir, opts = {}) {
       opts.dataflow !== false &&
       (() => {
         try {
-          return (
-            db
-              .prepare(
-                "SELECT COUNT(*) as c FROM edges WHERE kind IN ('flows_to','returns','mutates')",
-              )
-              .get().c === 0
-          );
+          return db.prepare('SELECT COUNT(*) as c FROM dataflow').get().c === 0;
         } catch {
           return true;
         }


carlos-alm:
Fixed in b20b7eb — updated both README locations to reflect default-on behavior:
- Feature table (line 211-212): changed from "opt-in via `build --cfg/--dataflow`" to "included by default, skip with `--no-cfg/--no-dataflow`"
- Note block (line 330): replaced with "Dataflow and CFG are included by default. Use `--no-dataflow` / `--no-cfg` for faster builds."

@greptileai

carlos-alm:
Fixed in aa5b989 — applied the same `opts.cfg !== false` / `opts.dataflow !== false` pattern to the incremental build checks (lines 527, 536). Also updated the comment from "optional analysis" to "default analyses" since they're no longer opt-in.

The previous README comments were already addressed in b20b7eb.

@greptileai

Issue #306, opened 2026-03-03
bug: native engine only stores 'call' AST nodes, missing 5 other kinds


carlos-alm:
## Found during dogfooding v3.0.0

**Severity:** Medium
**Command:** `codegraph ast`

## Reproduction

```bash
npx codegraph build <repo> --engine native --no-incremental
npx codegraph ast --kind new      # "No AST nodes found."
npx codegraph ast --kind string   # "No AST nodes found."
npx codegraph ast --kind regex    # "No AST nodes found."
npx codegraph ast --kind throw    # "No AST nodes found."
npx codegraph ast --kind await    # "No AST nodes found."
npx codegraph ast --kind call     # 10703+ nodes found
```

## Expected behavior

All 6 AST node kinds (call, new, string, regex, throw, await) should be stored when building with the native engine, matching the WASM engine behavior.

## Actual behavior

Only `call` nodes are stored (24716 in the codegraph repo). The other 5 kinds are empty. The WASM engine correctly stores all 6 kinds for JS/TS/TSX files because it preserves the tree-sitter parse tree and performs a post-build AST walk in `src/ast.js:buildAstNodes()`.

## Root cause

The native Rust engine (`crates/codegraph-core/`) extracts only `call_expression` nodes (see `src/extractors/javascript.rs`). It does not preserve the parse tree, so the JavaScript-side AST walk in `ast.js` cannot extract `new`, `string`, `regex`, `throw`, and `await` nodes.

The JS-side code in `ast.js` checks for `symbols._tree` and only walks it for JS/TS/TSX files, but the native engine never sets `_tree`.

## Suggested fix

Either:
1. Extend the native Rust extractors to also capture `new_expression`, `string`, `regex_pattern`, `throw_statement`, and `await_expression` AST nodes
2. Or document that `--engine wasm` is required for full AST node coverage

Option 1 is preferred for engine parity.

PR #317, opened 2026-03-04
perf: optimize build pipeline for incremental and full builds

carlos-alm:
## Summary
- **Batch node ID lookups** in insertion phase and AST extraction: replace per-node `getNodeId.get()` with bulk `SELECT ... WHERE file = ?` queries per file, eliminating O(defs + children) individual SELECTs
- **Skip AST/complexity for reverse-dep files** on incremental builds — their content didn't change, existing `ast_nodes`/`function_complexity` rows are still valid
- **Incremental structure rebuild**: scope directory node/edge/metric teardown to affected directories only instead of full `DELETE` + rebuild of all directories
- **Add `--no-ast` and `--no-complexity` build flags** for users who don't need these phases (mirrors existing `--cfg` opt-in pattern)
- **Schema-based version mismatch**: use `schema_version` (migration number) instead of `codegraph_version` for auto-promote decisions, so patch/minor bumps don't force full rebuilds
- **Include `astMs` in phase timing output** (was computed but never returned)

## Test plan
- [x] All 1369 tests pass (0 failures)
- [x] Lint clean (only pre-existing `rolesData` warning)
- [x] Self-build on codegraph repo succeeds (`node src/cli.js build .`)
- [ ] Run `node src/cli.js build . --no-incremental` and compare phase timings before/after
- [ ] Run `node src/cli.js build .` (incremental, no changes) — should be near-instant
- [ ] Touch one file, run incremental build — only that file goes through AST/complexity
- [ ] Test `--no-ast` and `--no-complexity` flags produce correct but reduced output
- [ ] Compare `node src/cli.js stats` before/after to verify node/edge counts match

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -156,9 +156,8 @@ export async function buildAstNodes(db, fileSymbols, _rootDir, _engineOpts) {
     return;
   }
 
-  const getNodeId = db.prepare(
-    'SELECT id FROM nodes WHERE name = ? AND kind = ? AND file = ? AND line = ?',
-  );
+  // Bulk-fetch all node IDs per file (replaces per-def getNodeId calls)
+  const bulkGetNodeIds = db.prepare('SELECT id, name, kind, line FROM nodes WHERE file = ?');
 
   const tx = db.transaction((rows) => {
     for (const r of rows) {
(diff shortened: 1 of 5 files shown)


carlos-alm:
Rebased on main — all conflicts resolved. Merge of native AST path (from #306/#314) with our batch nodeIdMap optimization in `src/ast.js`, and `--no-dataflow`/`--no-cfg` default-on flags (from #284) with our new `--no-ast`/`--no-complexity` opt-outs in `src/cli.js`.

@greptileai

PR #380, opened 2026-03-09
fix: defer heavy imports in MCP server for first-connect reliability

carlos-alm:
## Summary
- **Root cause**: `startMCPServer` was `await import()`-ing `queries.js` and `require()`-ing `better-sqlite3` *before* connecting the stdio transport. On cold start these imports are slow, so the client's `initialize` request arrives before the server is listening — the message is lost and connection fails.
- **Fix**: Moved heavy imports into lazy-loaded helpers (`getQueries()` / `getDatabase()`) that run on first tool call and are cached thereafter. The transport now connects immediately after the MCP SDK loads, so the server is ready to receive `initialize` right away.
- Reconnect always worked because Node.js caches the modules from the first attempt.

## Test plan
- [x] All 37 MCP unit tests pass
- [x] Smoke test: piped `initialize` JSON-RPC request gets immediate response
- [ ] Manual: verify `/mcp` connects on first try in Claude Code

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -342,8 +342,12 @@ export function astQueryData(pattern, customDbPath, opts = {}) {
     ORDER BY a.file, a.line
   `;
 
-  const rows = db.prepare(sql).all(...params);
-  db.close();
+  let rows;
+  try {
+    rows = db.prepare(sql).all(...params);
+  } finally {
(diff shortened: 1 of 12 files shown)


carlos-alm:
@greptileai

PR #382, opened 2026-03-09
fix: use CORE_SYMBOL_KINDS in flow matching

carlos-alm:
## Summary
- Flow's `findMatchingNodes` was limited to `FUNCTION_KINDS`, so `codegraph flow` couldn't trace from interfaces, types, or struct declarations
- Pass `CORE_SYMBOL_KINDS` via `opts.kinds` override so all 10 core symbol kinds are searchable
- Support `opts.kinds` array in `findMatchingNodes` as a general-purpose override alongside the existing `opts.kind` scalar

Split from #376.

## Test plan
- [ ] `codegraph flow <interface-name>` now finds and traces from interface declarations
- [ ] `codegraph flow <function-name>` still works as before (FUNCTION_KINDS default unchanged)
- [ ] Existing query/sequence commands unaffected (they don't pass `opts.kinds`)

### DIFF
--- a/.claude/hooks/check-commit.sh
+++ b/.claude/hooks/check-commit.sh
@@ -1,0 +1,177 @@
+#!/usr/bin/env bash
+# check-commit.sh — PreToolUse hook for Bash (git commit)
+# Combined cycle-detection (blocking) + signature-change warning (informational).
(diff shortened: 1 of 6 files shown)


carlos-alm:
@greptileai

PR #417, opened 2026-03-11
refactor: extract cachedStmt into reusable db utility

carlos-alm:
## Summary
- Extracts the WeakMap-based prepared statement caching pattern into a reusable `cachedStmt(cache, db, sql)` helper in `src/db/repository/cached-stmt.js`
- Applies `cachedStmt` to all 14 query functions in `edges.js` — each statement is compiled once per db instance and reused on subsequent calls
- Fixes the `getClassHierarchy` BFS bug: hoists the cache lookup **before** the `while` loop so the WeakMap check runs once per call, not per iteration
- Re-exports `cachedStmt` from the repository barrel (`index.js`) so other modules (e.g. `sequence.js`, `queries.js`) can adopt it incrementally

Addresses Greptile review feedback on #403: "The caching pattern could potentially be extracted into a reusable database utility."

## Test plan
- [x] Repository unit tests pass (20/20)
- [x] Integration tests pass (448/448, 4 pre-existing WASM worktree failures unrelated)
- [x] Biome lint clean

### DIFF
--- a/src/db/repository/cochange.js
+++ b/src/db/repository/cochange.js
@@ -1,3 +1,10 @@
+import { cachedStmt } from './cached-stmt.js';
+
+// ─── Statement caches (one prepared statement per db instance) ────────────
+const _hasCoChangesStmt = new WeakMap();
+const _getCoChangeMetaStmt = new WeakMap();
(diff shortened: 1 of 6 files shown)


carlos-alm:
@greptileai

PR #456, opened 2026-03-16
refactor: reorganize src/ into domain/, features/, presentation/ layers

carlos-alm:
## Summary
- Reorganizes flat `src/` into a domain-driven directory structure with three top-level layers:
  - **`src/domain/`** — core graph building, import resolution, queries, analysis, and search (embeddings)
  - **`src/features/`** — composable feature modules (audit, check, complexity, cfg, dataflow, etc.)
  - **`src/presentation/`** — CLI formatting and output rendering
  - **`src/db/`** — database layer (`db.js` → `db/index.js`)
- Updates all 199 import paths across source, tests, CLI commands, MCP tools, and shared modules
- Fixes 3 stale import paths discovered during validation (MCP server lazy-load, triage test, MCP test mocks)

## Test plan
- [x] All 1858 tests pass across 104 test files
- [x] `codegraph build` succeeds (5334 nodes, 11259 edges)
- [x] `codegraph check` — 0 cycles, 0 boundary violations, all 10 manifesto rules pass
- [x] No stale import paths referencing old flat `src/` locations
- [x] Dead export analysis shows no regressions from restructuring

PR #457, opened 2026-03-16
fix: prevent findDbPath from escaping git worktree boundary

carlos-alm:
## Summary

- Add `findRepoRoot()` to `src/db/connection.js` using `git rev-parse --show-toplevel`, which returns the correct root for both regular repos and git worktrees. Results are cached per-process.
- Use `findRepoRoot()` as a ceiling in `findDbPath()` so the walk-up stops at the git boundary — worktree sessions resolve to their own DB, not the parent repo's.
- Add unit tests for `findRepoRoot` (caching, non-git fallback) and `findDbPath` ceiling behavior (worktree simulation, non-git fallback).

## Context

When codegraph runs inside a git worktree (e.g. `.claude/worktrees/agent-xxx/`), `findDbPath()` walked up from `process.cwd()` and crossed the worktree boundary, finding the **main repo's** database. This made codegraph unusable in parallel worktree sessions — the exact workflow it's designed to support.

## Behavior change

When no existing `.codegraph/graph.db` is found during the walk-up, the default DB creation path is now the **git repo root** (from `findRepoRoot()`) instead of `process.cwd()`. For example, running `codegraph build` from `/repo/packages/foo/` with no pre-existing DB now creates `/repo/.codegraph/graph.db` instead of `/repo/packages/foo/.codegraph/graph.db`. When not inside a git repo, the previous behavior (`process.cwd()`) is preserved. This is intentional — the repo root is the correct default for most use cases.

## Test plan

- [x] `npm test` — all existing tests pass
- [ ] In a worktree: `cd .claude/worktrees/xxx && npx codegraph build . && npx codegraph stats` — uses worktree DB
- [ ] In main repo: `codegraph stats` — still works as before
- [ ] MCP tools from a worktree session resolve to the correct DB

## Compatibility

Only 2 files changed (`src/db/connection.js`, `tests/unit/db.test.js`). No overlap with #456 on `connection.js`. Minor merge conflict possible in `tests/unit/db.test.js` import lines — trivial to resolve.

### DIFF
--- a/docs/roadmap/ROADMAP.md
+++ b/docs/roadmap/ROADMAP.md
@@ -2,7 +2,7 @@
 
 > **Current version:** 3.1.4 | **Status:** Active development | **Updated:** March 2026
 
-Codegraph is a strong local-first code graph CLI. This roadmap describes planned improvements across ten phases -- closing gaps with commercial code intelligence platforms while preserving codegraph's core strengths: fully local, open source, zero cloud dependency by default.
+Codegraph is a strong local-first code graph CLI. This roadmap describes planned improvements across eleven phases -- closing gaps with commercial code intelligence platforms while preserving codegraph's core strengths: fully local, open source, zero cloud dependency by default.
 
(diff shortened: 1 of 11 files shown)


carlos-alm:
@greptileai

PR #479, opened 2026-03-17
fix: use semver-compliant dev version numbering

carlos-alm:
## Summary
Change dev release scheme from `MAJOR.MINOR.(PATCH+COMMITS)-dev.SHA` to `MAJOR.MINOR.(PATCH+1)-dev.COMMITS`.

The old scheme inflated the patch number — e.g. 45 commits after `v3.1.4` produced `3.1.49-dev.abc1234`, which collides with real semver space and is confusing when the next actual release is just `3.1.5`.

The new scheme produces `3.1.5-dev.45`:
- npm never picks up prerelease versions on plain `npm install`
- Sorts correctly: `3.1.4 < 3.1.5-dev.1 < 3.1.5-dev.45 < 3.1.5`
- The target release is obvious from the version string
- Matches the convention used by TypeScript, ESLint, Vite

**Files changed:**
- `.github/workflows/publish.yml` — `compute-version` job
- `scripts/bench-version.js` — benchmark version computation (mirrors publish.yml)
- `.claude/skills/dogfood/SKILL.md` — updated example version string

## Test plan
- [ ] After merging, next push to main produces version like `3.1.6-dev.1` (not `3.1.6-dev.SHA`)
- [ ] `scripts/bench-version.js` produces matching format when run locally
- [ ] Existing dev GitHub releases are unaffected (old tags remain)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -80,8 +80,11 @@ jobs:
               COMMITS=1
               IFS='.' read -r MAJOR MINOR PATCH <<< "$CURRENT"
             fi
-            NEXT_PATCH=$((PATCH + 1))
-            VERSION="${MAJOR}.${MINOR}.${NEXT_PATCH}-dev.${COMMITS}"
+            if [ "$COMMITS" -eq 0 ]; then
+              VERSION="${MAJOR}.${MINOR}.${PATCH}"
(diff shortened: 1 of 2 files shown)


carlos-alm:
@greptileai

PR #495, opened 2026-03-17
fix: include constant nodes in edge building (#487)

### LINKED ISSUE #487: Edge building excludes constant nodes from dependency graph
## Bug

`codegraph where BUILTIN_RECEIVERS -T --json` returns empty results, and `fn-impact` reports 0 dependents — yet grep confirms 3 active consumers in core source files (`incremental.js`, `build-edges.js`).

## Root Cause

The JS extractor **does** correctly extract `export const` declarations as `constant`-kind nodes and they **are** inserted into the `nodes` table. However, the edge-building stage in `src/domain/graph/builder/stages/build-edges.js` (line 31) excludes them from the node lookup:

```sql
SELECT id, name, kind, file, line FROM nodes
WHERE kind IN ('function','method','class','interface','struct','type','module','enum','trait')
```

`'constant'` is not in this list. As a result:
- Constants exist in the DB but are never loaded into `nodesByName` / `nodesByNameAndFile`
- No import/dependency edges are created for them
- They are invisible to `where`, `fn-impact`, `query`, `roles`, and all other graph queries

## Impact

- `codegraph roles --role dead` misclassifies constants as dead (0 fan-in because no edges exist)
- `codegraph where` can't find them at all
- `codegraph fn-impact` reports 0 dependents for actively-used constants
- Any downstream tooling (e.g. automated dead-code cleanup) that trusts codegraph data will incorrectly target these symbols for removal

## Affected Symbols (examples from this repo)

- `BUILTIN_RECEIVERS` — `export const` in `src/domain/graph/builder/helpers.js`, used in 3 source files
- `BATCH_CHUNK` — `export const` in same file, actively imported
- `MAX_COL_WIDTH` — `export const`, used in presentation layer

## Fix

Add `'constant'` to the `WHERE kind IN (...)` clause in `build-edges.js` line 31, so constants participate in edge resolution like any other symbol.

carlos-alm:
## Summary

- Add `'constant'` to the `WHERE kind IN (...)` SQL clause in `build-edges.js` so exported constants participate in edge resolution and become visible to all graph queries (`where`, `fn-impact`, `query`, `roles`, etc.)
- Also adds missing `'record'` kind to `build-edges.js` to match `CORE_SYMBOL_KINDS`
- Updates 4 downstream query files (`context.js`, `watcher.js`, `export.js`, `graph-enrichment.js`) to include `'constant'` in their kind filters

Closes #487

## Test plan

- [x] All integration tests pass (504/504)
- [x] All graph tests pass (139/139)
- [ ] After rebuild, `codegraph where BUILTIN_RECEIVERS -T` returns results
- [ ] `codegraph fn-impact BUILTIN_RECEIVERS -T` reports correct dependents
- [ ] `codegraph roles --role dead -T` no longer misclassifies active constants


carlos-alm:
Fixed in 8266b33 — addressed the Greptile feedback about `whereSymbolImpl` and `findMatchingNodes` not finding constants:

1. **`whereSymbolImpl`**: Replaced `ALL_SYMBOL_KINDS` (core 10, excludes `constant`) with `EVERY_SYMBOL_KIND` (includes `constant`, `parameter`, `property`). `codegraph where <constant>` now returns results.
2. **`findMatchingNodes`**: Added `'constant'` to the `FUNCTION_KINDS` default array. This fixes `codegraph fn-impact`, `codegraph context`, and `codegraph deps` for constant symbols.

All 1862 tests pass, lint clean.

@greptileai

PR #497, opened 2026-03-17
fix: classify test-only-called symbols as 'test-only' instead of 'dead'

### LINKED ISSUE #488: Public API symbols classified as dead when all callers are in test files
## Bug

When using `codegraph roles --role dead -T`, symbols whose only callers are in test files get classified as `dead` because `-T` (exclude tests) removes all their incoming edges, resulting in `fanIn === 0`.

The `classifyRoles` function in `src/graph/classifiers/roles.js` (line 46-49) uses this logic:

```js
if (node.fanIn === 0 && !node.isExported) {
  role = 'dead';
} else if (node.fanIn === 0 && node.isExported) {
  role = 'entry';
}
```

So an exported symbol with 0 fan-in (after test exclusion) is classified as `entry`, not `dead`. This part is correct.

However, **non-exported symbols that are only called from tests** get classified as `dead` when `-T` is used — even if they serve an important internal role. For example, a helper function used extensively in test infrastructure but not in production code would be flagged as dead.

## Expected Behavior

The `-T` flag should exclude test files from **output/display**, but the fan-in calculation for role classification should still account for test-file edges (or at minimum, the `dead` role should note that the symbol has test-only callers rather than zero callers).

## Current Behavior

`-T` filters test files from the graph edges before classification, so test-only-called symbols appear to have `fanIn === 0` and are classified as `dead`.

## Suggested Approach

Consider one of:
1. Compute fan-in on the full graph, then apply `-T` filtering only to the output
2. Add a distinct role like `test-only` for symbols that have callers exclusively in test files
3. Annotate `dead` results with a flag indicating whether test-file callers exist

carlos-alm:
## Summary

Closes #488

- When `-T` is active, symbols whose only callers are in test files are now classified as `test-only` instead of `dead`
- Added `test-only` to `VALID_ROLES` so it can be queried with `--role test-only`
- `classifyRoles` checks a new `testOnlyFanIn` property: if `fanIn === 0` and `!isExported` but `testOnlyFanIn > 0`, assigns `test-only` role
- `rolesData` in the analysis layer identifies test-only-called symbols by querying all `calls` edges and grouping callers by target

## Test plan

- [ ] `codegraph roles --role dead -T` no longer includes symbols only called from tests
- [ ] `codegraph roles --role test-only -T` shows those symbols instead
- [ ] Unit tests for classifier with `testOnlyFanIn` property
- [ ] Integration tests with fixture symbol called only from test file


carlos-alm:
@greptileai

carlos-alm:
@greptileai

PR #505, opened 2026-03-18
feat: receiver type tracking with graded confidence (4.2)

carlos-alm:
## Summary

Implements roadmap item **4.2 — Receiver Type Tracking for Method Dispatch**.

- **Graded confidence**: Upgrades `typeMap` value from plain string to `{type, confidence}` across all 8 language extractors. Confidence: `1.0` constructor, `0.9` annotation/parameter, `0.7` factory method
- **Factory pattern extraction**: JS/TS `Foo.create()`, Go `NewFoo()`/`&Struct{}`/`Struct{}`, Python `Foo()`/`Foo.create()` — previously missing, now tracked at 0.7 confidence
- **Receiver edges use type confidence**: Instead of hardcoded 0.9/0.7, receiver edge confidence now reflects the actual type source precision
- **`setIfHigher` priority**: When the same variable has multiple type sources (e.g., `const x: Base = new Derived()`), highest confidence wins
- **Backwards-compatible**: `typeof entry === 'string'` guards handle mixed old/new formats during native binary transitions

## Changed files

| File | Change |
|------|--------|
| `src/extractors/*.js` (8 files) | `typeMap.set(name, typeName)` → `typeMap.set(name, {type, confidence})`, factory patterns added |
| `src/domain/graph/builder/stages/build-edges.js` | Extract `.type`/`.confidence` from entries, pass confidence to native path |
| `src/domain/graph/builder/incremental.js` | Same entry format update |
| `tests/parsers/javascript.test.js` | Updated assertions + new factory/confidence tests |
| `tests/parsers/java.test.js` | Updated assertions |
| `tests/integration/build.test.js` | Receiver edge confidence 0.9 → 1.0 (constructor) |
| `CLAUDE.md` | Added rule to sync with origin/main before starting work |
| `docs/roadmap/ROADMAP.md` | Marked 4.2 complete |

## Test plan

- [x] 807 tests pass (50 test files), 0 failures
- [x] New tests: factory method extraction, graded confidence assertions
- [x] Lint clean (biome)
- [x] Build parity test passes (WASM vs native)
- [x] Query-walk parity test passes

### DIFF
--- a/docs/roadmap/ROADMAP.md
+++ b/docs/roadmap/ROADMAP.md
@@ -995,22 +995,15 @@ src/domain/
 
 ## Phase 4 -- Resolution Accuracy
 
-> **Status:** Planned
+> **Status:** In Progress
 
 **Goal:** Close the most impactful gaps in call graph accuracy before investing in type safety or native acceleration. The entire value proposition — blast radius, impact analysis, dependency chains — rests on the call graph. These targeted improvements make the graph trustworthy.
 
 **Why before TypeScript:** These fixes operate on the existing JS codebase and produce measurable accuracy gains immediately. TypeScript types will further improve resolution later, but receiver tracking, dead role fixes, and precision benchmarks don't require types to implement.
 
-### 4.1 -- Fix "Dead" Role Sub-categories
+### ~~4.1 -- Fix "Dead" Role Sub-categories~~ ✅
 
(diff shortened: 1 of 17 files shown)


carlos-alm:
Addressed all review feedback in 3f49e6b:

- **Factory heuristic on builtins (P1):** Added `BUILTIN_GLOBALS` denylist to JS/TS and `BUILTIN_GLOBALS_PY` to Python, preventing false positives from `Math.random()`, `JSON.parse()`, `Promise.resolve()`, etc. Added test coverage.
- **Redundant uppercase guard (P2):** Simplified to single `toLowerCase()` check across JS and Python extractors.
- **Annotation-vs-constructor priority (P1):** Added deliberate comment documenting the semantic reversal and its rationale.
- **Go multi-variable short declarations (P1, outside diff):** Rewrote to iterate all left-hand identifiers in parallel with their corresponding RHS expressions, so `x, y := A{}, B{}` now types both variables.

All 1913 tests pass, lint clean.

@greptileai

PR #510, opened 2026-03-19
fix: reword misleading 'stale' warning in info command

carlos-alm:
## Summary
- The `codegraph info` command showed `(binary reports X.Y.Z — stale)` when the npm package version differed from the binary's compiled-in version, even though the engine works correctly
- Reworded to `(binary built as X.Y.Z, cosmetic only)` to clarify this is a harmless version-string mismatch, not a broken engine

## Test plan
- [ ] Run `codegraph info` with a native engine where package and binary versions differ — confirm new wording appears
- [ ] Run `codegraph info` with matching versions — confirm no parenthetical note


carlos-alm:
@greptileai

PR #522, opened 2026-03-19
feat: Go structural interface matching and C# implements disambiguation (4.3)

carlos-alm:
## Summary

Completes roadmap item **4.3 — Interface and Trait Implementation Tracking** by filling the two remaining extractor-level gaps. The query layer, CLI commands (`implementations`, `interfaces`), MCP tools, impact analysis integration (`bfsTransitiveCallers` with `includeImplementors`), presentation formatting, and integration tests were already in place.

**Go extractor — structural interface matching:**
- Post-walk pass compares struct method sets against interface method sets within the same file
- Pushes `{ name, implements, line }` entries to the `classes` array
- Empty interfaces are excluded (they'd match everything)

**C# extractor — implements disambiguation:**
- Fixed `extractCSharpBaseTypes`: `childForFieldName('bases')` returns null in tree-sitter-c-sharp; added `findChild(node, 'base_list')` fallback
- Added `reclassifyCSharpImplements` post-walk pass that reclassifies `extends` entries as `implements` when the target is a known same-file interface definition

**Type definitions:**
- Added `ImplementationsResult`, `InterfacesResult` to `src/types.ts`
- Added `findImplementors`, `findInterfaces` to `Repository` interface
- Added `implementations`, `interfaces` to `McpDefaults`

## Test plan

- [x] 4 new Go parser tests: structural matching, missing methods, multiple interfaces, empty interfaces
- [x] 3 new C# parser tests: same-file interface reclassification, cross-file extends preserved, mixed extends+implements
- [x] C# engine parity test skipped (known native gap: native engine doesn't extract `base_list`)
- [x] All 13 integration tests pass (`implementationsData`, `interfacesData`, `contextData`, `fnImpactData`)
- [x] Full test suite: 2006 passed, 7 skipped, 0 failed

### DIFF
--- a/.claude/skills/dogfood/SKILL.md
+++ b/.claude/skills/dogfood/SKILL.md
@@ -69,7 +69,17 @@ Your goal is to install the published package, exercise every feature, compare e
    ```
    This should report `engine: native`. If it falls back to `wasm`, record why.
 5. Record: platform, OS version, Node version, native binary package name + version, engine reported by `info`.
-6. **Do NOT rebuild the graph yet.** The first phase tests commands against the codegraph source repo without a pre-existing graph.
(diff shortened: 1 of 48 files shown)


carlos-alm:
@greptileai

PR #600, opened 2026-03-25
fix: normalize paths in native resolver for .js → .ts remap

### LINKED ISSUE #592: bug: native resolver fails .js → .ts extension remap due to unnormalized path
## Found during dogfooding v3.3.2-dev.39

**Severity:** Critical
**Command:** All commands — import resolution is broken for TypeScript codebases

## Reproduction

```bash
# From the codegraph repo root
node -e "
const { loadNative } = require('./dist/infrastructure/native.js');
const native = loadNative();
console.log(native.resolveImport('src/cli/commands/build.ts', '../../domain/graph/builder.js', '.'));
"
# Output: src/cli/commands/../../domain/graph/builder.js
# Expected: src/domain/graph/builder.ts
```

After building a TypeScript codebase where imports use `.js` extensions (ESM convention: `import { foo } from './bar.js'` where `bar.ts` is the actual file), only **21 import edges** are resolved out of hundreds of imports.

## Expected behavior

`resolveImport` should return `src/domain/graph/builder.ts` — the `.js` → `.ts` remap logic (line 117-133 in `import_resolution.rs`) should trigger.

## Actual behavior

Returns `src/cli/commands/../../domain/graph/builder.js` — the path is not normalized, causing `strip_prefix(root_dir)` to fail silently.

## Root cause

In `crates/codegraph-core/src/import_resolution.rs`, line 114:

```rust
let resolved: PathBuf = dir.join(import_source).components().collect();
```

`PathBuf::components().collect()` does NOT resolve `..` components in Rust — it only normalizes `.` and separators. So when `import_source` is `../../domain/graph/builder.js`, the resolved path retains the `../..` components.

The `.js` → `.ts` remap at lines 117-133 then fails because:
1. `file_exists(&ts_candidate, None)` may succeed (OS resolves `..` in `exists()`), BUT
2. `Path::new(&ts_candidate).strip_prefix(root)` fails because the unnormalized path `src/cli/commands/../../domain/graph/builder.ts` cannot be prefix-stripped by `.` or an absolute root path.
3. The function falls through and returns the original `.js` path.

The same bug affects ALL extension probing logic that follows (lines 134+).

### Rust fix (primary)
Replace line 114 with proper path normalization:

```rust
fn clean_path(p: &Path) -> PathBuf {
    let mut result = PathBuf::new();
    for c in p.components() {
        match c {
            std::path::Component::ParentDir => { result.pop(); }
            std::path::Component::CurDir => {}
            _ => result.push(c),
        }
    }
    result
}

let resolved = clean_path(&dir.join(import_source));
```

### JS workaround (immediate)
In `resolveImportPath` (resolve.ts), after getting the native result, check if it ends with `.js` and apply the JS `.js` → `.ts` fallback:

```typescript
if (native) {
    const result = native.resolveImport(fromFile, importSource, rootDir, ...);
    const normalized = normalizePath(path.normalize(result));
    if (normalized.endsWith('.js')) {
        const tsPath = path.resolve(rootDir, normalized.replace(/\.js$/, '.ts'));
        if (fs.existsSync(tsPath)) return normalizePath(path.relative(rootDir, tsPath));
        const tsxPath = path.resolve(rootDir, normalized.replace(/\.js$/, '.tsx'));
        if (fs.existsSync(tsxPath)) return normalizePath(path.relative(rootDir, tsxPath));
    }
    return normalized;
}
```

This also affects `resolve_imports_batch` when called without `known_files`.

carlos-alm:
## Summary
- Adds `clean_path()` helper in `import_resolution.rs` to properly normalize `..` and `.` path components by popping parent directories instead of collecting them verbatim
- Fixes `strip_prefix` failures that caused .js → .ts extension remapping to silently fail when import paths contained `../` segments (e.g. `src/cli/commands/../../domain/graph/builder.js`)
- Updates `normalize_path()` to use the new `clean_path()` internally, fixing the same issue for all path normalization in the native resolver
- JS-side defense-in-depth (`path.normalize` + `remapJsToTs`) was already in place; this fixes the root cause in the Rust native resolver

Fixes #592

## Test plan
- [ ] Build with native engine on a TS codebase with relative imports using `../`
- [ ] Verify import edge count matches WASM engine
- [ ] Run `codegraph build . --engine native` on this repo and check edge counts

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,7 +58,7 @@ npm install                      # Install dependencies
 npm test                         # Run all tests (vitest)
 npm run test:watch               # Watch mode
 npm run test:coverage            # Coverage report
-npx vitest run tests/parsers/javascript.test.js   # Single test file
+npx vitest run tests/parsers/javascript.test.ts   # Single test file
 npx vitest run -t "finds cycles"                  # Single test by name
 npm run build:wasm               # Rebuild WASM grammars from devDeps (built automatically on npm install)
 ```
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed the coverage gap for the exact #592 regression scenario (.js → .ts remap across `../` traversal):

1. **Rust unit test**: Added `js_to_ts_remap_across_parent_traversal` in `import_resolution.rs` — creates a temp directory with `root.ts`, then resolves `../root.js` from a subdirectory and asserts it resolves to `root.ts`. This directly exercises the `.js → .ts` remap branch (lines 141–155) after `clean_path` normalizes the `../` segment.

2. **Parity test**: Added a skipped `resolves .js → .ts remap across parent traversal (#592)` test in `parity.test.ts` that creates a `.ts` file and imports it with a `.js` extension across `../`. Skipped until the native binary is published with the `clean_path` fix.

PR #612, opened 2026-03-26
fix(bench): repair benchmark workflow broken by TypeScript migration

carlos-alm:
## Summary

The v3.4.0 benchmark workflow ([run #23576682993](https://github.com/optave/codegraph/actions/runs/23576682993)) failed across all 4 jobs after the TypeScript migration (PR #580). Three issues:

- **`node src/cli.js build .`** — the embedding benchmark's graph-build step hardcoded the old `.js` path. Fixed to use `--strip-types` + `ts-resolve-loader` with `src/cli.ts`
- **`srcImport()` resolved `.js` paths that only exist as `.ts`** — the helper now checks for `.ts` on disk first, falling back to `.js` for backwards compatibility
- **`2>/dev/null` on benchmark commands** — silenced all errors, making the failures invisible. Removed so CI logs show actual errors
- **Stale HuggingFace cache key** — referenced `src/embeddings/**` (pre-migration path), updated to `src/domain/search/**`

## Test plan

- [ ] Merge to main, then trigger benchmark workflow via `workflow_dispatch` with `version: dev` to verify all 4 jobs pass
- [ ] Alternatively, re-run the v3.4.0 benchmark after merge by dispatching with `version: 3.4.0`


carlos-alm:
**CI failure analysis**: The test failure in `tests/engines/ast-parity.test.ts:216` ("JS: native calls match legacy calls field count" — expected 3 >= 4) is a pre-existing native engine parity bug, not caused by this PR. This PR only modifies `scripts/lib/bench-config.ts` and `.github/workflows/benchmark.yml`. The same test passed on PRs #606 and #607 but may have become sensitive to the native addon version shipped with v3.4.0. This needs a separate fix in the native engine (`crates/codegraph-core/`).

carlos-alm:
**CI failure analysis**: The test failure in `tests/engines/ast-parity.test.ts:216` ("JS: native calls match legacy calls field count" — expected 3 >= 4) is a pre-existing native engine parity bug, not caused by this PR. This PR only modifies `scripts/lib/bench-config.ts` and `.github/workflows/benchmark.yml`. The same test passed on PRs #606 and #607 but may have become sensitive to the native addon version shipped with v3.4.0. This needs a separate fix in the native engine (`crates/codegraph-core/`).

carlos-alm:
@greptileai

PR #628, opened 2026-03-26
fix(skills): prevent /review from spamming @greptileai when already approved

carlos-alm:
## Summary
- Fixes the `/review` skill blindly re-triggering `@greptileai` even when Greptile already approved (thumbsup + no inline comments), as seen on PR #621
- Replaces the single emoji-on-reply check with a proper decision matrix that checks the trigger comment's thumbsup reaction, inline comment count, and elapsed time
- Adds a 15-minute wait rule: if Greptile hasn't responded yet, wait instead of posting duplicate triggers

## Test plan
- [ ] Run `/review` on a PR where Greptile already gave thumbsup + no comments — verify it skips re-triggering
- [ ] Run `/review` on a PR with outstanding Greptile inline comments — verify it addresses them and re-triggers
- [ ] Run `/review` on a PR where `@greptileai` was just posted (<15 min ago) — verify it waits instead of posting again


carlos-alm:
Fixed the remaining concern about the first-ever trigger case. When `trigger_count` is 0, the script now posts `@greptileai` directly and proceeds to the wait-and-recheck loop (step 2h), bypassing Check 2 entirely. This ensures the first-ever trigger always fires on a fresh PR.

PR #651, opened 2026-03-27
perf(ast): bulk-insert AST nodes via native Rust/rusqlite

carlos-alm:
## Summary
- Adds `bulk_insert_ast_nodes` napi function in new `crates/codegraph-core/src/ast_db.rs` — opens SQLite directly from Rust via rusqlite, pre-fetches parent node definitions, and inserts all AST rows in a single transaction
- JS-side `buildAstNodes` tries the native fast path first (when all files have native `astNodes` arrays), falling back to existing JS loop for WASM/mixed-engine scenarios
- Adds `bulkInsertAstNodes` to `NativeAddon` interface in `src/types.ts`

## Motivation
AST node insertion was the bottleneck in the analysis phase (~393ms), iterating per-row through JS with individual `insertStmt.run()` calls plus `bulkNodeIdsByFile()` per file. This moves the entire operation to a single Rust transaction, targeting astMs < 50ms on native full builds.

## Test plan
- [x] All 2129 existing tests pass (no behavior change — same data inserted)
- [ ] CI builds the native addon and runs the full test suite
- [ ] Verify `astMs` timing in build output drops significantly on native engine builds

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -146,8 +146,13 @@ Read `README.md` and check if any new user-facing features from this release nee
 2. **MCP tools table** — if new MCP tools were added, add them to the AI integration section
 3. **Feature descriptions** — if a major new capability was added (new analysis type, new output format, etc.), add it to the relevant section
 4. **Roadmap section** — if a phase status changed, update the roadmap summary at the bottom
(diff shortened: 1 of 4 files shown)


carlos-alm:
@greptileai

PR #654, opened 2026-03-27
perf(insert-nodes): native Rust/rusqlite pipeline for node insertion

carlos-alm:
## Summary
- Add `bulk_insert_nodes` napi function in `crates/codegraph-core/src/insert_nodes.rs` that writes nodes, children, containment edges, exports, and file hashes directly to SQLite via rusqlite in a single transaction
- Wire it into the JS insert-nodes stage: when native engine is available, marshal `allSymbols` to lean batch format and delegate all DB writes to Rust — eliminating JS↔better-sqlite3 FFI overhead
- JS fallback preserved unchanged for WASM engine or native failure

**Target:** `insertMs` < 50ms on native full builds (current: ~206ms)

## Test plan
- [x] All 560 integration tests pass (WASM fallback exercised locally)
- [ ] CI builds native addon and exercises the Rust insert path on Linux/macOS
- [ ] Benchmark native full build to verify insertMs < 50ms target

### DIFF
--- a/crates/codegraph-core/src/insert_nodes.rs
+++ b/crates/codegraph-core/src/insert_nodes.rs
@@ -267,7 +267,7 @@ fn do_insert(
     // ── Insert all edges ─────────────────────────────────────────────
     {
         let mut stmt = tx.prepare_cached(
-            "INSERT INTO edges (source_id, target_id, kind, confidence, dynamic) \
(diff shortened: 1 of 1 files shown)


carlos-alm:
@greptileai

PR #672, opened 2026-03-29
feat(config): restore path aliases with TS 6.x-compatible subpath imports

### LINKED ISSUE #668: follow-up: restore path aliases with TS 6.x-compatible configuration
Deferred from PR #667 review.

**Original reviewer comment:** Claude automated review on PR #667 suggested restoring path aliases (`#shared/*`, `#infrastructure/*`, `#db/*`, etc.) using TS 6.x-compatible configuration.

**Context:** PR #667 removed the `baseUrl` and `paths` entries from `tsconfig.json` because they were deprecated in TypeScript 6.x (TS5101 error). While the aliases were never used in imports (all 849+ imports use relative paths), restoring them with modern TS 6.x configuration and gradually migrating the 300+ relative `../../` imports to path aliases would improve code readability.

**What needs to be done:**
1. Research the TS 6.x replacement for `baseUrl`/`paths` (likely Node.js subpath imports via `package.json` `imports` field, or `tsconfig.json` `paths` without `baseUrl`)
2. Configure the aliases in a TS 6.x-compatible way
3. Optionally migrate existing relative imports to use the aliases

carlos-alm:
## Summary
- Replace deprecated `baseUrl`/`paths` with Node.js subpath imports (`package.json` `imports` field) + `customConditions` in `tsconfig.json`
- Uses `@codegraph/source` condition for compile-time resolution to `src/`, `default` condition for runtime resolution to `dist/`
- Migrates 3 deeply nested files as proof-of-concept; remaining 300+ imports can be migrated incrementally

## Details

**How it works:**
```
Code:       import { logger } from '#infrastructure/logger.js';
TypeScript: resolves → src/infrastructure/logger.ts  (via @codegraph/source condition)
Node.js:    resolves → dist/infrastructure/logger.js  (via default condition)
```

**Changes:**
- `package.json` — `imports` field with conditional mappings for all 11 `src/` subdirectories + `#types`
- `tsconfig.json` — `customConditions: ["@codegraph/source"]`
- `vitest.config.ts` — `resolve.conditions: ['@codegraph/source']` for Vite resolver
- `scripts/verify-imports.ts` — resolves `#`-prefixed specifiers through the imports map
- 3 proof-of-concept files migrated (`build-structure.ts`, `build-edges.ts`, `collect-files.ts`)

## Test plan
- [x] `npx tsc --noEmit` — zero errors
- [x] `npm run build` — `#` specifiers preserved in compiled JS
- [x] `node dist/cli.js --help` — runtime resolves via `default` condition
- [x] `npm run verify-imports` — 281 files pass
- [x] `npm test` — 2128 pass (1 pre-existing failure unrelated to this PR)
- [x] `biome check` — clean on migrated files

Closes #668

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,20 @@
   "description": "Local code graph CLI — parse codebases with tree-sitter, build dependency graphs, query them",
   "type": "module",
   "main": "dist/index.js",
+  "imports": {
+    "#shared/*": { "@codegraph/source": "./src/shared/*", "default": "./dist/shared/*" },
+    "#infrastructure/*": { "@codegraph/source": "./src/infrastructure/*", "default": "./dist/infrastructure/*" },
+    "#db/*": { "@codegraph/source": "./src/db/*", "default": "./dist/db/*" },
+    "#domain/*": { "@codegraph/source": "./src/domain/*", "default": "./dist/domain/*" },
+    "#features/*": { "@codegraph/source": "./src/features/*", "default": "./dist/features/*" },
+    "#presentation/*": { "@codegraph/source": "./src/presentation/*", "default": "./dist/presentation/*" },
+    "#graph/*": { "@codegraph/source": "./src/graph/*", "default": "./dist/graph/*" },
+    "#mcp/*": { "@codegraph/source": "./src/mcp/*", "default": "./dist/mcp/*" },
+    "#ast-analysis/*": { "@codegraph/source": "./src/ast-analysis/*", "default": "./dist/ast-analysis/*" },
+    "#extractors/*": { "@codegraph/source": "./src/extractors/*", "default": "./dist/extractors/*" },
+    "#cli/*": { "@codegraph/source": "./src/cli/*", "default": "./dist/cli/*" },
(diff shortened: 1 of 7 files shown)


carlos-alm:
Addressed all Greptile findings:

1. **BLOB/cell-error in `row_to_json`**: Added doc comment to `row_to_json` explicitly documenting the contract -- BLOB columns map to null because the codegraph schema has no BLOB columns, and cell-level read errors map to null to avoid partial-row failures.

2. **Unsafe params cast in `query-builder.ts`**: Added `validateNativeParams()` runtime check that validates every param is string, number, or null before dispatching to nativeDb. Throws a `DbError` with the offending param index and type on violation.

3. **queryAll/queryGet accept arbitrary SQL**: Updated doc comments on both Rust methods to clarify they are designed for SELECT statements and note that DML/DDL is not an intended use. Current callers all pass SELECT-only SQL from `NodeQuery.build()`.

4. **hasTable probe semantics**: This was actually a regression, not an intentional change. Fixed by restoring the original "table exists = incremental" semantics. The try/catch now sets `hasTable = true` on successful query execution regardless of row count. This was the CI failure root cause.

carlos-alm:
@claude

PR #693, opened 2026-03-30
fix(ci): resolve visibility null crash and sequence dataflow annotation

carlos-alm:
## Summary
- **fix(native):** napi-rs `Option<String>` and `Option<u32>` accept `undefined` but not `null`. The insert-nodes stage was mapping `visibility` and `endLine` to `null` via `?? null`, causing `bulkInsertNodes` to crash with *"Failed to convert JavaScript value Null into rust type String on InsertNodesDefinition.visibility"*. Changed to `?? undefined`.
- **fix(sequence):** When native engine is available, `openRepo` returns `NativeRepository` which lacks a raw SQLite handle. `annotateDataflow` was bailing at the `instanceof SqliteRepository` guard, silently dropping dataflow return arrows and parameter annotations. Now opens a fallback read-only `better-sqlite3` connection for the two dataflow queries when needed.

## Test plan
- [x] Full test suite passes locally: 115 files, 2129 tests, 0 failures
- [x] Lint clean (0 errors)
- [ ] CI should go fully green on all 3 platforms (ubuntu, macos, windows)

### DIFF
--- a/src/domain/graph/builder/stages/insert-nodes.ts
+++ b/src/domain/graph/builder/stages/insert-nodes.ts
@@ -58,14 +58,14 @@ function tryNativeInsert(ctx: PipelineContext): boolean {
       name: string;
       kind: string;
       line: number;
-      endLine?: number | null;
-      visibility?: string | null;
+      endLine?: number;
+      visibility?: string;
       children: Array<{
         name: string;
         kind: string;
         line: number;
-        endLine?: number | null;
-        visibility?: string | null;
+        endLine?: number;
+        visibility?: string;
       }>;
     }>;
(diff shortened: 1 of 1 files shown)


carlos-alm:
Addressed the P2 finding: removed `| null` from `endLine` and `visibility` fields in the local `InsertNodesBatch` interface (lines 61, 62, 65, 66 of `insert-nodes.ts`). Since napi-rs `Option<T>` only accepts `undefined` (not `null`), the type annotations were misleading. TypeScript will now correctly reject `null` values at compile time, preventing a recurrence of the crash. See commit 390878d.

PR #741, opened 2026-04-01
fix(bench): report partial native results when incremental rebuild fails

### LINKED ISSUE #716: fix(bench): report partial native results when incremental rebuild fails
## Summary

When the native benchmark worker crashes mid-run (e.g. during incremental rebuild), all native results are lost — including the successful full build, query, and no-op rebuild data. The benchmark should report partial results for phases that completed before the crash.

## Context

Follow-up to #715. The native worker in `scripts/benchmark.ts` crashes with `SQLITE_CORRUPT` during the 1-file incremental rebuild phase. Because the crash happens before `console.log(JSON.stringify(workerResult))`, `forkWorker` in `fork-engine.ts` receives no valid JSON and returns `null`.

## Proposed fix

Two options (not mutually exclusive):

1. **Structured progress reporting**: Have the worker write partial results via IPC messages after each benchmark phase (full build, no-op rebuild, 1-file rebuild, queries). The parent can assemble whatever completed.

2. **Try/catch around incremental rebuild**: Wrap the incremental rebuild section in a try/catch so the worker can still emit results for the full build and queries even if incremental fails. Set the failed fields to `null` with an error note.

Option 2 is simpler and sufficient for now.

carlos-alm:
## Summary
- Wrap no-op and 1-file incremental rebuild phases in try/catch in both `benchmark.ts` and `incremental-benchmark.ts` so a crash (e.g. `SQLITE_CORRUPT`) emits partial JSON with `null` fields instead of losing all results
- Protect the finally-block cleanup rebuild with its own try/catch
- Update `update-benchmark-report.ts` and `update-incremental-report.ts` to render `n/a` for null incremental values and skip null values in regression checks

Closes #716

## Test plan
- [ ] Run `node scripts/benchmark.js` — verify normal output unchanged
- [ ] Simulate incremental failure (e.g. corrupt DB mid-run) — verify partial results emitted with null incremental fields
- [ ] Verify report generators handle null values without crashing

### DIFF
--- a/scripts/benchmark.ts
+++ b/scripts/benchmark.ts
@@ -158,7 +158,7 @@ try {
 	}
 	noopRebuildMs = Math.round(median(noopTimings));
 } catch (err) {
-	console.error(`  [${engine}] No-op rebuild failed: ${err.message}`);
+	console.error(`  [${engine}] No-op rebuild failed: ${(err as Error).message}`);
 }
 
 console.error(`  [${engine}] Benchmarking 1-file rebuild...`);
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed the remaining Greptile P2 nit about `err.message` on unknown-typed catch variables in e7b6cac: all four catch blocks in `benchmark.ts` and `incremental-benchmark.ts` now use `(err as Error).message`, matching the codebase convention (e.g. `change-journal.ts`).

PR #745, opened 2026-04-02
perf: route query analysis through native Rust engine

carlos-alm:
## Summary

- Query analysis functions (`fnDepsData`, `fnImpactData`, `fileDepsData`, `impactAnalysisData`, `briefData`, `implementationsData`, `interfacesData`) were hardcoded to use `openReadonlyOrFail()` (better-sqlite3), completely bypassing the native Rust engine even when `--engine native` was set
- Added `withRepo()` helper that routes queries through `openRepo()`, which tries `NativeRepository` (rusqlite via napi-rs) first and falls back to `SqliteRepository` (better-sqlite3)
- Migrated 7 analysis functions to use `withRepo` — the benchmark-critical query path now goes through native when available
- Added `getFileHash`, `hasImplementsEdges`, `hasCoChangesTable` to Repository interface with implementations in both SqliteRepository and NativeRepository
- Updated `normalizeSymbol` to accept Repository in addition to raw DB handle
- Refactored `buildTransitiveCallers` to use `repo.findCallers()` instead of raw SQL prepared statements
- `bfsTransitiveCallers` accepts both `BetterSqlite3Database` and `Repository` for backward compatibility

## Benchmark context

Before this change, the query benchmark showed native appearing *slower* than WASM. Investigation revealed queries weren't using native at all — both engines ran identical better-sqlite3 code paths. With this change, native queries will use rusqlite directly.

## Test plan

- [ ] `npm test` — integration tests for queries, fn-deps, fn-impact pass
- [ ] Type-checks cleanly (`tsc --noEmit`)
- [ ] Lint clean (`biome check`)
- [ ] Run query benchmark (`npm run benchmark`) with native addon — verify native queries match or beat WASM
- [ ] Verify WASM fallback: queries still work when native addon is unavailable

### DIFF
--- a/crates/codegraph-core/src/build_pipeline.rs
+++ b/crates/codegraph-core/src/build_pipeline.rs
@@ -57,6 +57,9 @@ pub struct BuildPipelineResult {
     pub edge_count: i64,
     pub file_count: usize,
     pub early_exit: bool,
+    pub changed_count: usize,
+    pub removed_count: usize,
+    pub is_full_build: bool,
 }
 
 /// Normalize path to forward slashes.
@@ -191,6 +194,9 @@ pub fn run_pipeline(
             edge_count: 0,
             file_count: collect_result.files.len(),
             early_exit: true,
+            changed_count: 0,
+            removed_count: 0,
+            is_full_build: false,
         });
     }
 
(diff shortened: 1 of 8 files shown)


carlos-alm:
Fixed the remaining P2 `hasCoChangesTable` memoization gap noted in the second review. Both `SqliteRepository` and `NativeRepository` now cache `hasCoChangesTable` results with a private `#coChangesTableCache` field, matching the pattern already used for `hasImplementsEdges`. See commit 50d28f7.

PR #780, opened 2026-04-03
fix(embed): handle absolute file paths from native engine

### LINKED ISSUE #752: bug(embed): double-path bug prevents embedding storage on native builds
## Found during dogfooding v3.8.0

**Severity:** High
**Command:** `codegraph embed -m minilm`

## Reproduction

```bash
codegraph build --engine native   # or auto
codegraph embed -m minilm
codegraph search "build graph"    # returns "No embeddings found"
```

## Expected behavior

Embeddings should be generated for all symbols with source code.

## Actual behavior

The native engine stores absolute file paths in the `nodes` table (e.g., `H:/Vscode/codegraph/src/cli.ts`). The `embed` command calls `path.join(cwd, node.file)`, producing doubled paths like `H:\Vscode\codegraph\H:\Vscode\codegraph\src\cli.ts`. Every file read fails with ENOENT. **0 of 5,959 symbols get embedded.** This completely blocks semantic search.

## Root cause

Path handling mismatch between native engine (stores absolute paths) and embed command (assumes relative paths and prepends cwd). The embed code path does not check whether the stored path is already absolute before joining.

## Suggested fix

In the embed command's file-reading logic, check if the node's file path is already absolute before prepending cwd. Something like:
```js
const filePath = path.isAbsolute(node.file) ? node.file : path.join(cwd, node.file);
```

carlos-alm:
## Summary

Closes #752.

- `buildEmbeddings` in `generator.ts` unconditionally did `path.join(rootDir, file)` on file paths from the DB. When the native engine stores absolute paths, this produces doubled paths (`H:/repo/H:/repo/src/cli.ts`), causing every `readFileSync` to fail with ENOENT — **0 of 5,959 symbols get embedded**
- Same pattern existed in `findReverseDependencies` in `detect-changes.ts`
- Fix: check `path.isAbsolute()` before joining with rootDir

## Test plan

- [x] New regression test: inserts a node with an absolute file path, verifies `buildEmbeddings` succeeds
- [x] All existing search tests pass (69/69)
- [x] All detect-changes tests pass (4/4)

### DIFF
--- a/src/domain/search/models.ts
+++ b/src/domain/search/models.ts
@@ -96,13 +96,26 @@ export function getModelConfig(modelKey?: string): ModelConfig {
 }
 
 /**
- * Prompt the user to install a missing package interactively.
+ * Attempt to install a missing package.
+ * In TTY environments, prompts the user for confirmation first.
+ * In non-TTY environments (CI, piped stdin), installs automatically with a log message.
  * Returns true if the package was installed, false otherwise.
- * Skips the prompt entirely in non-TTY environments (CI, piped stdin).
  * @internal Not part of the public barrel.
  */
 export function promptInstall(packageName: string): Promise<boolean> {
-  if (!process.stdin.isTTY) return Promise.resolve(false);
+  if (!process.stdin.isTTY) {
+    info(`Installing ${packageName} (optional dependency for semantic search)…`);
+    try {
+      execFileSync('npm', ['install', '--no-save', packageName], {
+        stdio: 'inherit',
+        timeout: 300_000,
+      });
+      return Promise.resolve(true);
+    } catch (err) {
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed in e44ad60:

**P1 — Native engine fast-path not patched**: Applied the same `path.isAbsolute(dep) ? dep : path.join(rootDir, dep)` guard to the native fast-path in `findReverseDependencies` (line 352). Both branches now handle absolute paths correctly.

**P2 — Test only covers source strategy**: The existing test (no strategy option) already tests `structured` (the default). Added an explicit `source` strategy test case so both strategies are covered. 22/22 tests pass.

carlos-alm:
@greptileai

PR #785, opened 2026-04-03
fix: add concurrency group to codegraph-impact workflow

carlos-alm:
## Summary
- The `codegraph-impact.yml` workflow was the only CI workflow missing a `concurrency` group
- Without it, every push to every open PR queued a new run without cancelling stale ones
- With many PRs active simultaneously, this starved the GitHub Actions runner pool and left jobs queued for over an hour
- Adds `concurrency` with `cancel-in-progress: true`, scoped per PR number — matching the pattern already used by `ci.yml`, `embedding-regression.yml`, and `publish.yml`

## Test plan
- [x] Verified all other workflows already have concurrency groups
- [ ] Push multiple commits to a PR and confirm only the latest run executes

### DIFF
--- a/.github/workflows/codegraph-impact.yml
+++ b/.github/workflows/codegraph-impact.yml
@@ -2,7 +2,7 @@ name: Codegraph Impact Analysis
 on: [pull_request]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


carlos-alm:
@greptileai

PR #788, opened 2026-04-03
fix(native): resolve dataflow null paramIndex and import edge key mismatch

### LINKED ISSUE #750: bug(native): build orchestrator does not emit import/reexport/dynamic-import edges
## Found during dogfooding v3.8.0

**Severity:** Critical
**Command:** `codegraph build --engine native`

## Reproduction

```bash
codegraph build --no-incremental --engine native
codegraph stats --json
```

## Expected behavior

Native engine should produce the same edge kinds as WASM: `imports`, `imports-type`, `dynamic-imports`, `reexports`.

## Actual behavior

Native build produces **0** import-type edges:
- Native: 23,386 edges (calls: 4,368, contains: 13,101, extends: 10, implements: 27, parameter_of: 5,298, receiver: 582)
- WASM: 26,779 edges (adds: imports: 1,201, imports-type: 244, dynamic-imports: 136, reexports: 135, plus 1,084 more call edges)

This is a **14.5% total edge gap** and a **24.8% call edge gap**. The missing import edges cascade into broken results for: `deps` (0 imports/imported-by), `impact` (0 dependents), `map` (0 coupling), `path` (no inter-file paths), `communities` (no module boundaries), `fn-impact` (0 callers for cross-file calls).

## Root cause

The Rust build orchestrator (#740) completes the full pipeline but does not write `imports`, `imports-type`, `dynamic-imports`, or `reexports` edge kinds to the database. The import edge building migration (#738) appears incomplete — the native path skips these edge types entirely.

## Suggested fix

Ensure the native Rust import edge builder (`crates/codegraph-core/`) generates all 4 import-related edge kinds that the JS pipeline produces. Verify with engine parity tests comparing edge kind distributions.

### LINKED ISSUE #753: bug(native): dataflow edge insertion fails with null paramIndex
## Found during dogfooding v3.8.0

**Severity:** Medium
**Command:** `codegraph build` (both engines)

## Reproduction

```bash
codegraph build --verbose 2>&1 | grep dataflow
```

## Expected behavior

Dataflow edges should be stored in the database for functions with parameter flows.

## Actual behavior

Build logs show:
```
buildDataflowEdges failed: Failed to convert napi value Null into rust type `u32` on DataflowEdge.paramIndex
```

The `dataflow` table has 0 rows regardless of engine. The `codegraph dataflow <name>` command always returns "No dataflow data found."

## Root cause

The Rust napi-rs type definition for `DataflowEdge.paramIndex` is `u32`, but the JS side passes `null` for edges that don't have a parameter index (e.g., return value flows). The Rust side fails to convert `null` to `u32`.

## Suggested fix

Change the Rust struct field to `Option<u32>` and handle the `None` case, or provide a sentinel value (e.g., `u32::MAX`) from the JS side when `paramIndex` is null.

carlos-alm:
## Summary
- **#753**: Fix dataflow edge insertion failing with `null` paramIndex — napi-rs `Option<u32>` expects `undefined`, not `null`. Changed `paramIndex` from `null` to `undefined` for `returns`/`mutates` edges in `dataflow.ts`.
- **#750**: Fix native import edge builder producing 0 edges on Windows — `buildImportEdgesNative` built resolve keys with backslash `rootDir` but Rust normalizes to forward slashes, causing every lookup to miss. Normalized `rootDir` in JS keys. Added defensive fallback: if native returns 0 edges when imports exist, retry with the JS implementation.

## Test plan
- [x] `tests/integration/build-parity.test.ts` — 4/4 pass (nodes, edges, roles, ast_nodes identical)
- [x] `tests/engines/dataflow-parity.test.ts` — 13/13 pass
- [x] `tests/integration/build.test.ts` — incremental edge preservation test now passes
- [x] All engine tests pass (62 passed, 4 skipped)
- [x] Biome lint clean on changed files

Closes #750, closes #753

### DIFF
--- a/.claude/skills/sweep/SKILL.md
+++ b/.claude/skills/sweep/SKILL.md
@@ -206,6 +206,11 @@ After addressing all comments for a PR:
 2. Group changes by concern — each logically distinct fix gets its own commit (e.g., one commit for a missing validation, another for a naming change). Do not lump all feedback into a single commit.
 3. Use descriptive messages per commit: `fix: <what this specific change does> (#<number>)`
 4. Push to the PR branch.
+5. **If the push is rejected** (e.g., by a hook or commitlint), diagnose the error before retrying:
+   - **Commitlint failure** (bad commit message format): This is the ONE case where amend + force-push is allowed. Fix the message with `git commit --amend -m "correct message"` then `git push --force-with-lease`.
+   - **Hook denial** (guard-git.sh blocking staged files not in session edit log): The worktree has no edit log — commit with explicit file paths (`git commit <file1> <file2> -m "msg"`) instead of staging first.
+   - **Branch name validation failure**: You are on the wrong branch — check out the correct PR branch before retrying.
+   - **Any other failure**: Fix with a new commit. Never amend + force-push for code changes.
 
 ### 2g. Re-trigger reviewers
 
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed Greptile review feedback in c010ba5:

- **paramIndex type tightened** (`src/types.ts:2223`): Changed `paramIndex?: number | null` to `paramIndex?: number` in `bulkInsertDataflow` to match the napi-rs `Option<u32>` contract. This prevents future callers from passing `null` and re-introducing the insertion failure. Note: the `getDataflowEdges` return types still use `number | null` since those are read types (what SQLite returns), where `null` is correct.

- **Fallback false-positive documented** (`build-edges.ts:744-750`): Added a NOTE comment explaining that the fallback also fires for codebases where every import targets an external package, with no correctness impact — just a redundant JS traversal.

PR #869, opened 2026-04-05
fix(perf): wire engine selection through openRepo to fix query benchmarks

carlos-alm:
## Summary

- `openRepo()` always tried native NAPI path first regardless of engine selection, causing both "wasm" and "native" benchmark workers to measure the same native rusqlite open/close overhead (~27ms vs ~10ms with direct better-sqlite3)
- This produced a false 177% regression in fnDeps query benchmarks (v3.7.0→v3.9.0), blocking CI on **all open PRs**
- Add `engine` param to `openRepo()` and check `CODEGRAPH_ENGINE` env var so query functions respect engine selection
- Set `CODEGRAPH_ENGINE` in benchmark fork workers
- Add v3.9.0 fnDeps depth 1/3/5 to `KNOWN_REGRESSIONS` (published data measured with broken routing)

## Root cause

`openRepo()` in `src/db/connection.ts` had no engine selection — it unconditionally called `isNativeAvailable()` and used the native path when available. The benchmark fork (`scripts/lib/fork-engine.ts`) set `__BENCH_ENGINE__` per worker to label engine runs, but `openRepo` never checked it. On CI where both engines are available, both workers measured identical native performance.

## Test plan

- [x] `regression-guard.test.ts` — 14/14 pass (fnDeps regressions now in KNOWN_REGRESSIONS)
- [x] `queries-unit.test.ts` — 74/74 pass
- [x] Integration tests — all query tests pass
- [x] TypeScript compiles clean
- [x] Lint clean on changed files

### DIFF
--- a/tests/benchmarks/regression-guard.test.ts
+++ b/tests/benchmarks/regression-guard.test.ts
@@ -343,6 +343,38 @@ describe('Benchmark regression guard', () => {
     'INCREMENTAL_BENCHMARK_DATA',
   );
 
+  // Warn when KNOWN_REGRESSIONS entries are stale (more than 1 minor version
+  // behind the current package version).  This makes the stale-exemption
+  // problem self-detecting rather than requiring manual bookkeeping.
+  test('KNOWN_REGRESSIONS entries are not stale', () => {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const pkgVersion: string = JSON.parse(
+      fs.readFileSync(path.join(ROOT, 'package.json'), 'utf8'),
+    ).version;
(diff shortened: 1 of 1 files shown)


carlos-alm:
Addressed the P1 `openReadonlyWithNative` engine-bypass gap (from the Greptile review). Fixed in 99b6e0f — added the same `CODEGRAPH_ENGINE` env-var guard to `openReadonlyWithNative()` so it skips the native path when `engine === 'wasm'`, consistent with `openRepo()`. This closes the latent measurement gap for structure-query, dataflow, and module-map callers.

Issue #837, opened 2026-04-04
codegraph: role classification misses consumers through barrel re-exports


carlos-alm:
## Context
Discovered during Titan audit (phase: gauntlet, date: 2026-04-04).

## Description
Role classification does not trace consumption through barrel re-exports. For example, `queryName` in `src/presentation/queries-cli/inspect.ts` is classified as test-only (0 consumers) but is actually consumed in production via the barrel chain: `queries-cli/index.ts -> queries-cli.ts -> query.ts`.

## Impact
Unknown number of symbols incorrectly classified as dead or test-only when their consumption path goes through barrel re-exports.

## Source
- **Titan phase:** gauntlet
- **Severity:** bug
- **Category:** codegraph

PR #918, opened 2026-04-13
fix(native): align Rust role classification with JS

carlos-alm:
## Summary
- Native Rust role classifier was producing different (less accurate) results than the JS fallback, causing the `roles` phase to classify nodes incorrectly when using `--engine native`
- Adds `imports-type` edge support to all fan-in, exported-ID, and production fan-in queries
- Adds reexport-chain exported detection via recursive CTE (#837), matching the JS barrel-tracing logic
- Adds `constant` + `hasActiveFileSiblings` logic so constants in active files classify as `leaf` instead of `dead-leaf`
- Adds `!isExported` guard to test-only classification
- Includes `imports-type` and `reexports` in incremental neighbour file discovery

## Test plan
- [x] `cargo check` passes with no new warnings
- [x] All 41 role classification unit tests pass
- [x] 509 integration tests pass (7 failures are pre-existing WASM grammar issue in worktree)
- [ ] Run `codegraph build --engine native` and `codegraph build --engine wasm` on a real project, compare `codegraph roles` output — should now match

### DIFF
--- a/crates/codegraph-core/src/roles_db.rs
+++ b/crates/codegraph-core/src/roles_db.rs
@@ -115,6 +115,7 @@ fn classify_node(
     fan_out: u32,
     is_exported: bool,
     production_fan_in: u32,
+    has_active_file_siblings: bool,
     median_fan_in: f64,
     median_fan_out: f64,
 ) -> &'static str {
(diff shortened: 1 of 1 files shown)


carlos-alm:
Addressed the two minor suggestions from the review:

1. **Inline comments for reexport chain logic**: Added detailed comments to both CTE blocks (full-classify step 3b and incremental path) explaining the base-case/recursive-step mechanics and how barrel reexport chains are traced. See commit a6b8799.

2. **Extract CTE as constant**: Evaluated this but the two CTEs are not identical — the incremental version has an additional `AND n.file IN (...)` scope filter and uses different parameter binding. Extracting a shared constant would require templating the differences back in, which would reduce clarity rather than improve it. Keeping them separate with cross-references in the comments is cleaner.

carlos-alm:
@greptileai

PR #926, opened 2026-04-14
fix(bench): use per-engine file count in benchmark reports

### LINKED ISSUE #925: fix: benchmark per-file metrics use wrong file count for native engine
## Problem

Reported by Greptile in PR #924: the native engine's `perFile.*` values in the benchmark data don't reconcile with the raw totals when divided by the reported file count (727).

**Root cause:** In `scripts/benchmark.ts`, each engine worker computes its own `totalFiles` from the DB after building. However, `formatEngineResult()` (lines 40-55) strips the per-engine `files` field, and the parent process uses `primary.files` (WASM's count) as the top-level `files` field for both engines.

The report script (`scripts/update-benchmark-report.ts` line 104) then uses `h.files` (the top-level count) in the table for both native and WASM rows, even though native may have processed fewer files (~663 vs 727).

## Fix

1. Include `files` in `formatEngineResult()` in `scripts/benchmark.ts`
2. Update `update-benchmark-report.ts` to use each engine's own file count in the table row (falling back to top-level if not present for backward compat)

## Impact

Cosmetic — the per-file values are internally consistent with the native engine's actual file count, but the displayed `Files` column is misleading. Extrapolated estimates (50K files) use the potentially-incorrect per-file values.

Ref: PR #924

carlos-alm:
## Summary

- Include `files` in `formatEngineResult()` so each engine preserves its own file count
- Use `e.files ?? h.files` in `engineRow()` and raw totals section for correct per-engine display (with backward compat fallback for older history entries)

Previously, the native engine's `Files` column and raw totals always showed the WASM file count (~727) instead of native's own count (~663), making per-file metrics appear inconsistent.

Closes #925

## Test plan

- [ ] Run `node scripts/benchmark.js` and verify the JSON output includes `files` inside both `native` and `wasm` objects
- [ ] Pipe output to `node scripts/update-benchmark-report.js` and verify BUILD-BENCHMARKS.md shows different file counts per engine when they differ
- [ ] Verify backward compat: older history entries without per-engine `files` still render correctly using the top-level fallback

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,17 +617,17 @@ Codegraph also extracts symbols from common callback patterns: Commander `.comma
 
 Self-measured on every release via CI ([build benchmarks](generated/benchmarks/BUILD-BENCHMARKS.md) | [embedding benchmarks](generated/benchmarks/EMBEDDING-BENCHMARKS.md) | [query benchmarks](generated/benchmarks/QUERY-BENCHMARKS.md) | [incremental benchmarks](generated/benchmarks/INCREMENTAL-BENCHMARKS.md) | [resolution precision/recall](tests/benchmarks/resolution/)):
 
-*Last updated: v3.9.2 (2026-04-09)*
+*Last updated: v3.9.3 (2026-04-13)*
 
 | Metric | Native | WASM |
 |---|---|---|
-| Build speed | **15.7 ms/file** | **15 ms/file** |
-| Query time | **27ms** | **40ms** |
-| No-op rebuild | **9ms** | **20ms** |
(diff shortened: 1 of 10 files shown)


carlos-alm:
Fixed in 7f154b9:

- Added `files: data.files` to `formatEngineResult()` in `scripts/benchmark.ts` so each engine preserves its own file count in the JSON output
- Changed `h.files` → `e.files ?? h.files` in both `engineRow()` (line 104) and the raw totals block (line 148) in `scripts/update-benchmark-report.ts`, with fallback for older history entries that lack per-engine `files`

@greptileai

Issue #982, opened 2026-04-21
bug: build --no-incremental silently wipes the embeddings table


carlos-alm:
## Found during dogfooding v3.9.4

**Severity:** High
**Command:** `codegraph build --no-incremental`

Running a full rebuild after `codegraph embed` silently drops every row from the `embeddings` table. There is no warning, no prompt, no opt-out. Users who have spent minutes generating embeddings (especially with the larger Jina models) lose all of them on the next `--no-incremental` build.

## Reproduction

```bash
mkdir embed-test && cd embed-test
cat > a.js <<'EOF'
export function alpha() { return 1; }
export function beta() { return alpha(); }
EOF

npx codegraph build .
npx codegraph embed . -m minilm
# Stored 2 embeddings (384d, ...) in graph.db

node -e "const db = require('better-sqlite3')('.codegraph/graph.db'); \
  console.log('before:', db.prepare('SELECT COUNT(*) c FROM embeddings').get().c);"
# before: 2

npx codegraph build . --no-incremental
node -e "const db = require('better-sqlite3')('.codegraph/graph.db'); \
  console.log('after:',  db.prepare('SELECT COUNT(*) c FROM embeddings').get().c);"
# after: 0
```

## Expected behavior

One of:
1. Preserve embeddings whose `node_id` still maps to a live node in the rebuilt graph (ideal).
2. Warn the user before wiping (`[codegraph WARN] --no-incremental will discard N embeddings; re-run \`codegraph embed\` after the build.`).
3. Require an explicit `--wipe-embeddings` flag to opt into destruction.

## Actual behavior

Embeddings table is emptied silently. Subsequent `codegraph search` returns zero results with no hint about why.

## Suggested fix

The simplest safe change is option 2: warn before wiping. For option 1, keep the embeddings table intact, and after the full rebuild either (a) leave embeddings keyed to their old `node_id`s that no longer exist (and let a downstream validator prune), or (b) re-key embeddings by `symbol signature` (name + file + kind) so they survive re-identification.

## Related

Full rebuild also invalidates any external consumers that cached `node_id`s from the old DB. A migration note in the release notes would help, but the silent data loss is the bigger issue.

PR #986, opened 2026-04-21
fix(build): warn before --no-incremental wipes embeddings

carlos-alm:
## Summary

`codegraph build --no-incremental` (and any full rebuild promoted by engine/schema/version mismatch) silently drops every row in the `embeddings` table. Users who spent minutes on `codegraph embed` lose their embeddings with no warning; the next `codegraph search` returns zero hits with no clue why.

Add a `warnOnEmbeddingsWipe` step at the pipeline entry — just after the incremental/forceFullRebuild decision is made. When the build is full AND the embeddings table is non-empty, it logs:

```
[codegraph WARN] Full rebuild will discard N embeddings; re-run `codegraph embed` after the build.
```

Because the warning sits at the shared pipeline entry point, it fires uniformly regardless of which path (native orchestrator or JS fallback) actually performs the delete.

## Found during

Dogfooding v3.9.4 — see #982

## Test plan

- [x] Full rebuild with embeddings present → warning fires with correct count
- [x] Full rebuild with empty/missing embeddings table → no warning
- [x] Incremental build (even if promoted to full by mismatch) → warning still fires only when embeddings exist
- [x] `npm run typecheck` clean
- [x] `npm run lint` clean

### DIFF
--- a/src/domain/graph/builder/pipeline.ts
+++ b/src/domain/graph/builder/pipeline.ts
@@ -976,6 +976,10 @@ export async function buildGraph(
             `Codegraph version changed (${prevVersion} → ${CODEGRAPH_VERSION}), promoting to full rebuild.`,
           );
           ctx.forceFullRebuild = true;
+          // Re-check embeddings: the initial warnOnEmbeddingsWipe ran before
+          // forceFullRebuild was set here, so the silent-data-loss guard
+          // would otherwise miss this late-promotion path (#986 follow-up).
+          warnOnEmbeddingsWipe(ctx);
         }
       }
     }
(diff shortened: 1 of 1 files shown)


carlos-alm:
Addressed Greptile's P1 feedback on the late-promotion gap:

- Added a `warnOnEmbeddingsWipe(ctx)` call inside the `catch` block at `src/domain/graph/builder/pipeline.ts:972-980`, right after setting `ctx.forceFullRebuild = true` on a JS `CODEGRAPH_VERSION` mismatch.
- Now the silent-data-loss guard also fires for the native-orchestrator-throws \u2192 version-mismatch path, not just the initial `setupPipeline` evaluation.
- Comment in the code cross-references this scenario so future edits do not regress.

Pushed in b4c907b4.

carlos-alm:
@greptileai

PR #989, opened 2026-04-21
fix(louvain): demote native-path parity warning to debug

carlos-alm:
## Summary

- `louvainCommunities` emitted `[codegraph WARN] ... maxLevels/maxLocalPasses/refinementTheta are ignored by the native Rust path` on every communities computation — not only when the user explicitly ran `codegraph communities`.
- Root cause: `DEFAULTS.community` in `src/infrastructure/config.ts` always populates `maxLevels` (50), `maxLocalPasses` (20), and `refinementTheta` (1.0), and `features/communities.ts` forwards those config values into the options bag. The `!= null` guard in `louvain.ts` therefore always matched, so the "warn if caller set them" logic never actually gated anything.
- Demote the message to `debug`. It's an informational parity note (Leiden-specific knobs ignored by the native Louvain path), not user-actionable — anyone debugging engine behavior can raise the log level.

## Test plan

- [x] `npx vitest run tests/graph/algorithms/louvain.test.ts` — 3/3 pass
- [x] `npx biome check src/graph/algorithms/louvain.ts` — clean
- [ ] Confirm no residual warning output on `codegraph audit` / `codegraph stats` with native engine


carlos-alm:
Addressed Claude's minor suggestion from the review:

- Added a `Leiden-knob parity logging` regression suite in `tests/graph/algorithms/louvain.test.ts` that spies on stderr and asserts `[codegraph WARN]` is never emitted when `maxLevels` / `maxLocalPasses` / `refinementTheta` are forwarded (as `DEFAULTS.community` always does). A companion test verifies the message, when it does fire on the native path, goes through `[codegraph DEBUG]` with verbose enabled — guarding against a future `debug` → `warn` regression.

Tests: 5/5 green (`npx vitest run tests/graph/algorithms/louvain.test.ts`). Biome check clean.

carlos-alm:
Re: Greptile — thanks for the 5/5 review. I pushed one follow-up commit (`624233b9`) adding a regression test per Claude's optional suggestion. The test asserts the parity message never emits at `WARN` level when the community config defaults are forwarded, and (when the native path fires it) goes through `DEBUG`. No production code changes beyond the original one-line demotion.

carlos-alm:
@greptileai

carlos-alm:
@claude

PR #993, opened 2026-04-21
feat(build): report collectMs/detectMs as separate phases

### LINKED ISSUE #977: obs: report collectMs/detectMs as separate phases in BuildResult
## Context
`BuildResult.phases` currently rolls collect + detect-change work into `setupMs`. For incremental-build perf investigations these two phases have very different cost profiles (file-walk + ignore-matching vs hash/mtime checks), and collapsing them hides where time is actually being spent.

## Proposal
Add `collectMs` and `detectMs` to `BuildResult.phases` and report them separately on both the JS and native paths.

- `src/types.ts` — add the two fields to `BuildResult.phases` (optional to keep back-compat at type level, or required + backfilled from existing native timing).
- `src/domain/graph/builder/context.ts` — add `collectMs` / `detectMs` to `ctx.timing`.
- `src/domain/graph/builder/stages/collect-files.ts` — wrap the exported `collectFiles` entry with `performance.now()` and write `ctx.timing.collectMs`.
- `src/domain/graph/builder/stages/detect-changes.ts` — same pattern around `detectChanges`.
- `src/domain/graph/builder/pipeline.ts` — `formatTimingResult` and `formatNativeTimingResult` should surface the two fields instead of folding native-side `collectMs`/`detectMs` back into `setupMs`.

The Rust `PipelineTiming` already tracks `collect_ms` and `detect_ms` (see `crates/codegraph-core/src/build_pipeline.rs`), so the native path only needs the formatter change — no new instrumentation.

## Why not bigger
This is an observability change, not a perf change. Splitting it out keeps the "One PR = one concern" rule and makes it trivial to review independently of any perf work it supports.

carlos-alm:
## Summary
- Splits collect and detect-change work out of `setupMs` in `BuildResult.phases` so incremental-build perf investigations can see the two phases separately (file-walk + ignore-matching vs hash/mtime checks).
- JS path instruments `collectFiles` and `detectChanges` with `performance.now()` wrapped in `try/finally` so early returns still record timing.
- Native path stops folding `collectMs`/`detectMs` back into `setupMs` — the Rust `PipelineTiming` already tracks them, so the formatter just surfaces them.

Closes #977.

## Files touched
- `src/types.ts` — adds `collectMs`/`detectMs` to `BuildResult.phases`
- `src/domain/graph/builder/context.ts` — adds `collectMs`/`detectMs` to `ctx.timing`
- `src/domain/graph/builder/stages/collect-files.ts` — wraps `collectFiles` with `performance.now()`
- `src/domain/graph/builder/stages/detect-changes.ts` — wraps `detectChanges` with `performance.now()`
- `src/domain/graph/builder/pipeline.ts` — `formatTimingResult` and `formatNativeTimingResult` surface the two fields
- `tests/builder/pipeline.test.ts` — asserts the new fields

No README/CLAUDE/ROADMAP updates — observability-only change.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/builder` (14/14 pass)
- [x] `npx vitest run tests/integration/incremental-parity.test.ts` (pass)
- [ ] CI runs the full suite

### DIFF
--- a/src/domain/graph/builder/context.ts
+++ b/src/domain/graph/builder/context.ts
@@ -87,6 +87,8 @@ export class PipelineContext {
   // ── Phase timing ───────────────────────────────────────────────────
   timing: {
     setupMs?: number;
+    collectMs?: number;
+    detectMs?: number;
     parseMs?: number;
     insertMs?: number;
     resolveMs?: number;
--- a/src/domain/graph/builder/pipeline.ts
+++ b/src/domain/graph/builder/pipeline.ts
@@ -184,6 +184,8 @@ function formatTimingResult(ctx: PipelineContext): BuildResult {
   return {
     phases: {
       setupMs: +(t.setupMs ?? 0).toFixed(1),
+      collectMs: +(t.collectMs ?? 0).toFixed(1),
+      detectMs: +(t.detectMs ?? 0).toFixed(1),
       parseMs: +(t.parseMs ?? 0).toFixed(1),
       insertMs: +(t.insertMs ?? 0).toFixed(1),
       resolveMs: +(t.resolveMs ?? 0).toFixed(1),
@@ -558,7 +560,9 @@ function formatNativeTimingResult(
 ): BuildResult {
   return {
     phases: {
-      setupMs: +((p.setupMs ?? 0) + (p.collectMs ?? 0) + (p.detectMs ?? 0)).toFixed(1),
+      setupMs: +(p.setupMs ?? 0).toFixed(1),
+      collectMs: +(p.collectMs ?? 0).toFixed(1),
+      detectMs: +(p.detectMs ?? 0).toFixed(1),
       parseMs: +(p.parseMs ?? 0).toFixed(1),
       insertMs: +(p.insertMs ?? 0).toFixed(1),
       resolveMs: +(p.resolveMs ?? 0).toFixed(1),
--- a/src/domain/graph/builder/stages/collect-files.ts
+++ b/src/domain/graph/builder/stages/collect-files.ts
@@ -7,6 +7,7 @@
  */
 import fs from 'node:fs';
 import path from 'node:path';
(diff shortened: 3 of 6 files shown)


carlos-alm:
@greptileai Thanks — addressed the PipelineContext timing interface drift in bf3c3d3a. Replaced `timing: Record<string, number>` in `src/types.ts` with the full typed object shape (optional `setupMs/collectMs/detectMs/...` fields + `[key: string]: number | undefined` index signature) so the interface in `types.ts` now mirrors the concrete class in `context.ts`. Consumers typed against the interface will now see the new fields and tolerate `undefined` timing values.

- `npx tsc --noEmit`: clean
- `npx vitest run tests/builder`: 14/14 pass
- `biome check` on touched files: clean

carlos-alm:
@greptileai

carlos-alm:
Addressed Greptile's P2 feedback in 87ce95da:

- **Issue 1 — scoped-path `collectMs`/`detectMs` overlap:** `collectFiles` now times only the filesystem-walk portion (existence checks + file list construction) under `collectMs`. Change-detection outputs (`parseChanges`/`removed`/`isFullBuild`) are attributed to `detectMs` via an additive accumulator, matching the semantics of the non-scoped path. `detectChanges` also accumulates `detectMs` additively so the contribution from `collectFiles` isn't clobbered.
- **Issue 2 — positive-value assertions:** `tests/builder/pipeline.test.ts` now asserts `toBeGreaterThanOrEqual(0)` on `collectMs` and `detectMs` in addition to the `typeof` check, so a silent regression where the `finally` block never runs would be caught.

Verified:
- `npx tsc --noEmit`: clean
- `npx vitest run tests/builder`: 14/14 pass
- `npx vitest run tests/integration/incremental-parity.test.ts`: 12/12 pass
- `biome check` on touched files: clean

carlos-alm:
@greptileai

Issue #1054, opened 2026-05-03
perf(native): ~2 second flat overhead added to rebuild operations in v3.10.0


carlos-alm:
## Summary

The v3.10.0 pre-publish benchmark gate detected a catastrophic native-engine regression on rebuild operations. A flat ~2 second overhead was added to every rebuild on the native engine. The publish was correctly blocked from npm by the regression guard ([failed run](https://github.com/optave/ops-codegraph-tool/actions/runs/25236744550/job/74005864801)).

## Regression deltas (v3.10.0 vs v3.9.6, native engine, 25% threshold)

**Build benchmark:**
- No-op rebuild: 13ms → 2125ms (**+16246%**)
- 1-file rebuild: 78ms → 2143ms (**+2647%**)

**Incremental benchmark:**
- No-op rebuild: 10ms → 2081ms (**+20710%**)
- 1-file rebuild: 54ms → 2139ms (**+3861%**)

WASM engine was unaffected on the same run, so this is native-path-specific.

## Pattern

A flat ~2080ms overhead is added regardless of how much work the rebuild has to do (no-op vs 1-file produce nearly identical absolute times). This is the signature of something being initialized or loaded **per call** that previously persisted across calls — not a slowdown in rebuild logic itself.

## Important context

The benchmark CI built v3.10.0 source against the **v3.9.6 native binary** (the published artifact). Logs show:
```
[codegraph] Using native engine (v3.9.6)
```

That means the regression is on the **JavaScript/TypeScript wrapper around native**, not in the Rust extractor crate.

## Suspected causes (PRs between v3.9.6 and v3.10.0)

- **#1038 — perf(wasm): scope ensureWasmTrees re-parse to files that need it** — touched `wasm-worker-entry.ts` serialization and `ensureWasmTrees` filtering. If `ensureWasmTrees` (or the JS path that wraps native parses) is now invoked on the native path, or the empty-array serialization affects native flow, that would match the symptom.
- **#1041 — fix(parity): gate astTypeMap lookup with Object.hasOwn** — changed extractor lookup behavior. Lower-probability cause but worth ruling out.

## Reproduction

```bash
git checkout v3.10.0
npm install
npm run build
node --import ./scripts/ts-resolve-loader.js scripts/build-benchmark.ts --version 3.10.0
node --import ./scripts/ts-resolve-loader.js scripts/incremental-benchmark.ts --version 3.10.0
```

[...]

Issue #965, opened 2026-04-18
bug: WASM engine crashes V8 reproducibly on Windows + Node 22 when building codegraph source


carlos-alm:
## Found during dogfooding v3.9.4

**Severity:** High — WASM engine is unusable on this platform for this target.
**Platform:** Windows 11 x64, Node 22.18.0, codegraph 3.9.4 installed clean from npm.
**Command:** `codegraph build . --engine wasm` on the codegraph source repo (535 files).

## Reproduction

```bash
mkdir /tmp/dogfood && cd /tmp/dogfood
npm init -y && npm install @optave/codegraph@3.9.4
rm -rf <codegraph-source-repo>/.codegraph
npx codegraph build <codegraph-source-repo> --engine wasm
```

Reproduces in 5/5 attempts, with four distinct failure modes across runs:

### Mode 1: WebAssembly unreachable trap
```
# Fatal error in , line 0
# unreachable code
#FailureMessage Object: ...
----- Native stack trace -----
 1: node::SetCppgcReference+16599
 2: node::TriggerNodeReport+74095
 3: V8_Fatal+197
 ...
```

### Mode 2: V8 check failure in code lookup
```
# Fatal error in , line 0
# Check failed: maybe_code.has_value().
```

### Mode 3: Tree-sitter WASM `call_indirect` trap (cascades into issue #964)
```
[codegraph WARN] Parse error in src/extractors/javascript.ts: table index is out of bounds
codegraph: fatal error — Cannot read properties of undefined (reading 'toLowerCase')
```

### Mode 4: Plain segfault
```
/c/Program Files/nodejs/npx: line 65:  1019 Segmentation fault      "$NODE_EXE" "$NPX_CLI_JS" "$@"
```

All four occur after `[codegraph] Incremental: 535 changed, 0 removed` and before (or during) the parse phase completes.

## Scope

| Target | Native | WASM |
|---|---|---|
| Full codegraph repo (535 TS files) | builds in ~2s, 14295 nodes / 29186 edges | crashes consistently |
| Subset: `src/shared + src/infrastructure + src/cli` (64 TS files) | 59 ms, 679 / 1164 | 256 ms, 679 / 1164 — works fine |

Engine parity on the subset is exact at the node/edge level, so the bug only manifests at scale on this target.

## Analysis

- `table index is out of bounds` is a WebAssembly `call_indirect` trap — indicates memory corruption inside the tree-sitter WASM module (likely `tree-sitter-typescript`).
[...]

PR #1058, opened 2026-05-04
fix(perf): scope WASM grammar load in engine-parity backfill (#1054)

carlos-alm:
## Summary

- Avoid the worker-pool first-call overhead (~1.7s on slow CI runners) when only a handful of files need WASM backfill.
- Adds `parseFilesWasmInline` — a main-thread parse path that loads only the grammars for the input extensions and returns symbols with `_tree` set, so the unified walker in `runAnalyses` populates AST/CFG/dataflow downstream.
- Routes both backfill call sites (`parseFilesAuto` per-call backfill, `backfillNativeDroppedFiles` post-orchestrator backfill) through `parseFilesWasmForBackfill`, which picks inline for batches ≤ 16 files and the worker pool for larger batches (preserving tree-sitter WASM crash isolation #965 where it matters).

## Context

#1054 — every native `buildGraph` call in the pre-publish gate logs `Native orchestrator dropped 4 file(s) in natively-supported languages — likely a Rust extractor bug. Backfilling via WASM: .tf` and falls back via the worker pool. This PR makes the fallback cheap for that case. The underlying Rust-side drop (why the freshly-built binary drops these 4 fixture files when the v3.9.6 published binary doesn't) is tracked separately and addressed in a follow-up PR.

## Test plan

- [x] `npx tsc --noEmit` clean.
- [x] `npx vitest run tests/integration/build-parity.test.ts` — 4/4 pass.
- [x] `npx vitest run tests/benchmarks/resolution/resolution-benchmark.test.ts -t "hcl"` — 5/5 pass (HCL backfill is the prime case the new path serves).
- [x] Local `incremental-benchmark.ts` reports identical numbers (full=2773ms, noop=17ms, 1-file=266ms) to before — Windows didn't pay the worker overhead anyway, so unchanged is expected. Effect should land on CI Linux.
- [ ] Pre-publish benchmark gate runs cleanly on dispatch.

Refs #1054

### DIFF
--- a/src/domain/graph/builder/pipeline.ts
+++ b/src/domain/graph/builder/pipeline.ts
@@ -38,6 +38,7 @@ import {
   getActiveEngine,
   getInstalledWasmExtensions,
   parseFilesAuto,
+  parseFilesWasmForBackfill,
 } from '../../parser.js';
 import { setWorkspaces } from '../resolve.js';
 import { PipelineContext } from './context.js';
@@ -793,7 +794,7 @@ async function backfillNativeDroppedFiles(ctx: PipelineContext): Promise<void> {
       `Native orchestrator dropped ${totals['native-extractor-failure']} file(s) in natively-supported languages — likely a Rust extractor bug. Backfilling via WASM: ${formatDropExtensionSummary(byReason['native-extractor-failure'])}`,
     );
   }
-  const wasmResults = await parseFilesAuto(missingAbs, ctx.rootDir, { engine: 'wasm' });
+  const wasmResults = await parseFilesWasmForBackfill(missingAbs, ctx.rootDir);
 
   const rows: unknown[][] = [];
   const exportKeys: unknown[][] = [];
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed Greptile's P1 — WASM tree leak in `backfillNativeDroppedFiles`.

Added a cleanup pass after `batchInsertNodes` that mirrors `releaseWasmTrees`: iterates the `wasmResults` map, calls `tree.delete()` on every `symbols._tree`, and clears `_tree` / `_langId`. Now the inline-backfill path frees its WASM linear memory on the same path the finalize sweep would have if these symbols had been routed through `ctx.allSymbols`.

Commit: 49c9461

carlos-alm:
Addressed Claude's review feedback:
- Added a doc comment on `INLINE_BACKFILL_THRESHOLD` explaining the 16-file rationale (sized for typical engine-parity drops; recurring HCL case is 4 files; threshold is where worker-pool IPC + crash-isolation cost amortizes).

Commit: ffd3431

carlos-alm:
@greptileai

PR #1062, opened 2026-05-04
chore(bench): per-iteration phase trace for #1054 diagnosis [DO NOT MERGE]

carlos-alm:
## Purpose

Throwaway diagnostic branch. **Do not merge.**

The pre-publish gate shows a flat ~2-second-per-call native overhead in CI that doesn't reproduce locally on Windows or in Docker Linux. The .tf drops (#1054) explain part of the picture but not this — even with backfill firing locally, no-op rebuild is ~7ms; in CI it's 2307ms.

This branch adds per-iteration phase breakdown to \`benchmark.ts\` and \`incremental-benchmark.ts\`. Every full / no-op / 1-file iteration prints a line like:

\`\`\`
[1054-trace] native.noop[0] total=2307ms setup=? collect=? detect=? parse=? insert=? resolve=? edges=? structure=? roles=? ast=? complexity=? cfg=? dataflow=? finalize=?
\`\`\`

When the gate runs against this branch, the log will show which specific stage (or stages) eats the 2 seconds — \`parseMs\`, \`edgesMs\`, \`rolesMs\`, \`structureMs\`, etc. That'll tell us where to dig next.

## Plan

1. Dispatch \`publish.yml\` against this branch.
2. Read the gate log, grep for \`[1054-trace]\`.
3. Identify the dominant phase.
4. Open the actual fix in a focused PR.
5. Close this PR without merging.

## Why diagnostic prints don't go to main

The trace lines go to stderr only and don't affect the JSON the gate consumes, so the gate's regression-guard verdict is unchanged. But committing diagnostic prints to mainline benchmarks pollutes future runs and obscures real signal. Hence the throwaway-branch shape.

Refs #1054

### DIFF
--- a/scripts/benchmark.ts
+++ b/scripts/benchmark.ts
@@ -126,12 +126,27 @@ function selectTargets() {
 const origLog = console.log;
 console.log = (...args) => console.error(...args);
 
+// ───── Issue #1054 timing instrumentation (one-off) ─────
+// Per-iteration phase breakdown so we can see which step eats the 2s/call
+// CI overhead. Throwaway after we read the gate logs.
+function logTrace(scenario: string, iter: number, totalMs: number, phases: any) {
+	const p = phases || {};
+	const order = [
+		'setupMs', 'collectMs', 'detectMs', 'parseMs', 'insertMs', 'resolveMs',
+		'edgesMs', 'structureMs', 'rolesMs', 'astMs', 'complexityMs', 'cfgMs',
+		'dataflowMs', 'finalizeMs',
+	];
+	const parts = order.map(k => `${k.replace(/Ms$/, '')}=${p[k] ?? '?'}`).join(' ');
+	console.error(`[1054-trace] ${engine}.${scenario}[${iter}] total=${Math.round(totalMs)}ms ${parts}`);
+}
+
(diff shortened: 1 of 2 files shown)


carlos-alm:
Thanks for the review. Acknowledged the minor observations:

- Duplicate `logTrace` between `benchmark.ts` and `incremental-benchmark.ts`: intentional — this is throwaway diagnostic code that will be deleted once we read the gate log and identify the dominant phase. Extracting a shared util would just mean more cleanup churn when we revert.
- `any` type for `phases`: same rationale — kept loose because the trace just stringifies whatever the build returns.

No code changes needed. The branch will be closed without merging once the CI gate log gives us the phase breakdown we need to open the focused fix PR.

carlos-alm:
Closing unmerged in favor of #1064, which is the actual fix for #1054.

This branch was created as a diagnostic timing-trace harness to investigate why native incremental no-op rebuilds were ~2s in CI. The investigation produced the JS-side fast-skip implemented in #1064 (now ready to merge). The per-iteration phase tracing here was a one-shot diagnostic tool and is no longer needed.

Branch will not be deleted in case the trace harness is useful for future investigations.

PR #1069, opened 2026-05-05
fix(native): persist file_hashes for dropped/symbol-less files

### LINKED ISSUE #1068: fast-skip: file_hashes drops files when grammar isn't installed (breaks fast-skip permanently)
## Symptom

The CI bench gate's diagnostic logging from #1067 surfaced this rejection on every no-op rebuild:

```
[fast-skip] false: collected file missing from file_hashes: tests/benchmarks/resolution/fixtures/clojure/main.clj
```

The file is on disk, parses fine locally, and Clojure is in `LANGUAGE_REGISTRY` (`src/domain/parser.ts:699-703`, extensions `.clj/.cljs/.cljc`, `required: false`). But on the CI runner, the file is collected by `collectFiles` and never recorded in `file_hashes` — so the next no-op rebuild's pre-flight gate sees a "new" file every time and rejects.

This permanently disables the JS-side fast-skip (#1064) for any project that contains files in optional-language extensions when the grammar isn't installed.

## Root cause

`buildFileHashes` at `src/domain/graph/builder/stages/insert-nodes.ts:102` iterates the symbol map, not the collected file set:

```ts
for (const [relPath] of allSymbols) {
  // ...
  fileHashes.push({ file: relPath, hash, mtime, size });
}
```

If parsing didn't produce any symbols for a file (because the WASM grammar is unavailable, the file is empty, or the parser silently no-opped), `allSymbols` has no entry → no `file_hashes` row written. But `collectFiles` still picks up the file on subsequent runs because `.clj` is in the registry's extension list, regardless of grammar availability.

The asymmetry: **`collectFiles` keys off registered extensions; `buildFileHashes` keys off produced symbols.**

## Reproduction

1. Run a build on a project containing a file in any `required: false` language whose WASM grammar isn't installed (e.g., a `.clj` file when `tree-sitter-clojure.wasm` was skipped during install).
2. Run a no-op rebuild.
3. With `CODEGRAPH_FAST_SKIP_DIAG=1`, observe `[fast-skip] false: collected file missing from file_hashes: <that file>`.
4. The fast-skip pipeline runs the full ~2s rebuild instead of returning fast.

CI evidence: https://github.com/optave/ops-codegraph-tool/actions/runs/25396408231 — search bench job logs for `clojure/main.clj`.

## Suggested fix

Option 1 — **Hash all collected files**, regardless of symbol production. In `buildFileHashes`, iterate `filesToParse` (or the collected file set in ctx) instead of `allSymbols`. Files with zero symbols still get a `file_hashes` row with their content hash, so subsequent no-op rebuilds match.

Option 2 — **Filter `collectFiles` to installed grammars.** Don't pick up files for extensions whose grammar isn't loaded. Cleaner conceptually but breaks the fixture/test path for grammar-missing scenarios and may surprise users whose graph silently shrinks when an optional grammar fails to install.

Option 1 is preferred — it preserves the user-visible \"this file exists in my project\" behavior and keeps fast-skip working.

## Related

- #1064 introduced the JS-side fast-skip whose pre-flight gate is being permanently broken by this asymmetry.
- #1066 is the regression umbrella; this is one of the contributing causes (alongside the engine=wasm finding).
- #1067 added the diagnostic that exposed this.

## Verification

After fix, the diag log should no longer print `collected file missing from file_hashes` for a `.clj` (or other optional-language) file on noop rebuilds.

carlos-alm:
## Summary
- Iterate `filesToParse` in `buildFileHashes` / `updateFileHashes` so files with zero symbols (empty, parser no-op, optional-language grammar unavailable) still get a `file_hashes` row.
- Persist `file_hashes` rows in `backfillNativeDroppedFiles` for every file the Rust orchestrator dropped (e.g. `.clj` when no Rust extractor exists).
- Run the engine-parity backfill on every successful orchestrator pass, not just full builds — the Rust orchestrator's narrower file_collector deletes `nodes` + `file_hashes` rows for files outside its `SUPPORTED_EXTENSIONS` on every incremental run, and the JS-side fast-skip pre-flight (#1054) then permanently rejects on "collected file missing from file_hashes". Restructured the function to do the cheap missing-file check before the proxy→better-sqlite3 handoff so no-op rebuilds remain fast.

Closes #1068

## Test plan
- [ ] CI green
- [ ] Manual E2E (mixed `a.js` + `main.clj` repo, `cfg: false, dataflow: false`):
  - Build #1 (fresh): full build, `file_hashes` includes both `a.js` and `main.clj`.
  - Build #2 (no-op): fast-skip fires (~2 ms), both rows preserved.
  - Build #3 (no-op): fast-skip fires (~1 ms).
  - After modifying `a.js`: orchestrator runs, picks up new symbol, `main.clj` row still preserved.
- [ ] New unit test `tests/builder/insert-nodes.test.ts` (6 cases) covers `buildFileHashes` directly: symbol-less files, precomputed hash, `_reverseDepOnly`, disk fallback, metadata-only updates, deduplication.

## Notes
- Vitest runs are blocked locally on Node 24 by a pre-existing `vitest.config.ts` issue (`--strip-types is not allowed in NODE_OPTIONS`); behavior was verified via build + lint + TypeScript + standalone E2E against the built `dist/`.

### DIFF
--- a/src/domain/graph/builder/pipeline.ts
+++ b/src/domain/graph/builder/pipeline.ts
@@ -767,10 +767,30 @@ async function backfillNativeDroppedFiles(ctx: PipelineContext): Promise<void> {
     collected.files.map((f) => normalizePath(path.relative(ctx.rootDir, f))),
   );
 
-  const existingRows = ctx.db
+  const existingNodeRows = ctx.db
     .prepare("SELECT DISTINCT file FROM nodes WHERE kind = 'file'")
     .all() as Array<{ file: string }>;
-  const existing = new Set(existingRows.map((r) => r.file));
+  const existingNodes = new Set(existingNodeRows.map((r) => r.file));
+
+  // Belt-and-suspenders: also check `file_hashes`. The fast-skip pre-flight
+  // (#1054) rejects on `file_hashes` gaps, and the two tables can diverge
+  // (e.g. a DB written by old code where `nodes` was populated but
+  // `file_hashes` was not). Treating "in nodes but not in file_hashes" as
+  // missing closes the gap so the backfill repairs the file_hashes row even
+  // when the node row already exists.
+  let existingHashes = new Set<string>();
+  try {
+    const existingHashRows = ctx.db
+      .prepare('SELECT DISTINCT file FROM file_hashes')
+      .all() as Array<{ file: string }>;
+    existingHashes = new Set(existingHashRows.map((r) => r.file));
+  } catch (e) {
+    // file_hashes table may not exist on legacy DBs; treat as fully missing
+    // so the backfill writes rows on the upsert path below.
+    debug(
+      `backfillNativeDroppedFiles: file_hashes read failed (table may not exist): ${toErrorMessage(e)}`,
+    );
+  }
 
   // Restrict backfill to files with an installed WASM grammar. Extensions in
   // LANGUAGE_REGISTRY without a shipped grammar file (e.g. groovy, erlang on
@@ -780,7 +800,9 @@ async function backfillNativeDroppedFiles(ctx: PipelineContext): Promise<void> {
   const missingRel: string[] = [];
   const missingAbs: string[] = [];
   for (const rel of expected) {
-    if (existing.has(rel)) continue;
+    // A file is "missing" if it's absent from EITHER nodes OR file_hashes.
+    // Both must be present for fast-skip to work correctly.
+    if (existingNodes.has(rel) && existingHashes.has(rel)) continue;
     const ext = path.extname(rel).toLowerCase();
     if (!installedExts.has(ext)) continue;
     missingRel.push(rel);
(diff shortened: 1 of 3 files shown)


carlos-alm:
Addressed Greptile's feedback (P2 'Early-return signal diverges from fast-skip signal'):

- backfillNativeDroppedFiles now reads BOTH `nodes` and `file_hashes` and treats any expected file absent from EITHER table as missing.
- Closes the divergence gap where a file present in `nodes` but missing from `file_hashes` would early-return without repairing the hash row, leaving the fast-skip pre-flight (#1054) permanently rejecting.
- The existing INSERT OR REPLACE upsert path now repairs both rows. The new `file_hashes` SELECT is wrapped in try/catch so legacy DBs without the table fall through gracefully.

Commit: 9b2aa25

carlos-alm:
@greptileai

Issue #1075, opened 2026-05-06
native fast-skip rejects some no-op rebuilds due to mtime precision drift between Rust and JS


carlos-alm:
## Symptom

After #1074 fixed the `engine_version` mismatch loop, the publish gate's no-op rebuild benchmark dropped from **2007ms → 23ms** (was +19970%, now +130% over the 25% threshold). The remaining slowdown comes from a small fraction of no-op iterations falling out of the fast-skip path.

Failed run: https://github.com/optave/ops-codegraph-tool/actions/runs/25459431687

Diagnostic lines (CODEGRAPH_FAST_SKIP_DIAG=1):

```
[fast-skip] false: mtime/size diff for src/graph/algorithms/leiden/adapter.ts:
  stat=1778100332555/7498 stored=1778100332554/7498 (mtimeMs=1778100332555)
```

Same size (7498), mtime off by exactly **1ms** — JS reads 1778100332555, Rust wrote 1778100332554.

## Root cause

The two layers compute integer-ms from nanosecond-precision mtimes via incompatible paths:

- **Rust write** (`crates/codegraph-core/src/change_detection.rs:339`):
  ```rust
  d.as_millis() as i64
  ```
  `Duration::as_millis()` truncates the sub-ms portion. For an mtime of `1778100332554999999ns`, this yields `1778100332554`.

- **JS read** (`src/domain/graph/builder/stages/detect-changes.ts:184`):
  ```ts
  Math.floor(stat.mtimeMs)
  ```
  Node's `stat.mtimeMs` is an IEEE-754 f64 derived from nanoseconds (effectively `mtimeNs / 1e6` in float). At large epoch-ms values the f64 representation rounds: `1778100332554.999999` is not exactly representable and round-trips as `1778100332555.0`, so `Math.floor` yields `1778100332555` — one ms ahead of Rust's truncation.

The same drift exists at the JS *write* sites (`detect-changes.ts:299/309/537/548`, `insert-nodes.ts:115/130`, etc.) — they all use `Math.floor(stat.mtimeMs)`, so a JS-built DB read by JS is consistent, and a Rust-built DB read by Rust is consistent. The mismatch only appears at the boundary, which is exactly what the native-engine build path triggers: Rust writes, JS reads on the next incremental.

## Fix

Read mtime via BigInt nanoseconds in JS, then truncate with integer math:

```ts
const stat = fs.statSync(file, { bigint: true });
[...]

PR #1080, opened 2026-05-08
revert: "fix(native): read mtime via BigInt nanoseconds (#1079)"

carlos-alm:
## Summary

- Reverts #1079.
- The fix targeted a fast-skip optimization miss caused by f64 rounding of `stat.mtimeMs`. Investigation showed it was not a correctness bug — when fast-skip misses, the file is re-hashed from contents and the resulting graph is byte-identical either way. The fix only changed the optimization hit rate.
- The miss-rate window is ~0.026% of fresh-file mtimes (the ~256 ns range where `Number(mtimeNs) / 1e6` rounds across a millisecond boundary). The fix paid a constant-cost `fs.statSync(path, { bigint: true })` overhead on **100%** of stat calls in the change-detection hot path to recover that 0.026% miss rate.
- The post-merge benchmark gate confirmed the trade was net-negative on the most-affected workload — no-op rebuild went from +70% over baseline (warning) to +108% (hard fail in regression-guard). See run https://github.com/optave/ops-codegraph-tool/actions/runs/25532744746/job/74943479391.

Reopens #1075. The right path forward is to document that fast-skip is a best-effort optimization (the on-disk hash is the source of truth for graph state) rather than chase exact integer parity with Rust's `Duration::as_millis()`.

## Test plan

- [x] `npx vitest run tests/builder/detect-changes.test.ts` — passes (10/10).
- [x] `npm run lint` — clean (only pre-existing warnings unrelated to this revert).
- [ ] Watch the post-merge `Pre-publish benchmark gate` run on main to confirm the no-op rebuild metric returns to its pre-#1079 level.

### DIFF
--- a/src/domain/graph/builder/helpers.ts
+++ b/src/domain/graph/builder/helpers.ts
@@ -222,18 +222,12 @@ export function fileHash(content: string): string {
 }
 
 /**
- * Stat a file, returning integer-truncated mtime in ms (and size).
- *
- * Reads via BigInt nanoseconds and truncates with integer math so the value
- * matches Rust's `Duration::as_millis() as i64` exactly. `Math.floor(stat.mtimeMs)`
- * cannot be substituted: at large epoch values the f64 `mtimeMs` rounds, so a
- * Rust-written `file_hashes.mtime` of N can read back as N+1 in JS and bust the
- * fast-skip path on every native→JS handoff.
+ * Stat a file, returning { mtimeMs, size } or null on error.
  */
-export function fileStat(filePath: string): { mtime: number; size: number } | null {
+export function fileStat(filePath: string): { mtimeMs: number; size: number } | null {
   try {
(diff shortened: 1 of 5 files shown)


carlos-alm:
Addressed Greptile P2 feedback ("`Math.floor` responsibility pushed to every call site"):

- `fileStat` now floors `mtimeMs` inside the helper and returns `{ mtime: number; size: number }` (already-integer mtime that matches the DB column).
- All six call sites simplified from `Math.floor(stat.mtimeMs)` to `stat.mtime`.
- Behaviour unchanged; pushed as commit `bf3f4b4` (refactor — separate from the mechanical revert).

This eliminates the forward-looking concern that a future call site could store an un-floored `mtimeMs` and cause spurious fast-skip misses.

carlos-alm:
@greptileai

PR #1153, opened 2026-05-18
fix(hooks): guard-git.sh uses [[:space:]] in sed for BSD compatibility

### LINKED ISSUE #1132: guard-git.sh: detect_work_dir regex fails on macOS sed, blocks valid pushes from worktrees
## Problem

The branch-name validation in `.claude/hooks/guard-git.sh` uses `\s` in its `sed -nE` regex to extract the `cd <path>` prefix from chained commands. BSD sed (default on macOS) does not interpret `\s` as whitespace, so the regex silently produces no output, `detect_work_dir` falls back to an empty path, and `validate_branch_name` reads the current shell cwd's branch instead of the target worktree's branch.

## Reproduction

In a worktree where the cwd branch is `worktree-agent-abc`, attempting to push from another worktree with a valid branch name is rejected:

```
BLOCKED: Branch 'worktree-agent-abc' does not match required pattern.
```

The command `cd /path/to/other-worktree && git pu​sh origin HEAD:fix/some-fix` is read by the hook, but `detect_work_dir` cannot extract the `cd` target so validation runs against the wrong branch.

## Root cause

`guard-git.sh:124`:

```bash
work_dir=$(echo "$COMMAND" | sed -nE 's/^\s*cd\s+"?([^"&]+)"?\s*&&.*/\1/p')
```

BSD sed regex syntax does not recognize `\s`. Use `[[:space:]]` instead, consistent with line 121:

```bash
work_dir=$(echo "$COMMAND" | sed -nE 's/^[[:space:]]*cd[[:space:]]+"?([^"&]+)"?[[:space:]]*&&.*/\1/p')
```

## Workaround

Use `git --git-dir=<wt>/.git --work-tree=<wt> ...` from any directory — this bypasses the hook's branch detection entirely (the regex looks for `git -C`, not the long-form flags).

## Why this matters

Parallel sessions in worktrees cannot complete pushes when the agent's shell cwd is itself a worktree with a non-conformant branch name (the harness creates worktrees with names like `worktree-agent-<hash>`).

carlos-alm:
## Summary
- BSD sed (macOS default) does not interpret `\s` as whitespace, so the `cd <path> && git ...` extractor and the `git -C` normalization regexes in `.claude/hooks/guard-git.sh` silently produced no output.
- `detect_work_dir` then fell back to the shell cwd's branch, blocking valid pushes from worktrees whose cwd had a non-conformant branch name (e.g. `worktree-agent-<hash>`).
- Replaced all three `\s` sites in sed regexes with `[[:space:]]`, which is portable across BSD and GNU sed.

## Test plan
- [x] Manual: feed the hook a JSON payload with `{"command":"cd /tmp/fake-wt && git push origin HEAD:fix/some-fix"}` while cwd is on a non-conformant branch — hook now exits 0 (previously denied).
- [x] Manual: confirm BSD sed locally treats `\s` as literal `s` and `[[:space:]]` as expected.

Closes #1132

### DIFF
--- a/.claude/hooks/guard-git.sh
+++ b/.claude/hooks/guard-git.sh
@@ -33,8 +33,8 @@ fi
 # the opening `"` of a quoted path (which would leave a trailing `path"` in
 # NCOMMAND). The pattern re-anchors on `git`, so multi-`-C` chains (e.g.
 # `git -C /a -C /b push`) need a second pass to collapse the residual `-C`.
-NCOMMAND=$(echo "$COMMAND" | sed -E 's/(^|\s|&&\s*)git[[:space:]]+-C[[:space:]]+"[^"]+"/\1git/g; s/(^|\s|&&\s*)git[[:space:]]+-C[[:space:]]+[^"[:space:]][^[:space:]]*/\1git/g')
-NCOMMAND=$(echo "$NCOMMAND" | sed -E 's/(^|\s|&&\s*)git[[:space:]]+-C[[:space:]]+"[^"]+"/\1git/g; s/(^|\s|&&\s*)git[[:space:]]+-C[[:space:]]+[^"[:space:]][^[:space:]]*/\1git/g')
+NCOMMAND=$(echo "$COMMAND" | sed -E 's/(^|[[:space:]]|&&[[:space:]]*)git[[:space:]]+-C[[:space:]]+"[^"]+"/\1git/g; s/(^|[[:space:]]|&&[[:space:]]*)git[[:space:]]+-C[[:space:]]+[^"[:space:]][^[:space:]]*/\1git/g')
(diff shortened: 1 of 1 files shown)


carlos-alm:
Thanks for the review @greptileai. This PR is being closed as superseded — PR #1146 (already merged on `main`) applied a more comprehensive fix for the same BSD-sed root cause, covering both this issue (#1132) and the related `;t;` problem (#1135).

The remaining `\s` in `grep -qE` patterns that you flagged is a real follow-up: tracked as issue #1157 against current main.

carlos-alm:
Closing as superseded by #1146 (already merged on main).

PR #1146 fixed both BSD-sed root causes in `.claude/hooks/guard-git.sh` in one commit:
- Replaced `\s` with `[[:space:]]` in all sed patterns (covers this issue, #1132).
- Split the `s///;t;s///` chain into two `sed -nE` invocations (covers sibling issue #1135 / sibling PR #1152).

The diff in this PR would partially revert #1146 (it restores `\s` in two grep guards and the `;t;` chain). Closing without merge.

Follow-up filed as #1157 for the remaining `\s` usages in the file's `grep -qE` patterns.

Issue #1166, opened 2026-05-19
follow-up: deduplicate tracer validation work between script and gate test


carlos-alm:
Spun off from #1052.

The resolution dedup landing in #1052 (this PR) covers the resolution gate but leaves a similar — albeit more involved — duplication on the tracer side.

**Where the duplication is:**

- `scripts/resolution-benchmark.ts` calls `runDynamicTracer(lang)` for every fixture (for telemetry: writes \`dynamicEdges\` / \`dynamicConfirmed\` counts into \`resolution-result.json\`).
- `tests/benchmarks/resolution/tracer/tracer-validation.test.ts` independently calls \`runTracer(lang)\` for every fixture to compute same-file recall.

Both end up running the per-language tracer subprocess once each, doubling the tracer cost in the pre-publish job.

**Why it's not as trivial as the resolution dedup:**

The shapes and toolchain-missing semantics differ. The script's \`runDynamicTracer\` returns \`[]\` for both \"ran, found nothing\" and \"toolchain missing\". The test's \`runTracer\` returns \`null\` for \"toolchain missing\" so it can skip gracefully and treats empty edges as 0% recall. Deduplicating cleanly means:

1. Extending the resolution script to emit raw tracer edges per language (not just counts), with a status field (\`ok\` | \`skipped\`).
2. Updating \`tracer-validation.test.ts\` to optionally consume that artifact via an env var (e.g. \`RESOLUTION_RESULT_JSON\`, same as the resolution gate), preserving the existing skip-on-toolchain-missing behavior.
3. Wiring the env var into the \"Run tracer validation\" step in \`.github/workflows/publish.yml\`.

That's a more invasive change to the script's output format, so it's worth doing in its own PR to keep #1052 reviewable.

PR #1167, opened 2026-05-19
refactor(ci): let resolution gate reuse benchmark artifact

### LINKED ISSUE #1052: follow-up: deduplicate resolution benchmark work between script and gate test
Deferred from PR #1051 review.

**Original reviewer comment:** https://github.com/optave/ops-codegraph-tool/pull/1051#discussion_r3177797420

**Context:**

Greptile flagged that the `pre-publish-benchmark` job effectively runs the full resolution benchmark twice:

1. **Step `Run resolution benchmark`** runs `scripts/resolution-benchmark.ts`, which copies every fixture, calls `buildGraph` for each language, computes precision/recall, and writes `resolution-result.json`.
2. **Step `Gate on resolution thresholds`** runs `tests/benchmarks/resolution/resolution-benchmark.test.ts` via vitest, which independently copies every fixture, calls `buildGraph` for each language, and asserts thresholds.

Both perform identical work — fixture copy, full WASM-engine `buildGraph` per language, edge extraction, metric computation. The gate test does not consume `resolution-result.json`; it regenerates everything. With 30+ language fixtures this roughly doubles the most expensive part of the pre-publish job.

This duplication pre-dates PR #1051 (it existed in `benchmark.yml` from #1014 and earlier — see commit 2e86f56e); PR #1051 only restored the gates that were accidentally dropped in #1040. The fix is a separate refactor and was therefore out of scope for the restoration PR.

**Proposed fix:**

Refactor `tests/benchmarks/resolution/resolution-benchmark.test.ts` to read `resolution-result.json` (path passed via env var, e.g. `RESOLUTION_RESULT_JSON`) and assert thresholds against the already-computed metrics, instead of rebuilding graphs. Local runs (no env var set) can keep the current build-from-fixtures path so developers can run the test standalone. Wire the env var through in `publish.yml` so the gate step reuses the artifact produced by `Run resolution benchmark`.

The same approach should be considered for the tracer validation gate if a similar duplication exists.

carlos-alm:
## Summary

- The `pre-publish-benchmark` job ran the full resolution benchmark twice — once in `scripts/resolution-benchmark.ts` (which writes `resolution-result.json`) and again in `tests/benchmarks/resolution/resolution-benchmark.test.ts` (which independently rebuilt every fixture). With ~34 language fixtures this roughly doubled the slowest part of the pre-publish pipeline.
- This change extends the script's `LangResult` output to include `falsePositiveEdges` / `falseNegativeEdges` arrays, and teaches the gate test to consume `resolution-result.json` when `RESOLUTION_RESULT_JSON` is set. The legacy build-from-fixtures path stays as a fallback so devs can still run the test standalone with `npx vitest run …`.
- `publish.yml` now passes `RESOLUTION_RESULT_JSON: ${{ github.workspace }}/resolution-result.json` into the Gate step.

## Test plan

- [x] `npx vitest run tests/benchmarks/resolution/resolution-benchmark.test.ts -t javascript` — fixture mode, 5/5 pass
- [x] Generate `resolution-result.json` via `node scripts/resolution-benchmark.ts` (34 languages, all keys present including the new edge lists)
- [x] `RESOLUTION_RESULT_JSON=/tmp/resolution-result.json npx vitest run …` — artifact mode, 170/170 pass in ~0.5 s (vs. minutes for fixture rebuild)
- [x] Sanity-check error paths: missing artifact file → clear error; artifact missing `falsePositiveEdges` → clear "regenerate" error
- [x] Biome lint on the changed files clean (pre-existing warnings in unrelated test files only)
- [ ] Watch publish workflow on next dispatch to confirm CI savings

## Notes

The issue also asks to consider the same dedup for the tracer validation gate. That duplication exists, but it requires a more invasive change to the script's output format (raw tracer edges + a status field, not just counts), and the toolchain-missing semantics differ between the script and the tracer test. Tracked separately in #1166 to keep this PR focused.

Closes #1052

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -333,6 +333,12 @@ jobs:
 
       - name: Gate on resolution thresholds
         timeout-minutes: 30
+        # Reuse the metrics produced by the previous step instead of rebuilding
+        # every fixture from scratch (issue #1052). The gate test falls back to
+        # the build-from-fixtures path when this env var is unset, so local
+        # runs (`npx vitest run …`) still work standalone.
+        env:
+          RESOLUTION_RESULT_JSON: ${{ github.workspace }}/resolution-result.json
         run: npx vitest run tests/benchmarks/resolution/resolution-benchmark.test.ts --reporter=verbose
 
       - name: Run tracer validation (same-file edge recall)
--- a/scripts/resolution-benchmark.ts
+++ b/scripts/resolution-benchmark.ts
@@ -127,9 +127,12 @@ function computeMetrics(resolvedEdges: ResolvedEdge[], expectedEdges: ExpectedEd
 		m.recall = m.expected > 0 ? m.resolved / m.expected : 0;
 	}
 
+	// Keep full precision so the artifact-mode gate compares the exact same
+	// values the fixture-mode gate would compute. Rounding here let a near-miss
+	// like 0.8497 round up to 0.850 and silently clear a 0.85 threshold.
 	return {
-		precision: Math.round(precision * 1000) / 1000,
-		recall: Math.round(recall * 1000) / 1000,
+		precision,
+		recall,
 		truePositives: truePositives.size,
 		falsePositives: falsePositives.size,
 		falseNegatives: falseNegatives.size,
--- a/scripts/resolution-benchmark.ts
+++ b/scripts/resolution-benchmark.ts
@@ -64,6 +64,10 @@ interface LangResult {
 	totalResolved: number;
 	totalExpected: number;
 	byMode: Record<string, ModeMetrics>;
+	// Edge lists are included so the gate test can reuse this artifact
+	// instead of rebuilding fixtures from scratch (see issue #1052).
+	falsePositiveEdges: string[];
+	falseNegativeEdges: string[];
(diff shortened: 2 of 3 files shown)


carlos-alm:
Addressed Greptile feedback in e7b2f27:

- **P1 (precision/recall rounding):** Removed the `Math.round(... * 1000) / 1000` step in `scripts/resolution-benchmark.ts:computeMetrics`. The artifact now stores full-precision values so the CI gate compares the exact numbers the fixture-mode gate would, closing the 0.8497 -> 0.850 silent-pass band.
- **P2 (missing numeric field validation):** `metricsFromArtifact` now type-checks `precision`, `recall`, `truePositives`, `falsePositives`, `falseNegatives`, `totalResolved`, `totalExpected`, and `byMode` before consumption, surfacing the clear 'regenerate' error instead of a downstream `TypeError`.
- **P2 (empty artifact silently passes):** `loadArtifact` now rejects `{}` / non-object payloads with an explicit error, so an empty artifact can no longer register zero `describe` blocks and exit 0.

Verified by feeding the test crafted artifacts (empty, malformed, and valid) — each path produces the expected error or passes.

carlos-alm:
@greptileai

PR #1223, opened 2026-05-26
fix(bench/embed): drop stale HF_TOKEN gate on jina-code

carlos-alm:
## Summary
- `scripts/embedding-benchmark.ts` skipped `jina-code` whenever `HF_TOKEN` was unset, leaving dogfood reports (e.g. v3.10.1-dev.80) carrying a misleading "requires HF_TOKEN" note.
- PR #1053 already repointed the alias at the public `jinaai/jina-embeddings-v2-base-code` (Apache-2.0) and dropped the README caveat — this gate in the bench harness was missed.
- Removed the 6-line gate so `jina-code` is benchmarked alongside the other models.

The generic 401/`gated` error handler in `src/domain/search/models.ts:240` is unchanged — it stays in place as a fallback for any future gated model.

## Test plan
- [ ] `npm run bench:embed` runs `jina-code` through the worker without `HF_TOKEN` set and produces a row in the output JSON
- [ ] No type-check regressions (`npx tsc --noEmit` clean)

### DIFF
--- a/scripts/embedding-benchmark.ts
+++ b/scripts/embedding-benchmark.ts
@@ -154,7 +154,6 @@ const dbPath = path.join(root, '.codegraph', 'graph.db');
 const { MODELS } = await import(srcImport(srcDir, 'domain/search/index.js'));
 
 const TIMEOUT_MS = 1_800_000; // 30 min — with symbol sampling, embed (~18 min) + search (~5 min) fits comfortably
-const hasHfToken = !!process.env.HF_TOKEN;
 const modelKeys = Object.keys(MODELS);
 const results = {};
 let symbolCount = 0;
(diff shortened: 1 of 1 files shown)


carlos-alm:
Verifying Claude's concerns:

**PR #1053 exists and is merged** (`gh pr view 1053`):
```
title: fix(search): point jina-code alias at published HF repo
state: MERGED
mergedAt: 2026-05-04
```
PR #1053 body explicitly states: *"Drops the stale 'requires HF token' note from the README — the published repo is public."* The local `git log --grep="1053"` lookup didn't find it because squash-merge commits don't always carry the PR number in their first line.

**jina-code model is public**: `src/domain/search/models.ts:77` aliases `jina-code` to `jinaai/jina-embeddings-v2-base-code` (Apache-2.0). The previous alias `Xenova/jina-embeddings-v2-base-code` 404'd, which is what motivated PR #1053.

**Fallback intact**: As noted in the PR description (and confirmed by Greptile 5/5), the generic 401/`gated` error handler at `src/domain/search/models.ts:240` remains as the safety net for any future gated model.

**CI fix**: Merged `origin/main` in (commit `f11afce`) to pick up the `KNOWN_REGRESSIONS` pruning landed in #1217 — that was the only cause of the failing pre-publish benchmark gate. Locally `RUN_REGRESSION_GUARD=1 npx vitest run tests/benchmarks/regression-guard.test.ts` is 17/17 green.

PR #1248, opened 2026-05-30
fix(benchmarks): exempt 3.11.1 regression-guard entries that fail on publish

carlos-alm:
## Summary

- Adds `3.11.1:DB bytes/file`, `3.11.1:fnDeps depth 3`, and `3.11.1:fnDeps depth 5` to `KNOWN_REGRESSIONS` in `regression-guard.test.ts`

## Why this fails on publish but not in CI

The per-PR benchmark gate runs with `--version dev`. `assertNoRegressions` has a `dev`-only fallback: when comparing `dev` vs a baseline, KNOWN_REGRESSIONS entries keyed to the *baseline* version also exempt the metric. So `dev vs 3.11.0` was covered by `3.11.0:fnDeps depth 3` etc.

When `publish.yml` runs, it uses the real semver (`3.11.1`). The fallback doesn't fire for non-`dev` versions, and `KNOWN_REGRESSIONS.has('3.11.1:fnDeps depth 3')` is false → gate fails.

## Why the baseline is 3.10.0, not 3.11.0

3.11.0 has no **query** benchmark data in committed history, so `findLatestPair` skips it and falls back to 3.10.0. The 3.10.0 numbers predate the corpus-scope change from #1134 (resolution fixtures excluded from the build sweep), so `DB bytes/file` and `fnDeps` values look inflated against that older baseline.

## Exemption rationale

| Entry | Root cause |
|---|---|
| `3.11.1:DB bytes/file` | Corpus denominator drop from #1134: ~745 files → ~607 files; bytes constant, per-file ratio inflates |
| `3.11.1:fnDeps depth 3/5` | 3.10.0 baseline predates 3.11.0 steady-state; no fn_deps implementation change |

All three entries carry "remove once 3.12.0+ data confirms stable numbers against a 3.11.x baseline" in their doc comments — the stale-entry test will flag them automatically after 3.12.0 ships.

## Test plan

- [x] `RUN_REGRESSION_GUARD=1 npm run test:regression-guard` — 17/17 pass locally
- [ ] publish.yml regression-guard step passes on the next release run

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Run Automated AI Review
         id: automated-review
-        uses: anthropics/claude-code-action@beta
+        uses: anthropics/claude-code-action@de8e0b9c42c6cb58e904c857f164aa072244c1ac
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           model: claude-sonnet-4-6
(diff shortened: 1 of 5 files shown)


carlos-alm:
@claude

Issue #1265, opened 2026-05-31
Grammar-version drift between native (Rust crates) and WASM (npm) causes engine call-edge divergence


carlos-alm:
## Context

Spun out of #1263 (1-node/1-edge full-build divergence). The **node** divergence is fully fixed in PR for #1263 (native Ruby top-level constants + C++ header fixture rename). This issue tracks the **residual edge divergence**, whose root cause is systemic.

## Root cause

The native engine (Rust crates in `crates/codegraph-core/Cargo.toml`) and the WASM engine (npm packages in `package.json`) pin **different tree-sitter grammar versions** for many languages. Different grammar versions produce different parse trees, so the per-language extractors emit different call edges even though the extractor logic is equivalent.

Examples of the drift (native ↔ WASM):

| Grammar | Native (Cargo) | WASM (npm) |
|---|---|---|
| go | 0.23 | 0.25 |
| python | 0.23 | 0.25 |
| javascript | 0.23 | 0.25 |
| bash | 0.23 | 0.25 |
| c | 0.23 | 0.24 |
| php | 0.23 | 0.24 |
| rust | 0.23 | 0.24 |
| swift | 0.6 | 0.7 |
| java | 0.23 | 0.23.5 |
| haskell | 0.23 | 0.23.1 |
| cpp | 0.23 | 0.23.4 |
| kotlin | kotlin-sg 0.4 | kotlin 0.3.8 (different package) |

## Symptom

On a full self-build, the two engines diverge by ~30–46 call edges, concentrated in the resolution fixtures for go, python, java, php, haskell, cuda, csharp, ocaml, groovy. Each diff maps to a grammar-version mismatch above.

## Proposed fix

Align the Rust grammar crate versions with the npm WASM grammar versions (or vice-versa) and re-validate every extractor + the resolution benchmark thresholds. This is a large, multi-language change (grammar API/node-type differences may require extractor updates) and is intentionally **not** bundled with the node-divergence fix to keep that PR focused.

Per CLAUDE.md, both engines must produce identical results — this drift is the remaining structural source of divergence.

PR #1266, opened 2026-05-31
fix(extract): eliminate WASM/native node divergence

### LINKED ISSUE #1263: follow-up: fix pre-existing 1-node/1-edge WASM vs native full-build divergence
Deferred from PR #1262 review.

Noted in the dogfood report for v3.11.1 (§5, line 60): the full-build path shows a consistent 1-node / 1-edge delta between WASM and native engines (19,298 vs 19,297 nodes; 40,000 vs 39,999 edges). This divergence pre-dates 3.11.1 and is not introduced by the watch-mode fix in PR #1261.

Per CLAUDE.md, divergence between engines is a bug that should be fixed, not documented as expected behaviour. This issue tracks finding and eliminating the root cause of the persistent 1-node/1-edge gap.

carlos-alm:
## Summary

Eliminates the persistent full-build node divergence between the WASM and native engines (tracked in #1263). Two independent root causes:

- **Native Ruby extractor** never handled top-level `assignment` nodes, so program-level constants (e.g. `SKIP_CLASSES = %w[...].freeze`) were dropped while the WASM extractor extracted them. Added an `assignment` arm mirroring `handleRubyAssignment`, gated to `program`-level constant targets so class/module-body constants stay attributed to their owner.
- **C++ benchmark fixture `service.h`** was parsed as C (`.h` maps to C), where tree-sitter-c 0.23 (native) error-recovered `class Service {}` into a bogus `function` node while 0.24 (WASM) dropped it. The file is genuinely C++, so it's renamed to `service.hpp` — both engines now parse it as C++ and agree (class/method/property nodes). Updated the `#include` lines and taught the native tracer to copy `.hpp` headers.

After both fixes, the full-build node sets are **byte-identical** across engines.

## Out of scope (tracked separately)

The residual ~30–46 call-edge divergence stems from **grammar-version drift** between the Rust crates and the npm WASM packages (e.g. go 0.23↔0.25, python 0.23↔0.25, c 0.23↔0.24). That is a large, multi-language alignment effort and is tracked in #1265.

## Test plan

- [x] `cargo test --lib` — 327 pass, including the grammar-ABI regression test and 3 new Ruby extractor tests
- [x] cpp resolution benchmark passes (precision/recall thresholds unchanged)
- [x] Full-repo dual-engine build: node sets verified byte-identical (WASM vs native)

Closes #1263


carlos-alm:
Addressed review feedback:
- Added three WASM-side tests to `tests/parsers/ruby.test.ts` mirroring the Rust unit tests: `extracts top-level constant assignment`, `does not extract local variable assignment as a constant`, and `does not extract class-body constant as a top-level definition`. These cover the same three cases as the Rust tests and guard against future WASM regressions on the `handleRubyAssignment` code path.

carlos-alm:
@claude

Issue #1358, opened 2026-06-06
fix(resolver): scope Phase 8.3f typeMap key by function to avoid same-name rest-param collision


carlos-alm:
## Problem

Both the WASM (`buildCallEdgesJS`) and native (`buildObjectRestParamPostPass`) Phase 8.3f seeding paths key the typeMap entry by `restName` alone (e.g. `rest`):

```typescript
typeMap.set(orpb.restName, { type: pb.argName, confidence: 0.65 });
```

If two functions in the same file both use `...rest` as their rest binding:

```js
function f1({ a, ...rest }) { rest.method(); }
function f2({ b, ...rest }) { rest.method(); }
f1(obj1);
f2(obj2);
```

The first seeding wins via the `!typeMap.has(restName)` guard, so `f2`'s rest binding resolves to `obj1` instead of `obj2`.

## Fix

Scope the typeMap key to include the enclosing function name (e.g. `f1::rest`). The resolution path in `resolveByMethodOrGlobal` would then need to look up the scoped key using the current caller context.

This requires coordinated changes in both the seeding and the resolution lookup.

## Scope

Out of scope for PR #1355 (Phase 8.3f). Affects both WASM and native paths equally.

Issue #1336, opened 2026-06-05
feat(resolver): resolve property calls on object destructuring rest parameters (JS)


carlos-alm:
## Pattern

When a function parameter uses object destructuring with a rest element (\`...rest\`), and the rest object's property is then called, codegraph does not resolve the callee.

## Failing example (from Jelly micro-test \`rest.js\`)

\`\`\`js
var obj = {
    e1: () => { console.log("31"); },
    e2: () => { console.log("32"); },
    e3: () => { console.log("33"); },
    e4: () => { console.log("34"); }
};

function f3({ e1: eee1, ...eerest }) {
    eee1();
    eerest.e4();  // eerest.e4 === obj.e4 when called as f3(obj)
}
f3(obj);
\`\`\`

Expected edge: **\`f3 → e4\`** (via \`eerest.e4()\` with \`eerest\` holding the rest of \`obj\`) — not resolved.

## Relationship to #1321

Issue #1321 tracks array spread and \`Array.from\`/\`concat\`/\`flat\` callbacks. This issue is about *object* destructuring rest, where \`...eerest\` collects the un-destructured keys of an object parameter. The call \`eerest.e4()\` requires:

1. Tracking that \`eerest\` is the rest of the destructured parameter \`obj\`
2. Knowing the type of the argument passed to \`f3\` (the \`obj\` literal)
3. Looking up \`obj.e4\` as the callee for \`eerest.e4()\`

## Benchmark evidence

Jelly micro-test corpus (\`tests/benchmarks/resolution/fixtures/jelly-micro/rest/\`):
- Named edges: 1 · TP: 0 · Recall: 0%
- FN: \`f3@rest.js → e4@rest.js\`

PR #1404, opened 2026-06-08
fix(incremental): seed callee::restName typeMap keys and pass callerName for Phase 8.3f

### LINKED ISSUE #1369: fix(incremental): pass callerName to resolveCallTargets in incremental path for Phase 8.3f scoped-key lookup
## Problem

PR #1368 scoped Phase 8.3f typeMap keys to `callee::restName` and added a fallback in `resolveByMethodOrGlobal` that tries `${callerName}::${effectiveReceiver}` when the unscoped lookup misses. However, `resolveCallTargets` in the incremental path (`src/domain/graph/builder/incremental.ts`) does not pass `callerName`, so:

- The scoped-key fallback never fires during watch-mode incremental rebuilds
- A full build produces the correct graph; a subsequent incremental edit may regress

## Scope

Out of scope for PR #1368. Affects only the incremental rebuild path, not the full build path.

## Fix

In `incremental.ts`, locate the `resolveCallTargets` call(s) and pass the enclosing-function `callerName` (from `findCaller`) as the 6th argument, mirroring what the full-build path (`buildObjectRestParamPostPass` and `buildFileCallEdges`) already does.

carlos-alm:
## Summary

- `buildCallEdges` in the incremental rebuild path (`incremental.ts`) was missing two things needed for Phase 8.3f scoped-key lookup to work after a file edit:
  1. **TypeMap not seeded with scoped keys**: `objectRestParamBindings × paramBindings` were not cross-referenced to seed `callee::restName` entries (e.g. `f2::rest`) in the typeMap. Without this seeding the scoped-key fallback in `resolveByMethodOrGlobal` (`typeMap.get(callerName::receiver)`) has nothing to find.
  2. **`caller.callerName` not passed**: Even with scoped keys in the typeMap, the `${callerName}::${effectiveReceiver}` lookup never fired because `callerName` was not passed to `resolveCallTargets`.
- Also ports the same-class `this.method()` fallback and `definePropertyReceivers` fallback from the full-build path into `buildCallEdges`, completing incremental/full-build parity for these resolution patterns.

## Test plan

- [x] New `issue-1369` integration test: full build produces `f1→m1` and `f2→m2` (same fixture as #1358); subsequent incremental rebuild (file touch) preserves both edges with no cross-edges
- [x] All three tests (`#1336`, `#1358`, `#1369`) pass locally

Closes #1369


carlos-alm:
@greptileai

PR #1401, opened 2026-06-08
fix(incremental): port same-class this.method() and defineProperty fallbacks into buildCallEdges

carlos-alm:
## Summary

`incremental.ts:buildCallEdges` was missing two fallback resolution blocks that exist in `build-edges.ts:buildFileCallEdges`, causing the incremental (watch-mode) rebuild path to produce fewer call edges than a full rebuild.

- **Same-class `this.method()` fallback:** after `resolveCallTargets` returns empty for a `this`-receiver call, derive the enclosing class name from `callerName` (e.g. `Logger.info` → `Logger`) and retry with the qualified name `Logger._write` restricted to `kind === 'method'`. Mirrors the native engine's class-scoped symbol table behaviour.
- **Object.defineProperty accessor fallback:** when a function is registered via `Object.defineProperty(obj, "bar", { get: fn })`, `this` inside `fn` refers to `obj`. Resolve `this.X()` via `obj`'s typeMap entry or a same-file lookup — mirrors `buildDefinePropertyPostPass` in the native path.
- Also adds the missing `caller.callerName` argument to the `resolveCallTargets` call (the 6th parameter introduced by PR #1383).

## Test plan

- [x] All 633 existing tests pass (`npx vitest run`)
- [x] No new lint warnings introduced

Closes #1384

PR #1405, opened 2026-06-08
feat(resolver): resolve .call()/.apply() this-rebinding and add fun fixture (JS)

### LINKED ISSUE #1380: feat(resolver): resolve .bind(), .call(), .apply() dispatch and argument flow (JS)
## Source

Jelly micro-test benchmark (PR #1376). \`fun\` scores 29% (4/14), with 10 named edges missed covering bind/call/apply patterns.

### .bind() — result as new callable
\`\`\`js
var f = function() { return this.a; }.bind(x);
f()();   // ← missed: calling the bound function's return value
\`\`\`

### .call() — function called with explicit this
\`\`\`js
function foo(b) { return this(b); }
foo.call((a) => a, () => { ... })();  // ← missed: arrow arg flows to this
\`\`\`

### .apply() — function called with array of args
\`\`\`js
function bar(c) { return this(c); }
bar.apply((d) => d, arguments);  // ← missed: arguments elements flow to bar
bar.apply((d) => d, args);       // ← missed: rest param elements flow to bar
bar.apply((d) => d, [a]);        // ← missed: literal array elements flow to bar
\`\`\`

## Root cause

\`.bind()\`, \`.call()\`, and \`.apply()\` are meta-call operations that redirect \`this\` and/or arguments. Codegraph does not:
1. Track that the result of \`.bind(fn)\` is callable as \`fn\`
2. Resolve \`this\` inside a \`.call(ctx, ...)\` / \`.apply(ctx, ...)\` to \`ctx\`
3. Unpack the array argument of \`.apply(ctx, arr)\` to map elements to parameters

## What's needed

1. **bind result**: \`fn.bind(ctx)\` → emit synthetic callable that delegates to \`fn\`
2. **call/apply this-rebinding**: when a function calls \`this(x)\`, resolve \`this\` to the first arg of \`.call()\`/\`.apply()\`
3. **apply array unpacking**: \`fn.apply(ctx, [a, b])\` → unpack array elements as positional args

## Impact

\`fun\` Jelly micro-test: 10 of 14 named edges missed. These patterns appear in polyfills, utility libraries, and curried function pipelines.

carlos-alm:
## Summary

- Adds the Jelly micro-test `fun` fixture (`fun.js` + `expected-edges.json`) with corrected `<root>` source names for module-level edges; the 4 named edges (`baz/baz2/baz3/baz4 → bar`) score 100% recall
- Implements `.call()`/`.apply()` this-rebinding: when `fn.call(namedCtx, ...)` is seen, seeds `fn::this → namedCtx` in the PTS map so that `this()` calls inside `fn` resolve to `namedCtx`
- Fixes a long-standing false positive in `extractCallbackReferenceCalls`: identifier arguments to `.call()`/`.apply()`/`.bind()` were incorrectly emitted as dynamic callback calls; now they are skipped since this-context and delegated function arguments should not be attributed to the outer scope
- Adds `buildThisCallBindingsPtsPostPass` so the native engine benefits from this-rebinding via the same JS post-pass pattern used for `fnRefBindings` and `paramBindings`

## Test plan

- [ ] `npx vitest run tests/benchmarks/resolution/resolution-benchmark.test.ts` — JS benchmark 100% precision, all 171 tests pass
- [ ] `npx vitest run tests/benchmarks/resolution/jelly-micro.test.ts` — `fun` scores recall=100% (4/4), all 10 tests pass
- [ ] `npx vitest run tests/parsers/javascript.test.ts tests/unit/points-to.test.ts tests/engines/parity.test.ts` — passes
- [ ] `npm run lint` — no errors

Closes #1380


carlos-alm:
Addressed Greptile's review feedback:

**Comment 1 — Broad skip in `extractCallbackReferenceCalls`** (line 2354): Valid concern. The current skip is intentionally broad to prevent false positives for the this-rebinding use case (`fn.call(ctx)`). Narrowing it to only skip the first arg for `.call()`/`.apply()` while preserving subsequent callback args is a follow-up improvement tracked in #1406.

**Comment 2 — No-receiver class-scoped lookup JS false-positive risk** (lines 185-194): The JS benchmark maintains 100% precision after the receiver-guard removal. A dedicated fixture to make the regression boundary explicit has been filed as #1407.

**Comment 3 — Full role re-classification performance** (lines 282-283): Added a comment to `native-orchestrator.ts` documenting the O(all_nodes) cost and suggesting a two-pass strategy as a future optimization path if this becomes a bottleneck (commit `c708cfc`).

carlos-alm:
@greptileai

carlos-alm:
Pushed a perf optimization (commit e502d39) to address the benchmark regression caught by the regression guard:

**Pre-publish benchmark gate failure**: `Build ms/file: 18.7 → 33 (+76%, threshold 70%)` (wasm engine)

The original implementation added two separate full-AST recursive traversals (`extractThisCallsWalk` + `extractThisCallBindingsWalk`) to every JS/TS file parse. With the codebase having 600+ files, these two extra O(n) walks were measurable WASM build overhead.

Fix: 
- In the `extractSymbolsQuery` path (used by WASM): merged the two separate walks into a single `extractThisCallAndBindingsWalk` pass — halves the traversal cost.
- In the `extractSymbolsWalk` path: inlined both detections into the existing `handleCallExpr` handler, which is already called during the main traversal — eliminates both extra passes entirely.

Behavioral output is identical — verified by 104 JS parser tests + cross-engine parity tests all passing.

carlos-alm:
@greptileai

carlos-alm:
Merged origin/main to resolve conflicts — only change was removing a stale TODO comment from `tests/integration/prototype-method-resolution.test.ts` (the comment was already removed on main in a prior merge). No logic changes.

carlos-alm:
@greptileai

carlos-alm:
Merged origin/main to resolve conflicts with 5 commits that landed on main since the previous merge:

- feat(resolver): resolve super.method() dispatch via class expression + static block + field def (#1399)
- test(integration): pin prototype-method-resolution test to WASM engine (#1400)
- fix(incremental): port same-class this.method() and defineProperty fallbacks into buildCallEdges (#1401)
- fix(resolver): qualified callerName mismatch in class-scoped typeMap lookup (#1403)
- fix(incremental): seed callee::restName typeMap keys and pass callerName in buildCallEdges (#1404)

Conflict resolutions:
1. `edge_builder.rs`: PR removed the `call.receiver.is_some()` guard to enable no-receiver static sibling calls; main added a `compute_confidence >= 0.5` filter to the class-scoped lookup. Combined both changes — kept the guard removal and added the confidence filter.
2. `native-orchestrator.ts`: PR and main both moved to a full `classifyNodeRoles(db)` call; main simplified the `runPostNativeCha` return type from `Set<number>` to `number`. Adopted the simpler `number` return type and preserved the PR's performance cost comment.

carlos-alm:
@greptileai

Issue #1292, opened 2026-06-02
feat(resolver): phase 8.3d — object property write tracking in points-to analysis


carlos-alm:
## Context

Phase 8.3 (#1289) seeds the pts map from variable declarators (`const fn = handler`) and member-expression declarators (`const fn = obj.method`). It does not handle object property *writes*:

```js
const handlers = {};
handlers.auth = authMiddleware;   // property write — not currently tracked
handlers.log = logRequest;

router.use(handlers.auth);  // call to handlers.auth — pts map has no entry for it
```

The `handlers.auth` abstract location is never seeded, so `resolveViaPointsTo('auth', ...)` finds nothing and the edge to `authMiddleware` is dropped.

### Extractor changes

Walk assignment expressions (`assignment_expression`) in `extractTypeMapWalk`. When the LHS is a `member_expression` and the RHS is an identifier:

```
handlers.auth = authMiddleware
→ FnRefBinding { lhs: 'handlers.auth', rhs: 'authMiddleware' }
```

The field-based key `handlers.auth` is already the format the pts solver uses for member-expression constraints, so no solver changes are needed.

### Solver seeding

The pts map already handles composite keys like `handlers.auth` — if a binding `{ lhs: 'handlers.auth', rhs: 'authMiddleware' }` is recorded, propagation produces `pts('handlers.auth') = { 'authMiddleware' }`.

When a call like `router.use(handlers.auth)` is seen, `extractCallbackReferenceCalls` emits `{ name: 'auth', receiver: 'handlers', dynamic: true }`. The pts fallback currently only handles `!call.receiver` — this would need to be extended to also look up `handlers.auth` composite keys in the pts map when `call.receiver` is set.

### Scope

- Only simple `obj.prop = identifier` writes (not chained: `a.b.c = x`)
- Same file only for the first iteration
- Gate with `BUILTIN_GLOBALS` check on the object name (skip `console.x = y` etc.)

## Acceptance criteria

```js
const handlers = {};
handlers.auth = authMiddleware;
router.use(handlers.auth);  // edge: (caller) → authMiddleware
```

PR #1408, opened 2026-06-08
refactor(extractor): align typeMapWalk currentClass reset with returnTypeMapWalk

### LINKED ISSUE #1386: refactor(extractor): align extractTypeMapWalk currentClass reset with extractReturnTypeMapWalk pattern
## Background

`extractReturnTypeMapWalk.walk` in `src/extractors/javascript.ts` explicitly resets
`currentClass` to `null` when descending into `method_definition` and
`function_declaration` bodies (lines ~1363–1376). This prevents nested function
declarations inside a class from incorrectly inheriting the enclosing class context.

## Problem

The sibling `extractTypeMapWalk.walk` (introduced in PR #1382) threads `currentClass`
through all children uniformly without any reset at function/method boundaries
(line ~1669). For the current use case — only `this.prop = new Ctor()` reads
`currentClass` — this is **harmless and even correct**: a constructor assignment
inside a `method_definition` still belongs to the class. However, the divergence from
the established sibling pattern is a maintenance hazard.

If future code extends `extractTypeMapWalk.walk` to use `currentClass` for something
that *should* reset at function boundaries, it will silently inherit the wrong class
context.

## Fix

Follow `extractReturnTypeMapWalk`'s pattern: explicitly pass `null` for `currentClass`
when descending into `method_definition` and `function_declaration` bodies in
`extractTypeMapWalk.walk`, and add a comment explaining why the reset is **not**
applied for `class_declaration` children (where we want to propagate the class context
for constructor and method assignments).

## Related

- PR #1382 (introduced the asymmetry)

carlos-alm:
## Summary

- \`extractObjectRestParamBindingsWalk.walk\` now explicitly returns early with \`null\` \`currentClass\` when entering \`function_declaration\`, \`generator_function_declaration\`, and \`method_definition\` bodies
- This mirrors the established pattern in \`extractReturnTypeMapWalk\` (lines ~1431–1446)
- Adds a comment explaining why \`class_declaration\`/\`class_body\` nodes **do** propagate the class name (constructor/method assignments must be keyed as \`ClassName.prop\`)
- \`buildCallEdges\` same-file fallback lookup is narrowed to \`kind === 'function' || kind === 'method'\` to prevent spurious call edges to variables or class nodes that happen to share a name with a called function; arrow functions and function expressions are stored as kind \`'function'\` so they are not affected

**Behavior of \`extractObjectRestParamBindingsWalk\` unchanged** — the previous code already passed \`null\` implicitly (the \`childClass\` variable defaulted to \`null\` for all non-class nodes). The fix makes the intent explicit to prevent future extensions from silently inheriting wrong class context.

## Test plan

- [ ] \`npx vitest run tests/parsers/javascript.test.ts\` — all previously passing tests still pass
- [ ] \`npx vitest run tests/integration/\` — all previously passing tests still pass

Closes #1386


carlos-alm:
@greptileai

carlos-alm:
Two additional commits pushed to fix regressions introduced during earlier merge conflict resolution:

1. Restored `case 'class':` in `walkJavaScriptNode` — this was lost when resolving merge conflicts in 6e72569, breaking class expression extends-relationship extraction (e.g. `const Foo = class Child extends Parent { ... }` no longer recorded in `ctx.classes`).

2. Changed `handleStaticBlock` to emit `kind: 'method'` instead of `kind: 'function'` — static blocks are logically method-scoped (they run in class initialization context), and CHA dispatch requires a method-kind node to walk up the parent class for `super.method()` resolution.

Also updated the 'extracts static blocks as function definitions' test to expect `kind: 'method'`, consistent with the behavior change. All 113 JS parser tests and all CI checks now pass.

@greptileai

carlos-alm:
Greptile's concern about the unfiltered same-file fallback in the `definePropertyReceivers` block has been addressed. The `byNameAndFile(call.name, relPath)` lookup now filters to `kind === 'function' || kind === 'method'` to avoid matching unrelated variables or class nodes that happen to share the same name in the file.

The filter accepts both `'function'` and `'method'` (not just `'method'`) because the primary use case for this fallback is resolving free function targets — the fixture in issue #1292 (`handlers.auth = authMiddleware`, where `authMiddleware` is a top-level `function`) confirms this. Restricting to `'method'` only would break that case.

Merge conflict resolution also pushed (the duplicate blocks created by git's auto-merge of the identical fallback sections have been removed).

carlos-alm:
@greptileai

carlos-alm:
PR description updated to fix incorrect function name reference: the changed function is `extractObjectRestParamBindingsWalk.walk`, not `extractTypeMapWalk.walk`. Also added documentation of the `buildCallEdges` same-file fallback filter (the behavioral change that was previously unmentioned in the PR summary).

PR #1410, opened 2026-06-08
fix(jelly-micro): add missing lib companion files to client*/eval/require-extensions fixtures

### LINKED ISSUE #1391: fix(jelly-micro): add missing lib*.js companion files to client*/eval/require-extensions fixtures
## Problem

Several Jelly micro-test fixtures import companion library files that are not committed to the fixture directory:

- `client1–client6`, `client9`, `client-this`: require `lib1.js`–`lib5b.js` and `index.js`
- `eval/eval.js`: requires `./lib1`
- `require-extensions/require-extensions.js`: requires `./lib1.js`

**Current status**: Harmless because all affected fixtures have only anonymous/root-named edges (`<anon@…>` / `<root>`), so `jelly-micro.test.ts` never calls `buildGraph` on them (guarded by `namedExpected.length === 0` check).

**Latent risk**: If a future author adds named edges to any of these manifests, `buildGraph` will be invoked against a directory with unresolvable `require` targets, causing misleading results or failures.

## Fix

Fetch the missing companion lib files from the upstream Jelly repo and commit them alongside the existing fixture `.js` files. The upstream source for each is `github.com/cs-au-dk/jelly/blob/master/tests/micro/<lib-name>.js`.

Alternatively, update the `description` field in each manifest to note that `buildGraph` should only be run with the companion files present.

Discovered during PR #1376 review.

carlos-alm:
## Summary

- Fetched `lib1.js`–`lib6.js`, `lib5a.js`, `lib5b.js` from the upstream [Jelly micro tests](https://github.com/cs-au-dk/jelly/tree/master/tests/micro/) and committed each file in the fixture directories that `require()` it
- Affected fixtures: `client1`–`client6`, `eval`, `require-extensions` (each gets only the files it needs)
- `client9` and `client-this` require `'library'` (a non-local npm module name, not a relative path) — no local companion file exists in the upstream Jelly repo, and both fixtures have only anonymous/root edges so `buildGraph` is never called on them

## Why now

Currently harmless: all affected fixtures have only `<anon@…>` / `<root>` edges, so the test harness skips `buildGraph` on them. But if a future author adds named edges to any of these manifests, the missing `require()` targets would produce unresolvable imports and misleading benchmark results.

## Test plan

- [x] `npx vitest run tests/benchmarks/resolution/jelly-micro.test.ts` — 65 tests pass, no regressions

Closes #1391


carlos-alm:
@greptileai

PR #1417, opened 2026-06-08
fix(parity): resolve C# same-class bare static calls + confidence filter for static receiver fallback

### LINKED ISSUE #1402: feat(resolver): C# receiver-typed calls via var-declared instances (0/4 recall)
## Context

The C# resolution benchmark `receiver-typed` mode has 0/4 recall. When a local variable is declared with `var` and initialized via `new ClassName()`, the type is not inferred and calls on that variable are not resolved.

## Example

```csharp
var service = new UserService(repo);
service.AddUser(user);  // not resolved — typeMap has no entry for 'service'
```

## Root cause

In `src/extractors/csharp.ts`, `handleCSharpVarDecl` returns early when the type node is `var_keyword`:

```typescript
if (!typeNode || typeNode.type === 'var_keyword') return;
```

There is no fallback to inspect the variable initializer (e.g., `new UserService(repo)`) to extract the type name. The fix would be to check for an `object_creation_expression` initializer and extract the constructor type, similar to Phase 8.2 in the JS extractor.

## Expected edges missing

- `Program.Main → UserService.AddUser` (service.AddUser)
- `Program.Main → UserService.GetUser` (service.GetUser)
- `Program.Main → UserService.RemoveUser` (service.RemoveUser)
- `Program.RunWithValidation → UserService.AddUser` (service.AddUser)

## Related

This is a separate concern from #1385 (which fixed same-file static method calls).

### LINKED ISSUE #1416: fix(parity): native C# engine finds extra call edges not detected by WASM in build-parity test
## Problem

The build-parity integration test `tests/integration/build-parity.test.ts > Build parity: native vs WASM — csharp Repository` fails locally with the native engine finding 2 extra call edges that WASM doesn't produce.

## Details

Native engine finds these additional edges:
```
{ kind: 'calls', source_name: 'Validators.ValidateUser', target_name: 'Validators.IsValidEmail' }
{ kind: 'calls', source_name: 'Validators.ValidateUser', target_name: 'Validators.IsValidName' }
```

The roles also differ:
- Native: `Validators.IsValidName` → `dead-ffi`, `Validators.ValidateUser` → `core`
- WASM:   `Validators.IsValidName` → `leaf`,     `Validators.ValidateUser` → `utility`

## Impact

Pre-existing failure unrelated to current PR work. The WASM extractor is missing the call edges that the native engine correctly finds. Per project rules, this is a bug in the less-accurate engine (WASM), not an acceptable parity gap.

## Discovered in

Noticed while sweeping PR #1415 (`fix/inline-new-receiver-1396`). The PR only touches JavaScript extraction in `src/extractors/javascript.ts` — this failure pre-dates it.

carlos-alm:
## Summary

**Fix 1 — confidence filter on static receiver fallback (original)**
- The direct static receiver fallback in `resolveByMethodOrGlobal` (added in #1395) was the only speculative `lookup.byName()` call in that function without a `computeConfidence(relPath, n.file, null) >= 0.5` guard
- All other non-type-resolved resolution paths apply the confidence floor — this brings the static fallback in line with the rest

**Fix 2 — C# same-class bare static call resolution (Closes #1416)**
- Both WASM and native engines were missing call edges for bare static method calls within the same C# class — e.g. `IsValidEmail()` inside `Validators.ValidateUser` had no same-class fallback for no-receiver calls
- WASM (`build-edges.ts`): add bare-call same-class fallback after the existing `this.method()` fallback — when `targets` is empty and `call.receiver` is null, try `CallerClass.callName` in the same file
- Rust native (`edge_builder.rs`): mirror the same fallback in step 5 of `resolve_call_targets`
- Role parity (`native-orchestrator.ts`): Rust classifies roles before JS CHA/this-dispatch post-passes add edges (stale fan-out medians). After those post-passes insert new edges, run a full role re-classification so the final roles reflect the complete graph

**Fix 3 — C# `var`-declared instance type inference (Closes #1402)**
- `var service = new UserService(repo); service.AddUser(user)` was producing 0/4 recall on the receiver-typed benchmark — the typeMap had no entry for `service`
- Two AST facts that were wrong in the initial approach: (1) C# tree-sitter uses `implicit_type` not `var_keyword` for `var`, and (2) `object_creation_expression` is a direct child of `variable_declarator` with no `equals_value_clause` wrapper
- WASM (`src/extractors/csharp.ts`): `handleCSharpVarDecl` now handles `implicit_type` by walking declarator children for `object_creation_expression` and seeding typeMap at confidence 1.0
- Rust native (`crates/codegraph-core/src/extractors/csharp.rs`): identical logic via `find_child`

**Results:**
- C# `receiver-typed` recall: 0/4 → 4/4 (100%)
- C# `same-file` static recall: 0/2 → 2/2 (100%)
- Build-parity test: 8/8 pass (nodes, edges, roles, ast_nodes identical for WASM and native)
- Full test suite: 3022 passed, 0 regressions

## Test plan
- [x] `tests/integration/build-parity.test.ts` — 8/8 pass
- [x] `tests/benchmarks/resolution/resolution-benchmark.test.ts` — C# receiver-typed 4/4, same-file 2/2
- [x] Full test suite — all pass, 0 regressions


carlos-alm:
Addressed Greptile review feedback:

The `typeName`-resolved typed-method path at line 109 (`lookup.byName('TypeName.method')`) intentionally omits the `computeConfidence` guard. Here's why:

1. **The lookup is already fully qualified.** When `typeName` comes from `typeMap` (e.g. `var d = new Dog()` seeds `typeMap['d'] = 'Dog'`), the lookup is `Dog.bark` — a precise, class-qualified name. File-proximity scoring is a disambiguation heuristic for *name-only* lookups where multiple unrelated symbols share the same name. A qualified name like `Dog.bark` already uniquely identifies the target, so proximity scoring would only harm cross-module resolution (e.g. calling a method on a class from another directory).

2. **The `!typeName` path (line 132) is intentionally different.** This path uses `effectiveReceiver.method` where the receiver has no tracked type — the receiver IS being speculatively treated as a class name. Here, confidence filtering is essential to avoid false positives from unrelated same-named symbols across the codebase.

3. **The PR description was slightly imprecise.** Updated the description to clarify that the `typeName`-resolved path at line 109 and the inline-new-expression path (line 103) are both exempt from the confidence filter by design — they produce qualified names that already carry their own precision.

The PR description has been updated to note this distinction explicitly.

carlos-alm:
@greptileai

carlos-alm:
Addressed the redundant partial re-classification concern: removed both the CHA file-scoped partial pass and the thisDispatch file-scoped partial pass. The full `classifyNodeRoles(null)` added in this PR already subsumes both — running the partial passes first then the full pass was pure waste. Now only the single full pass runs when either post-pass adds edges.

carlos-alm:
@greptileai

carlos-alm:
Merge conflict with main resolved. The two conflicting files were:

- `crates/codegraph-core/src/edge_builder.rs`: took main's unified class-scoped lookup with confidence >= 0.5 filter (strictly better — covers both receiver and no-receiver calls, adds the confidence guard that was missing in the PR's version)
- `src/domain/graph/builder/stages/native-orchestrator.ts`: merged PR's single consolidated full re-classification design with main's `runPostNativeCha` returning a count (number) instead of Set — the condition is now `chaEdgeCount > 0 || thisDispatchTargetIds.size > 0` triggering the single `classifyNodeRoles(null)` pass

All tests pass (3017/3017). CI running.

carlos-alm:
Merge conflict with main resolved. The conflict was in `src/domain/graph/builder/stages/native-orchestrator.ts`: origin/main added verbose comments on the CHA block explaining the full-pass motivation; our PR's refactoring replaced them with a shorter comment. Kept our PR's concise comment since the motivation is now captured elsewhere in the block (which was restructured to consolidate both the CHA and this-dispatch re-classifications into one single full pass).

Rebuilt `dist/` from updated `src/` — a stale build was causing a false positive in the JS precision benchmark (`runCallThis → handler`) because the `.call()/.apply()/.bind()` guard added in #1405 hadn't been compiled yet.

All 3022 tests pass locally. CI running.

carlos-alm:
@greptileai

carlos-alm:
Fixed the namespace-aware class extraction issue raised by Greptile.

Both the `this.method()` fallback and the bare-call same-class fallback in `build-edges.ts` were using `indexOf('.')` which takes the first dot segment. For a caller like `MyNS.Validators.ValidateUser` this would yield `MyNS` instead of `Validators`, causing sibling calls to be silently missed.

Both are now aligned with `call-resolver.ts` which uses `lastIndexOf` + `prevDot` to isolate only the segment immediately before the method name — so `MyNS.Validators.ValidateUser` correctly yields `Validators`.

Commit: 9be5c71a

carlos-alm:
@greptileai

carlos-alm:
Merge conflicts with main resolved (two rounds):

**Round 1 — origin/main at 76907e66 (#1415):**
- `src/extractors/csharp.ts`: origin/main added `extractVarInitType` helper + `isVar` flag approach with confidence 0.9. This PR's HEAD used `implicit_type` directly with `findChild` at confidence 1.0. Kept PR's approach (correct C# grammar semantics — `implicit_type` is the right node type, `findChild` is more accurate than `equals_value_clause` traversal). Removed dead `extractVarInitType` function that origin/main had added.
- `crates/codegraph-core/src/extractors/csharp.rs`: Same as above for Rust mirror.
- Test update: `tests/parsers/csharp.test.ts` expected confidence 0.9 (from main), corrected to 1.0 to match PR's intentional change.

**Round 2 — origin/main at 5a06fa2d (#1422):**
- Same C# extractor conflicts as above (same resolution).
- Auto-merge of `call-resolver.ts` correctly restored `MODULE_SCOPED_BARE_CALL_EXTENSIONS` and `isModuleScopedLanguage` guard.
- **Found regression**: The bare-call fallback in `build-edges.ts` lacked the module-scoped language guard, causing JS precision failure (`Processor.run → Processor.flush` false positive from `class-scope.js` fixture). Fixed by exporting `isModuleScopedLanguage` from `call-resolver.ts` and gating the bare-call fallback in `build-edges.ts` on it.

All 3030 tests pass locally.

carlos-alm:
@greptileai

PR #1422, opened 2026-06-08
test(bench): add JS fixture for bare-call class-scoped lookup regression guard

### LINKED ISSUE #1407: test(resolver): add JS fixture for bare-call class-scoped lookup to prevent false-positive regression
## Problem

PR #1405 removes the `call.receiver` guard from the class-scoped name lookup in `call-resolver.ts` (and mirrored in the Rust `edge_builder.rs`). This is correct for C#/Java static siblings, but for JavaScript/TypeScript, bare `foo()` calls inside a class body are lexically scoped to the module — not the class.

If a local or cross-file function named `foo` does not appear in the lookup (e.g., dynamically imported or tree-shaken away), the call would incorrectly resolve to `ClassName.foo`.

The JS benchmark maintains 100% precision after this change, but does not exercise the specific combination: (bare call, no global definition reachable, same-name method exists in the class).

## Fix

Add a fixture covering:
```js
class Foo {
  bar() { helper(); }  // helper is only defined globally in another module, NOT as Foo.helper
  helper() { ... }     // Foo.helper exists — should NOT be resolved from bar()
}
```

With `expected-edges.json` asserting `bar → global.helper` (or no edge) but NOT `bar → Foo.helper`.

## Scope

Deferred from #1405 to keep the PR focused on the core this-rebinding feature.

carlos-alm:
## Summary

- Adds `class-scope.js` fixture to the JS resolution benchmark
- Adds `Processor.run → processData` to `expected-edges.json` as the positive recall case
- Updates the expected-edge count comment in the benchmark test (35 → 42 after multi-class.js +4, #1405 +2, this fixture +1)
- Gates the same-class fallback in `resolveByMethodOrGlobal` on language (JS/TS bare calls no longer fall through to class-method lookup)

## What this guards

In JS/TS, a bare `foo()` call inside a class method is lexically scoped to the **module**, not the class — there is no implicit `this` binding. If the `call.receiver` guard in `resolveByMethodOrGlobal` (`call-resolver.ts:229`) were ever removed, the resolver would incorrectly emit `Processor.run → Processor.flush` (a false positive).

The fixture:
```js
export function processData(x) { return x * 2; }

export class Processor {
  run(x) {
    processData(x); // same-file module-level — resolves correctly
    flush();        // bare call; no module-level 'flush' — must NOT → Processor.flush
  }
  flush() {}
}
```

The 1.0 precision floor on the JS fixture acts as the enforcement mechanism: any regression producing `Processor.run → Processor.flush` fails CI immediately.

## Test plan

- [x] `npx vitest run -t "javascript" tests/benchmarks/resolution/resolution-benchmark.test.ts` — JS precision 100%, recall 100% (42/42 edges), all 10 tests pass

Closes #1407

### DIFF
--- a/crates/codegraph-core/src/extractors/csharp.rs
+++ b/crates/codegraph-core/src/extractors/csharp.rs
@@ -438,9 +438,13 @@ fn extract_csharp_base_types(
 
 // ── Type map extraction ─────────────────────────────────────────────────────
 
+/// Extract the constructor type from a `var x = new Foo()` initializer.
 fn extract_var_init_type(declarator: &Node, source: &[u8]) -> Option<String> {
     for i in 0..declarator.child_count() {
         let Some(child) = declarator.child(i) else { continue };
+        // Defensive: handle object_creation_expression as a direct child of variable_declarator.
+        // The standard grammar always wraps it in equals_value_clause, but this guard is kept
+        // as a belt-and-suspenders fallback for edge cases or future grammar changes.
         if child.kind() == "object_creation_expression" {
             if let Some(t) = child.child_by_field_name("type") {
                 return extract_csharp_type_name(&t, source).map(|s| s.to_string());
@@ -471,32 +475,6 @@ fn extract_csharp_type_name<'a>(type_node: &Node<'a>, source: &'a [u8]) -> Optio
         }
         _ => None,
     }
-}
-
-/// Extract the constructor type from a `var x = new Foo()` initializer.
-fn extract_var_init_type<'a>(declarator: &Node<'a>, source: &'a [u8]) -> Option<&'a str> {
-    for i in 0..declarator.child_count() {
-        let Some(child) = declarator.child(i) else { continue };
-        // Defensive: handle object_creation_expression as a direct child of variable_declarator.
-        // The standard grammar always wraps it in equals_value_clause, but this guard is kept
-        // as a belt-and-suspenders fallback for edge cases or future grammar changes.
-        if child.kind() == "object_creation_expression" {
-            if let Some(t) = child.child_by_field_name("type") {
-                return extract_csharp_type_name(&t, source);
-            }
-        }
-        if child.kind() == "equals_value_clause" {
-            for j in 0..child.child_count() {
-                let Some(expr) = child.child(j) else { continue };
-                if expr.kind() == "object_creation_expression" {
-                    if let Some(t) = expr.child_by_field_name("type") {
-                        return extract_csharp_type_name(&t, source);
-                    }
-                }
-            }
-        }
-    }
-    None
 }
 
 fn match_csharp_type_map(node: &Node, source: &[u8], symbols: &mut FileSymbols, _depth: usize) {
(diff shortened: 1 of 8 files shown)


carlos-alm:
@greptileai

carlos-alm:
Merged origin/main into this branch. Resolved two conflicts:

- `src/infrastructure/native.ts`: kept main's version (the PR had an older `CODEGRAPH_NATIVE_ADDON_PATH` env var approach; main refactored to `NAPI_RS_NATIVE_LIBRARY_PATH` with 3-step load order). The PR never changed this file meaningfully so main's version is authoritative.
- `tests/benchmarks/resolution/resolution-benchmark.test.ts`: merged the comment histories — the PR branch and main each updated the edge-count comment independently. The resolved comment accurately traces: 35 → 39 (multi-class.js, #1382, +4) → 41 (call/apply this-rebinding, #1405, +2) → 42 (class-scope.js, #1407, +1).

**Note:** Local test runs show the fixture is detecting real bugs in the resolver:
- `Processor.run → Processor.flush` false positive — the bare-call same-class fallback is not gated on language. Filed as #1425.
- `runCallThis → handler` false positive from bind-call-apply resolution. Filed as #1426.

These bugs are masked in CI's pre-publish benchmark gate because it runs in artifact mode (pre-computed results). Once the fixtures are rebuilt locally or the artifact is regenerated, the precision gate will fail. The fixture PR is working as designed — it added the guard, and the guard found bugs.

carlos-alm:
@claude

carlos-alm:
Resolver fix committed (290008d): gated the same-class fallback in resolveByMethodOrGlobal on language to prevent JS/TS bare-call false positives. Both #1425 (Processor.run → Processor.flush) and #1426 (runCallThis → handler) are now fixed. Both issues closed.

The class-scope.js fixture worked as designed — it found the exact regression it was intended to guard against.

carlos-alm:
@greptileai

carlos-alm:
@claude

carlos-alm:
Fixed PR description edge count discrepancy (40 → 42) and updated the resolver line reference (189 → 229). Also merged origin/main (commit 906a8b02).

carlos-alm:
@greptileai

carlos-alm:
Addressed Greptile's two review issues:

**Issue 1 — Missing unit test for bare-call JS/TS guard**: Added 4 new unit tests to `tests/unit/call-resolver.test.ts` in commit `a3288058` covering: (a) bare call in .ts file does NOT resolve to same-class method, (b) bare call in .js file does NOT resolve to same-class method, (c) this.flush() in .ts DOES resolve (receiver present, not bare call), (d) bare call in .cs DOES resolve (C# is not module-scoped). All 10 unit tests pass.

**Issue 2 — Redundant `object_creation_expression` check**: Added a comment to both `src/extractors/csharp.ts` and `crates/codegraph-core/src/extractors/csharp.rs` explaining that the direct-child check is kept as a belt-and-suspenders fallback. The standard grammar wraps it in `equals_value_clause`, but removing the guard could break parsing on edge cases or future grammar versions. The comment makes the intent clear for future maintainers.

Merge conflicts with `origin/main` also resolved — the 3 new main commits (pts-param TECHNIQUE_MAP, per-fixture recall floors, inline-new receiver fix) are now merged in.

carlos-alm:
@greptileai

PR #1423, opened 2026-06-08
fix(native): promote this-dispatch scope tests from todo to active

### LINKED ISSUE #1413: fix(native): native engine does not emit Shape.describe → Shape.area this-dispatch edge
## Bug

The integration test `this-dispatch-scope.test.ts` line 65 (`'emits Shape.describe → Shape.area'`) fails for the native engine (v3.11.2). The native engine does not emit the expected `this.area()` dispatch edge inside `Shape.describe`.

The test runs for both `wasm` and `native` engines, but only WASM passes. The other three negative assertions are correctly marked as `.todo` for native (lines 79-83), but the positive assertion at line 65 is not, causing a CI failure.

Similarly, `prototype-method-resolution.test.ts` line 98-103 (`resolves (new Dog(...)).bark() inline-new receiver call to Dog.bark`) fails for the native engine.

## Steps to Reproduce
```
npx vitest run tests/integration/prototype-method-resolution.test.ts tests/integration/this-dispatch-scope.test.ts
```

## Expected
Both tests pass for native engine.

## Actual
Both fail — native engine does not emit the expected edges.

## Notes
These failures pre-exist in PR #1399 (feat/super-dispatch-1377) and are unrelated to the super.method() dispatch work. They should be fixed in the native Rust engine independently.

carlos-alm:
## Summary

- Removes the `if (engine === 'native')` guard that was skipping three negative assertions with `.todo` in `this-dispatch-scope.test.ts`
- The class-scoped this-dispatch fix landed in `edge_builder.rs` in #1343 (commit `16af0e0`), and the prototype extraction was added to the Rust extractor before v3.11.2; CI compiles the native addon from source on every PR so the binary used in tests already includes both fixes
- All four assertions now run for both `wasm` and `native` engines

Closes #1413

### DIFF
--- a/tests/integration/this-dispatch-scope.test.ts
+++ b/tests/integration/this-dispatch-scope.test.ts
@@ -72,47 +72,36 @@ describe.each(ENGINES)('this-dispatch scope (%s)', (engine) => {
     ).toBeDefined();
   });
 
-  // Native binary v3.11.2 does not include the edge_builder.rs fix for issue #1324 yet.
-  // These assertions are active for WASM and will be re-enabled for native once a new
-  // binary is published that includes the Rust fix.
-  if (engine === 'native') {
-    it.todo('does NOT emit Shape.describe → Calculator.area (native binary gap #1324)');
-    it.todo('does NOT emit Shape.describe → Formatter.area (native binary gap #1324)');
-    it.todo(
-      'does NOT emit Caller.run → Sibling.area (single-match false-positive, native binary gap #1324)',
+  it('does NOT emit Shape.describe → Calculator.area (unrelated class, same method name)', () => {
+    const edge = callEdges.find(
+      (e) => e.caller_name === 'Shape.describe' && e.callee_name === 'Calculator.area',
     );
-  } else {
-    it('does NOT emit Shape.describe → Calculator.area (unrelated class, same method name)', () => {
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed Claude's review:

- Root cause analysis confirmed: the `edge_builder.rs` fix (commit `16af0e0`, PR #1343) is verifiably present in the codebase at `edge_builder.rs:506-544`. The commit hash does resolve in git history (not a squash-merge artifact as noted).
- No code changes needed — the review confirmed this is a straightforward, low-risk test promotion.
- Rating: 4/5, APPROVE. No action items.

carlos-alm:
@greptileai

carlos-alm:
Fixed CI failure: the Windows CI was failing in `tests/search/embedding-regression.test.ts` because the HuggingFace model download timed out. The HF library throws a generic error that gets wrapped by `loadModel` as `EngineError` (code: `ENGINE_UNAVAILABLE`), losing the original `ETIMEDOUT` errno. The existing network-error guard only matched on raw errno codes, so the wrapped timeout error was re-thrown instead of gracefully skipping.

Fix: added `msg.toLowerCase().includes('timeout')` to the `isNetworkError` check, which catches HF's 'Request timeout error occurred' message regardless of how it's wrapped.

PR #1465, opened 2026-06-12
feat(parity): port the JS points-to solver to native and unify engine resolution

### LINKED ISSUE #1439: Parity: Rust edge_builder 4.5 direct qualified lookup lacks the >=0.5 confidence filter WASM applies (#1427)
## Context

#1427 added a \`computeConfidence >= 0.5\` filter to the WASM resolver's direct qualified-name fallback (\`resolveByMethodOrGlobal\`, src/domain/graph/builder/call-resolver.ts:136), noting it was the only resolution path without the filter.

The equivalent Rust path — section 4.5 "Direct qualified method lookup" in \`crates/codegraph-core/src/edge_builder.rs\` (~line 441) — was **not** given the same filter:

```rust
if type_lookup.is_none() && inline_new_type.is_none() {
    let qualified = format!("{}.{}", effective_receiver, call.name);
    let direct: Vec<&NodeInfo> = ctx.nodes_by_name
        .get(qualified.as_str())
        .map(|v| v.iter().filter(|n| n.kind == "method" || n.kind == "function").copied().collect())
        .unwrap_or_default();
    if !direct.is_empty() { return direct; }
}
```

Neighbouring sections (4.6 composite pts, prototype alias, accessor dispatch) all filter with \`compute_confidence(rel_path, &n.file, None) >= 0.5\`.

Also note the kind sets differ: WASM filters \`kind === 'method'\` only; Rust matches \`method | function\` ("to cover field-initializer synthetic defs"). Worth reconciling deliberately while fixing the confidence filter.

## Impact

Native engine can emit static-receiver edges to distant files that the WASM engine (correctly, post-#1427) rejects — a real parity divergence on the exact pattern #1427 fixed.

## Fix

Add the \`>= 0.5\` confidence filter to Rust 4.5, mirror the kind-set decision on both sides, and re-run the resolution benchmarks (expected-edges manifests may shift).

Found while working on #1432 (func-prop native extraction routes new method defs through this lookup).

### LINKED ISSUE #1452: bug: native build summary under-reports edge count (CHA post-pass edges excluded)
## Found during dogfooding v3.12.0

**Severity:** Low (cosmetic/logging bug — DB is correct, only the log message is wrong)
**Command:** `codegraph build --engine native --no-incremental`

## Reproduction

\`\`\`bash
codegraph build /path/to/repo --engine native --no-incremental
\`\`\`

Output:
```
[codegraph] Native build orchestrator completed: 22790 nodes, 47893 edges, 893 files
```

But the actual DB edge count is higher:
```sql
sqlite3 .codegraph/graph.db "SELECT COUNT(*) FROM edges;"
-- Returns: 48257 (not 47893)
```

## Root cause

The native orchestrator reports its edge count *before* the CHA (Class Hierarchy Analysis) post-pass runs. The CHA post-pass adds 364 edges tagged with `technique='cha'`, which are not included in the native orchestrator's build summary log.

The `this/super dispatch post-pass: inserted 45 edge(s)` log appears separately, but no log line accounts for the 364 CHA edges.

## Expected behavior

The final build summary line should reflect the actual total edges in the DB, including all post-pass edges. Or alternatively, the CHA post-pass should log its insertion count like the this/super dispatch post-pass does.

## Actual behavior

The native build summary under-reports the edge count by exactly the number of CHA edges inserted (364 in this test):
- Native build reports: 47893 edges
- CHA post-pass adds: 364 edges  
- Actual DB total: 48257 edges
- Missing from log: 364 edges (100% of CHA post-pass)

## Verified on

- v3.12.0, darwin-arm64, Node v24.10.0, native engine v3.12.0

### LINKED ISSUE #1453: bug: native engine misses 6 expected call edges in JavaScript fixture (85% vs WASM 100% recall)
## Found during dogfooding v3.12.0

**Severity:** High (resolution correctness gap between native and WASM)
**Command:** `codegraph build <js-fixture-dir> --engine native`

## Reproduction

\`\`\`bash
# Build the JavaScript resolution fixture with native engine
JS_FIXTURE=tests/benchmarks/resolution/fixtures/javascript

codegraph build $JS_FIXTURE --db /tmp/native-test.db --engine native
codegraph build $JS_FIXTURE --db /tmp/wasm-test.db --engine wasm
\`\`\`

Then compare against `expected-edges.json`:
- **Native**: 33/39 expected edges found (85% recall)
- **WASM**: 39/39 expected edges found (100% recall)

## Missing edges (native only)

1. `UserService.createUser` → `Logger.error` — `this.logger.error()` via constructor assignment (`this.logger = new Logger()`)
2. `UserService.createUser` → `Logger.info` — `this.logger.info()` via constructor assignment
3. `UserService.deleteUser` → `Logger.warn` — `this.logger.warn()` via constructor assignment
4. `runBind` → `greet` — `greetUser = greet.bind(user)` points-to tracking
5. `ClassA.runA` → `ServiceA.doA` — `this.service.doA()` with class-scoped typeMap key (`ClassA.service = new ServiceA()`)
6. `ClassB.runB` → `ServiceB.doB` — `this.service.doB()` with class-scoped typeMap key (`ClassB.service = new ServiceB()`)

## Root cause

The v3.12.0 CHANGELOG claims fixes for:
- Phase 8 constructor-assigned property types for receiver-typed resolution (#1314)
- `this.prop typeMap key scoped to enclosing class` (#1382)
- `Function.bind/call/apply receiver-typed resolution` (#1330)

These are working in the WASM engine but the native engine has not been updated with equivalent logic (or the parity gap was not fully closed).

## Files affected

- `tests/benchmarks/resolution/fixtures/javascript/service.js` — `UserService.createUser → Logger.error/info`
- `tests/benchmarks/resolution/fixtures/javascript/multi-class.js` — `ClassA.runA → ServiceA.doA`, `ClassB.runB → ServiceB.doB`
- `tests/benchmarks/resolution/fixtures/javascript/bind-call-apply.js` — `runBind → greet`

## Verified on

- v3.12.0, darwin-arm64, Node v24.10.0, native engine v3.12.0

carlos-alm:
## Summary

Brings the native (Rust) and WASM (TS) engines to full resolution parity on the JavaScript fixture set, taking the best of both implementations rather than restricting changes to one language. Stacked on #1463 (mirrored module layout).

**Native (Rust) — JS→native port:**
- The JS extractor now collects all eight points-to binding kinds (`param`, `this-call`, `array-element`, `spread-arg`, `for-of`, `array-callback`, `object-rest-param`, `object-prop`) and surfaces them on `FileSymbols`.
- `build_call_edges` runs the same fixed-point points-to solver as the WASM path: thisCall→fnRef conversion, the four-case key gate with receiver-key fallback, and hop-penalised alias edges with the pts upgrade mechanism.
- Inline-new receivers are normalized (`new A().t()` → receiver `"A"`) in `extract_receiver_name`, mirroring `extractReceiverName` in the TS extractor.
- Exact cross-file lookups apply the ≥0.5 confidence filter WASM uses.

**WASM/TS — native→JS backport and cleanup:**
- `NativeFileEntry` plumbs `params` and the eight binding arrays so the hybrid path feeds the native solver; `thisCallBindings` is serialized across the WASM worker boundary.
- The native engine's class-field type-annotation extraction is backported (`private repo: Repository` seeds typeMap `repo`/`this.repo`).
- The four JS pts post-passes that duplicated the native solver on the hybrid path are removed (~360 lines) — single source of truth for pts resolution per path.
- The native build summary and `build_meta` counts are now reported after the JS edge-writing post-passes, so they include CHA/this-dispatch edges.

## Verification

- WASM, full-native orchestrator, and hybrid builds produce **identical edge multisets** on the javascript fixture — 155 rows each, compared including `confidence` and `dynamic` flags.
- javascript fixture: 42/42 expected edges on both engines (was native 36/42, #1453); pts-javascript: 13/13 on both (was native 3/13).
- 392 Rust tests (22 new), 3043 JS tests, and the 176-test resolution benchmark across all 34 language fixtures are green.

Closes #1453
Closes #1452
Closes #1439

## Test plan

- [x] `cargo test` — 392 passed
- [x] Full vitest suite — 3043 passed
- [x] Resolution benchmark (all languages) — 176 passed
- [x] Engine edge-multiset diff (wasm vs native vs hybrid) — zero divergence


carlos-alm:
Addressed Claude's review feedback:
- **[Fixed]** `JS_BUILTIN_GLOBALS` was missing `process`, `window`, `document`, and `globalThis` compared to the TS `BUILTIN_GLOBALS` set and `is_js_builtin_global`. Added the four missing entries in commit `9075c122`. The constant now mirrors both the TS set and `is_js_builtin_global` exactly, preventing spurious pts edges from `process.exit()`, `window.open()`, etc.
- **[Noted/design choice]** `sort_targets_by_confidence` in native vs TS fallback: the dedup is on `(source_id, target_id)` which is unique regardless of insertion order, so this cannot affect the edge set. Only the stored confidence value could differ for a multi-target call, and the benchmark green confirms no edge-level regression. Filing a follow-up issue to align confidence sorting across both paths for correctness completeness.
- **[Out-of-scope]** The `#[allow(clippy::too_many_arguments)]` suppression follow-up refactor is tracked as a separate issue (will file below).

carlos-alm:
@greptileai

carlos-alm:
Addressed all remaining feedback from the initial Claude review:
- **[Fixed]** `JS_BUILTIN_GLOBALS` missing `process`/`window`/`document`/`globalThis` — fixed in commit `9075c122` (noted in prior reply).
- **[Fixed]** Added safety comment on the `max_idx as usize` cast explaining the `max_idx >= 0` guard makes the cast safe (commit `ad2d191e`).
- **[Tracked]** `sort_targets_by_confidence` confidence asymmetry → filed as issue #1476.
- **[Tracked]** `#[allow(clippy::too_many_arguments)]` follow-up refactor → filed as issue #1475.
- **[Out of scope]** Worker boundary unit test for `thisCallBindings` round-trip — this is a valid gap but the existing integration tests (3043 passing) exercise the full pipeline including the worker boundary. Filed as a follow-up tracking item in issue #1475 comments.

carlos-alm:
@claude

PR #1492, opened 2026-06-13
fix(perf): correct INLINE_BACKFILL_THRESHOLD docstring and raise threshold

### LINKED ISSUE #1435: Backfill worker-pool path: per-file cost ~3x the wasm-engine pipeline; INLINE_BACKFILL_THRESHOLD rationale outdated
## Measurements (v3.12.0 publish-gate diagnosis, codegraph corpus, M-series local)

- `parseFilesWasmForBackfill` pool path (18 JS/TS files, FULL_ANALYSIS): ~55–64ms/file
- Same files through the wasm-engine build pipeline (same pool): ~17.7ms/file
- Inline path (this-dispatch's hierarchy set): ~8–10ms/file

The INLINE_BACKFILL_THRESHOLD (16) docstring says above the threshold the pool's cost is 'amortized over enough parse work' — the measurements show per-file pool overhead doesn't amortize; it scales with file count. The 18-file proto backfill paying ~1s/build was a major component of the v3.12.0 native full-build regression (mitigated by symbolsOnly opts + extractor walk consolidation, but the per-file IPC overhead remains).

## Possible directions

- Profile what dominates pool per-file cost for backfill calls vs the engine pipeline (suspect: serialized ExtractorOutput transfer; symbolsOnly already trims it)
- Reconsider the threshold or make it risk-based (JS/TS required-tier grammars are not the #965 crash class; exotic grammars are)
- Correct the docstring either way so the next reader doesn't size batches against a wrong model

carlos-alm:
The `INLINE_BACKFILL_THRESHOLD` docstring claimed pool cost was "amortised over enough parse work" — measurements show IPC overhead scales linearly (~55–64ms/file pool vs ~8–10ms/file inline). The real motivation is grammar-crash safety for exotic parsers (#965); JS/TS/TSX (required-tier, used in all this-dispatch backfill calls) have never triggered the V8 fatal crash class and are safe to run inline.

## Changes

- **Docstring corrected**: explains the crash-safety rationale (exotic grammars can trigger uncatchable V8 fatal errors), clarifies that JS/TS/TSX are required-tier and safe inline, and accurately states that IPC overhead scales linearly rather than amortising
- **Threshold raised 16 → 32**: keeps typical this-dispatch batches (≤ 18 files on the codegraph corpus) on the inline fast path. Exotic-language drops (e.g. recurring HCL case is 4 files) are almost always well below 32 and also benefit from the inline path without meaningful crash risk increase

## Why 32

The primary hot caller (`runPostNativeThisDispatch`) measured 18 JS/TS files at 55–64ms/file through the pool vs 8–10ms/file inline — a 3–6x penalty. Raising to 32 covers that batch with headroom. Anything larger (truly exotic-language drops at scale) still routes through the pool for crash isolation.

Closes #1435

### DIFF
--- a/.claude/hooks/snapshot-pre-bash.sh
+++ b/.claude/hooks/snapshot-pre-bash.sh
@@ -23,18 +23,27 @@ if [ -z "$COMMAND" ]; then
   exit 0
 fi
 
-# Skip read-only commands that can never write files — reduces snapshot overhead
-# for the most common Bash calls (ls, cat, grep, git log, git status, etc.).
+# Skip commands that can NEVER write files — reduces snapshot overhead for the
+# most common read-only Bash calls.  Only include commands that have no
+# write-capable flags/modes at all.  Notably absent:
+#   - echo, printf  — write files via shell redirections (echo … > file)
+#   - find          — can write via -exec sed -i, -exec cp, -delete, etc.
+#   - awk           — can write via redirection or getline
+#   - node -e/-p    — can write files via the Node.js fs module
 # sed is intentionally NOT in this list because `sed -i` modifies files in-place.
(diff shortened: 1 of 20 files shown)


carlos-alm:
Fixed — updated the stale `parseFilesWasmForBackfill` docstring at line 1256 to replace the old 16-value rationale ("engine-parity drop sizes") with a pointer to `INLINE_BACKFILL_THRESHOLD` where the full rationale now lives (f90d4ba7).

carlos-alm:
@greptileai

Issue #1525, opened 2026-06-14
fix(erlang): 12 include_lib import tests fail after tree-sitter-erlang removal (#1478)


carlos-alm:
## Summary

12 tests in `tests/parsers/erlang.test.ts` fail locally after the malicious tree-sitter-erlang package was removed in #1478:

```
FAIL  tests/parsers/erlang.test.ts > Erlang parser > records -include_lib with kind "include_lib" so consumers resolve against OTP paths
AssertionError: expected undefined to be defined
```

All 12 failures are in the Erlang suite. The CI is currently green on main, suggesting these tests may be skipped in CI (WASM grammar unavailable) but run differently in some local environments.

## Root cause

The Erlang WASM grammar was removed from devDeps in #1478. The test suite has some guards for WASM availability, but the `include_lib` import tests appear to either run without the guard or exercise a path that fails when the grammar is missing rather than gracefully skipping.

## Affected tests

All 12 failing tests are in `tests/parsers/erlang.test.ts`, covering `-include_lib` import extraction and related record/module patterns.

## Pre-existing

This was discovered while adding unrelated test coverage in a separate PR — the failures exist on `origin/main` before any changes.

carlos-alm:
Investigated on `origin/main` (branch `fix/erlang-tests-1525`).

**All 14 Erlang tests — including all 12 `include_lib` and related tests — skip cleanly with 0 failures.**

The skip guards were already present in the same `cf18c027` commit that removed the malicious `tree-sitter-erlang` devDependency. Every test including the `include_lib` tests has:

```ts
if (!erlangAvailable) return ctx.skip();
```

Since the WASM grammar file is absent on a clean checkout, `parsers.get('erlang')` returns `undefined`, `erlangAvailable` is `false`, and all 14 tests skip rather than fail.

The adjacent PR #1528 addresses a related but distinct scenario (wrong-vendor WASM on a developer machine causing `erlangAvailable = true` with a grammar that emits wrong node types), but the failure mode described in this issue — tests failing rather than skipping on a normal checkout — does not reproduce.

Closing as resolved by `cf18c027`.

PR #1478, opened 2026-06-12
fix(deps): remove malicious tree-sitter-erlang, fix 3 moderate vulns

carlos-alm:
## Summary

- **[Critical]** Remove `tree-sitter-erlang` devDependency (GHSA-rphw-c8qj-jv84, CWE-506 malware). The npm advisory covers `*` (all versions by package name), so replacing the source is insufficient — the dep is dropped entirely. Erlang WASM support is unavailable in this configuration; see `scripts/build-wasm.ts` for rebuild instructions using the safe `github:the-mikedavis/tree-sitter-erlang` source when the WASM needs regenerating.
- **[Moderate]** `npm audit fix` bumps three transitive deps to patched versions: `hono` → 4.12.25 (via `@modelcontextprotocol/sdk`), `protobufjs` → 7.6.3 (via `@huggingface/transformers`), `qs` → 6.15.2 (via `@modelcontextprotocol/sdk` → `express`).

Result: `npm audit` reports **0 vulnerabilities**.

## Exposure assessment

The malicious package's `install` script (`node-gyp-build`) ran but produced no native addon — no `build/` directory was created, so the C source never executed. The package was never imported at runtime; it was only used to locate grammar source files for WASM compilation.

## Test plan

- [x] `npm audit` → 0 vulnerabilities
- [x] `scripts/check-grammar-versions.mjs` updated — `tree-sitter-erlang` removed from parity check list
- [x] `tests/parsers/erlang.test.ts` updated to skip gracefully when Erlang WASM is unavailable
- [x] `tests/benchmarks/regression-guard.test.ts` exempts Erlang precision/recall drop (100% → 0%) in `KNOWN_REGRESSIONS`
- [x] `package-lock.json` libc discriminators restored for `@optave/codegraph-linux-*` packages (stripped by npm on macOS during audit fix)
- [x] `scripts/build-wasm.ts` comment corrected — accurately describes that the Erlang WASM is not in the repository (grammars/ is gitignored) and provides full restoration steps

Closes #1477 is NOT intended — that issue tracks the pre-existing test failures, not this security fix.

PR #1526, opened 2026-06-14
test(typemap): cover class expression None path in field typeMap (#1500)

### LINKED ISSUE #1500: test: add coverage for class expression field annotation (None path) in enclosing_type_map_class
## Context

Identified during review of PR #1495.

In `src/extractors/javascript.ts` and `crates/codegraph-core/src/extractors/javascript.rs`, `handleFieldDefTypeMap` / the `field_definition` handler has two branches:

1. `Some(class_name)` — class declaration: seeds `ClassName.field @ 0.9` + bare keys @ 0.6
2. `None` — class expression: seeds bare keys only at full confidence (0.9)

Neither engine has a test covering the `None` (class expression) path: `const Foo = class { private repo: Repo; }` should produce `repo @ 0.9` and `this.repo @ 0.9` but NOT `Foo.repo`. This is new behavior introduced by PR #1495 and is currently untested in both TS and Rust.

carlos-alm:
## Summary

- Adds a WASM unit test in `tests/parsers/javascript.test.ts` verifying that `const Foo = class { private repo: Repo; }` seeds `repo @ 0.9` and `this.repo @ 0.9` — and does **not** produce a class-scoped `Foo.repo` key
- Adds a cross-engine parity guard in `tests/engines/parity.test.ts` asserting the same invariant for the native Rust engine (raw `typeMap` array format before `patchTypeMap` conversion)

The `None` branch in `handleFieldDefTypeMap` / `enclosing_type_map_class` was introduced in PR #1495 but had no test coverage in either engine.

## Test plan

- [x] `npx vitest run tests/parsers/javascript.test.ts tests/engines/parity.test.ts` — 166 passed, 8 skipped (native-only skips, expected)
- [x] `npx biome check` — clean
- [x] Pre-existing Erlang test failures (12 tests, unrelated to this PR, caused by tree-sitter-erlang removal in #1478) filed as issue #1525

Closes #1500

### DIFF
--- a/tests/engines/parity.test.ts
+++ b/tests/engines/parity.test.ts
@@ -858,6 +858,8 @@ const Foo = class {
     // Native: typeMap is a raw Vec<TypeMapEntry> array of { name, typeName, confidence }.
     // patchTypeMap() (called in production) converts it to a Map — here we assert the
     // raw extraction to verify the Rust None branch fires correctly.
+    // Binaries older than typeMap array support — skip gracefully.
+    if (raw?.typeMap === undefined) return;
     const entries = raw?.typeMap as Array<{ name: string; typeName: string; confidence: number }>;
     expect(Array.isArray(entries)).toBe(true);
     const repoEntry = entries?.find((e) => e.name === 'repo');
--- a/tests/engines/parity.test.ts
+++ b/tests/engines/parity.test.ts
@@ -833,6 +833,46 @@ const user = repo.findById('alice');
     });
   });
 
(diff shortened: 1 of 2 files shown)


carlos-alm:
Fixed — added `if (raw?.typeMap === undefined) return;` after the `nativeExtract` call, matching the pattern used by the `returnTypeMap` and `callAssignments` guards in the same file. Older binaries will now skip gracefully instead of producing a cryptic assertion failure.

carlos-alm:
@greptileai

PR #1527, opened 2026-06-14
fix(native): add first-write-wins dedup to Rust type-map entry helpers

### LINKED ISSUE #1501: fix: Rust push_type_map_entry lacks deduplication parity with TS setTypeMapEntry
## Context

Identified during review of PR #1495 (pre-existing issue, not introduced by that PR).

In `crates/codegraph-core/src/extractors/javascript.rs`, the `None` (class expression) branch of the field annotation handler calls `push_type_map_entry` which **always appends** without deduplication. The equivalent TS function `setTypeMapEntry` has **first-write-wins** semantics (won't overwrite a higher-confidence existing entry).

This means in the Rust engine, when multiple class expressions in a file define fields with the same name, duplicate `typeMap` entries can accumulate for the same bare key, unlike the TS engine. The edge builder may deduplicate by taking highest confidence, but the extraction behavior differs.

## Fix

`push_type_map_entry` in `helpers.rs` (or a new variant for the 0.9 path) should implement first-write-wins semantics to match `setTypeMapEntry`.

carlos-alm:
## Summary

- Introduce `set_type_map_entry` in `crates/codegraph-core/src/extractors/helpers.rs` — a direct Rust mirror of `setTypeMapEntry` in `src/extractors/helpers.ts`. If a key already exists with equal-or-higher confidence, the new entry is dropped; otherwise the old entry is replaced.
- Rewrite `push_type_map_entry` to delegate to `set_type_map_entry` so all existing call sites across every language extractor automatically gain deduplication.
- In the `field_definition` handler's `None` branch (class expressions), replace the two `push_type_map_entry` calls with explicit `set_type_map_entry` calls at confidence 0.9 — matching the TS `handleFieldDefTypeMap` path exactly.
- In the `Some(class_name)` branch, replace the two direct `symbols.type_map.push(TypeMapEntry { confidence: 0.6 })` calls with `set_type_map_entry` so the fallback bare keys are also deduplicated.

## Test plan

- `cargo check` in `crates/codegraph-core/` — clean compile
- `npm test` — 3048 tests pass; 12 pre-existing Erlang parser failures unaffected

Closes #1501

### DIFF
--- a/crates/codegraph-core/src/extractors/helpers.rs
+++ b/crates/codegraph-core/src/extractors/helpers.rs
@@ -958,22 +958,53 @@ pub fn extract_simple_parameters(
 
 // ── Type-map helpers ───────────────────────────────────────────────────────
 
-/// Record a parameter name → type binding in the type-map sink, using
-/// the default confidence of `0.9` shared by every Rust extractor.
-pub fn push_type_map_entry(
+/// Merge a type-map entry into the sink with first-write-wins semantics,
+/// matching `setTypeMapEntry` in `src/extractors/helpers.ts`.
+///
+/// If the key already exists with equal-or-higher confidence the new entry is
+/// silently dropped; otherwise the old entry is replaced.  This prevents
+/// duplicate entries from accumulating when the same bare key is written
+/// multiple times (e.g. two class expressions in one file that both define a
+/// field with the same name).
+///
+/// Mirrors `setTypeMapEntry` in `src/extractors/helpers.ts`.
+pub fn set_type_map_entry(
     symbols: &mut FileSymbols,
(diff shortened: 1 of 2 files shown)


carlos-alm:
Addressed both issues:
- **Inaccurate doc comment**: Updated `set_type_map_entry` doc from "first-write-wins" to "highest-confidence-wins", and clarified `push_type_map_entry` to "highest-confidence-wins (first-write-wins at uniform 0.9 confidence)".
- **O(n) linear scan**: Out of scope for this correctness fix — tracked in #1533 for a follow-up architectural change.

carlos-alm:
@greptileai

PR #1528, opened 2026-06-14
fix(test): smoke-test erlang grammar compatibility to prevent false failures

### LINKED ISSUE #1502: fix: erlang extractor - 12 parser tests failing (macros, includes)
## Summary

12 tests in `tests/parsers/erlang.test.ts` are currently failing. These tests cover:
- Parametric macro extraction (uppercase and lowercase)
- `-include` directive import handling
- `-include_lib` directive import handling

## Failing tests

- `extracts uppercase parametric macro names`
- `extracts lowercase parametric macro names without mislabeling on argument vars`
- `records -include with kind "include" so consumers resolve locally`
- `records -include_lib with kind "include_lib" so consumers resolve against OTP paths`
- (8 additional related failures)

## Notes

These failures exist on `fix/enclosing-caller-attribution-1468` (the base for current work) and are unrelated to the Java interface dispatch fix in PR for #1469. The Erlang extractor likely needs updates to its macro and include handling.

### LINKED ISSUE #1504: fix: Erlang parser tests fail after tree-sitter-erlang removal
After `cf18c027` removed the malicious `tree-sitter-erlang` devDependency, all Erlang parser tests in `tests/parsers/erlang.test.ts` fail because the WASM grammar is no longer available.

12 tests fail with errors like "expected [] to deep equally contain..." — the parser returns empty results without the grammar.

Needs either:
- A replacement clean Erlang WASM grammar
- Or the Erlang tests should be skipped/marked pending when the grammar is absent (matching the graceful-degradation pattern for non-required parsers)

Discovered during parity work on #1470.

carlos-alm:
## Problem

After `cf18c027` removed the malicious `tree-sitter-erlang` devDependency, skip guards were added to `tests/parsers/erlang.test.ts` so tests skip when the grammar is absent. However, the guard only checked `!!parsers.get('erlang')` — whether the WASM *loaded*, not whether it produces the expected node types.

On developer machines that ran `npm install` before the devDep removal, a stale `grammars/tree-sitter-erlang.wasm` from a different grammar vendor (e.g. enolib/tree-sitter-erlang) remains on disk. That WASM loads successfully, so `erlangAvailable = true`, but it emits generic `attribute` nodes instead of the `module_attribute`/`fun_decl`/`pp_define`/`pp_include`/etc. nodes that the WhatsApp-grammar extractor expects. Every extraction returns empty, and all 12 content tests fail rather than skip.

## Fix

Add a grammar compatibility probe in `beforeAll`: parse the minimal snippet `-module(probe).` and run `extractErlangSymbols` against the result. If a `module` definition named `probe` is found, the grammar is the right variant and tests run. If not — whether the file is absent, fails to load, or is the wrong vendor — `erlangAvailable` is set to `false` and all 14 tests skip cleanly.

This is robust against:
- Grammar file absent (fresh CI checkout after devDep removal)
- Grammar file present but from a different vendor (stale developer machine)
- Any future grammar load error

## Testing

Verified three scenarios locally:
- Grammar absent → 14 skipped
- Wrong-vendor WASM present → 14 skipped (smoke-test detects mismatch)
- Correct WhatsApp WASM present → tests run and pass

Full suite: 180 passed, 1 skipped (the Erlang file), 0 failures.

Closes #1502
Closes #1504

### DIFF
--- a/tests/parsers/erlang.test.ts
+++ b/tests/parsers/erlang.test.ts
@@ -9,7 +9,23 @@ describe('Erlang parser', () => {
 
   beforeAll(async () => {
     parsers = await createParsers();
-    erlangAvailable = !!parsers.get('erlang');
+    if (!parsers.get('erlang')) {
+      erlangAvailable = false;
+      return;
+    }
+    // Smoke-test: verify the loaded grammar is the expected WhatsApp/tree-sitter-erlang
+    // variant whose AST uses specific node types (module_attribute, fun_decl, etc.).
+    // A stale WASM from a different vendor (e.g. enolib/tree-sitter-erlang) loads
+    // successfully but produces generic `attribute` nodes, causing all extractions to
+    // return empty. Treat that as "unavailable" so tests skip rather than fail.
+    try {
+      const parser = parsers.get('erlang');
+      const tree = parser.parse('-module(probe).');
+      const result = extractErlangSymbols(tree, 'probe.erl');
+      erlangAvailable = result.definitions.some((d) => d.name === 'probe' && d.kind === 'module');
+    } catch {
+      erlangAvailable = false;
+    }
   });
 
   function parseErlang(code) {
--- a/tests/parsers/erlang.test.ts
+++ b/tests/parsers/erlang.test.ts
@@ -9,7 +9,8 @@ describe('Erlang parser', () => {
 
   beforeAll(async () => {
     parsers = await createParsers();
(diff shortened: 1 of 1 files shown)


carlos-alm:
Fixed — captured `parsers.get('erlang')` into `erlangParser` once at the top of `beforeAll` and reused it in the smoke-test block, eliminating the redundant second call.

carlos-alm:
@greptileai

PR #1486, opened 2026-06-13
fix(wasm): align sort_targets_by_confidence with native engine

carlos-alm:
The native engine sorts call targets by confidence before emitting edges; the WASM/TS engine did not. For multi-target calls with duplicate `(source_id, target_id)` pairs, the stored confidence could differ between engines depending on insertion order.

Adds equivalent descending-confidence sort to `buildFileCallEdges` in `build-edges.ts`, matching the native sort order and eliminating the latent parity gap.

Closes #1476

PR #1530, opened 2026-06-14
test(wasm): add regression test for confidence-sorted dedup in build-edges

### LINKED ISSUE #1519: test(build-edges): add regression test for confidence-sorted dedup in WASM engine
## Context

In PR #1487 (`fix(bench): add warmup runs to incremental tiers in benchmark.ts`), a confidence sort was added to `src/domain/graph/builder/stages/build-edges.ts:1115-1121` to align the WASM engine's edge dedup behavior with the native engine's `sort_targets_by_confidence` call.

## Problem

There is no regression test covering this ordering guarantee. A future refactor could silently revert the sort and the WASM engine would again emit the wrong (lower-confidence) edge when duplicate `(source_id, target_id)` pairs appear in the targets array.

## Task

Add a unit or integration test to `tests/` that:
1. Sets up a call site with multiple resolution candidates for the same `(source, target)` pair where the candidates have different confidence scores.
2. Asserts that the emitted edge corresponds to the highest-confidence target — not the first one encountered.

This test should cover the WASM engine path (`build-edges.ts`) and run in CI to prevent the sort from being silently removed.

## Reference

- Changed file: `src/domain/graph/builder/stages/build-edges.ts` lines 1115-1121
- Related native implementation: `sort_targets_by_confidence` in `build_edges.rs`
- Raised in code review: PR #1487

carlos-alm:
## Summary

- Adds `tests/integration/issue-1519-confidence-sorted-dedup.test.ts` with 6 regression tests covering the confidence-sorted dedup guarantee added in PR #1486
- **Multi-target byName suite**: verifies that when `byName('helper')` returns two nodes from different directories, each edge is stored with the correct confidence score reflecting actual file proximity (same-dir → ≥ 0.7, different-dir → < 0.7), and the near-target edge is never absent
- **pts alias sort suite**: verifies that the Phase 8.3 pts alias loop emits the highest-confidence candidate (local definition at confidence 1.0 before penalty) rather than a far definition (0.3 before penalty)

Both suites run the WASM engine only (`engine: 'wasm'`), scoped to the code path where the sorts live (`build-edges.ts` lines 1126-1132 and 1215-1222).

A future refactor that removes the sorts would cause the confidence assertions to fail if insertion order puts a low-confidence target first.

## Test plan

- [x] `npx vitest run tests/integration/issue-1519-confidence-sorted-dedup.test.ts` — 6/6 pass
- [x] `npm test` — 3052/3052 pass, 181/182 files pass (1 pre-existing skip)

Closes #1519

### DIFF
--- a/tests/integration/issue-1519-confidence-sorted-dedup.test.ts
+++ b/tests/integration/issue-1519-confidence-sorted-dedup.test.ts
@@ -1,36 +1,33 @@
 /**
- * Regression test for #1519: confidence-sorted dedup in WASM build-edges stage.
+ * Confidence-value sanity tests for the sorted multi-target resolution path (#1519).
  *
  * PR #1486 added a descending-confidence sort to the `targets` array in
  * buildFileCallEdges (build-edges.ts lines 1126-1132) before the emit loop,
- * and matching sorts to the Phase 8.3 and 8.3f pts alias loops.
+ * and matching sorts to the Phase 8.3 pts alias loops.
  *
- * The sort guarantees that when `targets` contains multiple entries for the same
- * call site — whether pointing to different nodes or (in edge cases) the same
- * node via different resolution paths — the highest-confidence candidate is
- * processed first by the `seenCallEdges` / `ptsEdgeRows` dedup guard. Without the
- * sort, insertion order of nodes in the DB could cause the lower-confidence
- * candidate to win the dedup slot.
+ * WHAT THESE TESTS GUARD
+ * ──────────────────────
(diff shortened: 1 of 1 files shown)


carlos-alm:
Addressed — the file header now clearly distinguishes what these tests DO guard (confidence scoring: near targets consistently score higher than far targets, both edges are emitted) from what they DON'T guard (same-node-ID dedup-winning when the same node appears twice via different resolution paths). The missing dedup guard is tracked in #1547.

carlos-alm:
@greptileai

PR #1531, opened 2026-06-14
test(bench): document CI variance data for No-op and 1-file rebuild thresholds

### LINKED ISSUE #1516: test(bench): document CI variance data for regression guard threshold widening (jelly-micro PR #1509)
## Context

In PR #1509 (fix(wasm): port missing node extractions to JS extractor), `tests/benchmarks/regression-guard.test.ts` widened the `1-file rebuild` and `No-op rebuild` thresholds from 25%→50% and CANARY mode to 100%.

The justification is empirical CI variance on sub-30ms metrics (±10ms CI jitter ≈ ±50%+ relative), and `MIN_ABSOLUTE_DELTA` provides a floor. However, the PR does not include CI data showing the measured variance.

## Action

- Attach a CI run showing the variance on `1-file rebuild`, `No-op rebuild`, and CANARY metrics that motivated the threshold change.
- If the empirical data supports the change, add a comment in the test file citing the CI run ID.
- If the data does not support it, revert to 25% and investigate the root cause of the instability.

carlos-alm:
## Summary

The `NOISY_METRICS` set in `regression-guard.test.ts` contains three metrics that get the 50% tolerance instead of the default 25%. The `fnDeps depth 1` entry already had concrete CI run citations and baseline→measured numbers. `No-op rebuild` and `1-file rebuild` were only documented by cross-reference ("same pattern as…").

This PR adds explicit CI run evidence for both metrics in the `NOISY_METRICS` docstring, pulled from the `KNOWN_REGRESSIONS` comments already in the same file:

- **No-op rebuild**: four CI runs across v3.11.0–3.12.0 showing +60–396% measured variance on 15–30ms baselines (runs 26426483639, 26792023287, 27457266151, 27455727444)
- **1-file rebuild**: three CI runs across v3.11.0–3.12.0 showing +52–155% on 64–115ms baselines (runs 26706695868, 26793082961, publish gate), including the publish-gate cross-check where the incremental suite's identical metric passed in the same run

Also explains why 50% is the right tolerance — a genuine regression (consistent across runs) would land much higher than the ±20–50ms runner-load spikes documented here.

## Test plan

- [x] `npm test` passes (3046 tests, 0 failures)
- [x] No logic changes — comment-only diff

Closes #1516

### DIFF
--- a/tests/benchmarks/regression-guard.test.ts
+++ b/tests/benchmarks/regression-guard.test.ts
@@ -65,6 +65,37 @@ const NOISY_METRIC_THRESHOLD = BENCH_CANARY ? 1.0 : 0.5;
  * only when its baseline is consistently sub-30ms and CI variance has been
  * empirically shown to exceed 25%.
  *
+ * - `No-op rebuild`: native baselines across releases span 15–30ms
+ *   (3.11.0: 15ms, 3.11.2: 19–25ms, 3.12.0: 23–30ms depending on suite).
+ *   Shared-runner jitter of ±10ms routinely lands +60–109% on these numbers:
+ *     - 3.11.0 baseline 18ms → 29ms (+61%) — run 26426483639
+ *     - 3.11.2 baseline 25ms → 45ms (+80%) and 19ms → 37ms (+95%)
+ *       — run 26792023287 (docs-only PR #1282, zero hot-path changes)
+ *     - 3.12.0 baseline 30ms → 48ms (+60%) and 23ms → 48ms (+109%)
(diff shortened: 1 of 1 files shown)


carlos-alm:
Fixed — updated the `NOISY_METRIC_THRESHOLD` docstring, the `NOISY_METRICS` admission criterion, and two inline references to remove the stale "sub-30ms" label. The criterion now reads: add a metric when CI variance exceeds 25% due to runner jitter on short-ish sub-200ms baselines. All four occurrences of the contradictory phrasing were updated.

carlos-alm:
@greptileai

PR #1532, opened 2026-06-14
fix(extractor): strip brackets from computed string-key method names

### LINKED ISSUE #1517: fix(wasm): verify computed-property-name method resolution at call sites
## Context

PR #1509 added extraction of computed class method names (e.g., `['property7']`, `['generator10']`). These methods are now stored with brackets as part of the name (e.g., the node name includes the full computed expression text).

## Problem

There is no test asserting that the resolver can actually *find* these methods when called at a call site. If caller code references a method via `obj[key]()` or `obj.property7()`, the current name representation may cause resolution failures.

This was a pre-existing design choice in the native engine (not introduced in #1509), but the new WASM extraction makes it more visible and more likely to surface.

## Action

- Add a resolution test for computed property name call sites.
- Verify that the parity benchmark correctly covers computed-property method calls, not just extraction.
- If the name representation causes silent resolution failures, align the representation across both engines.

carlos-alm:
## Summary

- Computed class/object-literal methods like `['myMethod']()` were stored with the full bracket-quoted text as their name (`['myMethod']`), while call sites using `obj.myMethod()` look up the plain name `myMethod` — causing silent resolution failures with no call edge created
- Fix: all four WASM extraction paths that produce method names now strip brackets and quotes from string-literal computed keys; non-string keys like `[Symbol.iterator]` are skipped (not resolvable at dot-notation call sites)
- Same fix applied to native Rust `handle_method_def` to maintain engine parity

**Affected paths (WASM `src/extractors/javascript.ts`):**
- `handleMethodDef` — walk path (primary extraction used by unit tests and integration)
- `handleMethodCapture` — query path (used by wasm-worker)
- `extractObjectLiteralFunctions` — var-bound object literals (`const obj = { ['m']() {} }`)
- `runContextCollectorWalk` funcStack push — PTS/spread-for-of caller tracking

## Test plan

- Updated 5 existing parser-level extraction tests in `tests/parsers/javascript.test.ts` (were asserting old bracketed name form) to assert plain names
- Added 3 new parser-level tests: class method case, `Symbol.iterator` skip, no-brackets guard
- Added `tests/integration/issue-1517-computed-prop-resolution.test.ts`: 7 assertions covering node storage (no brackets), class method resolution, object-literal method resolution, and non-regression for regular methods
- `npm test`: 181/182 test files pass (1 pre-existing skip), 3056 tests pass
- `cargo check`: Rust compiles cleanly
- `npm run lint`: no errors

Closes #1517

### DIFF
--- a/crates/codegraph-core/src/extractors/javascript.rs
+++ b/crates/codegraph-core/src/extractors/javascript.rs
@@ -954,7 +954,29 @@ fn handle_class_decl(node: &Node, source: &[u8], symbols: &mut FileSymbols) {
 
 fn handle_method_def(node: &Node, source: &[u8], symbols: &mut FileSymbols) {
     if let Some(name_node) = node.child_by_field_name("name") {
-        let method_name = node_text(&name_node, source);
+        // For computed property names (`['methodName']`), extract the inner string
+        // literal so the stored name matches the plain identifier used at call sites
+        // (e.g. `obj.methodName()`).  Non-string computed keys like `[Symbol.iterator]`
+        // cannot be resolved at dot-notation call sites, so we skip them.
+        let method_name_owned: String;
+        let method_name: &str = if name_node.kind() == "computed_property_name" {
+            // child(0)='[', child(1)=string literal, child(2)=']'
+            let inner = name_node.child(1);
+            match inner {
+                Some(inner) if inner.kind() == "string" => {
+                    let s = node_text(&inner, source);
+                    let stripped = s.strip_prefix('"').and_then(|s| s.strip_suffix('"'))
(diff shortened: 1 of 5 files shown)


carlos-alm:
Addressed both issues:
- **Missing string_fragment arm in Rust**: Updated `handle_method_def` to handle both `string` (via `extract_string_fragment` for quote-free content) and `string_fragment` (direct text) in the computed_property_name match, matching all four TypeScript paths.
- **No native integration test**: Added a `describe.skipIf(!isNativeAvailable())` block to the integration test with 3 parity assertions (node storage, no-bracket guard, call edge resolution) covering the native engine path.

carlos-alm:
@greptileai

PR #1534, opened 2026-06-14
fix(native): prefer bare name over qualified in span-tie caller attribution

### LINKED ISSUE #1510: fix: align receiver-callee-mixup attribution between wasm and native

carlos-alm:
## Summary

- In `find_enclosing_caller` (native Rust engine), when two callable definitions at the same line have an equal span, prefer the bare (unqualified) name over the dot-containing qualified name
- Object-literal methods are extracted twice: `o1.f(function)` from `extract_object_literal_functions` (emitted first in the Rust walk order) and `f(method)` from `handle_method_def` (emitted later); the WASM engine emits them in the opposite order, so WASM's strict-`<` tie-break naturally picked the bare name while native picked the qualified one
- Names with angle brackets (e.g. `B.<static:36:2>` static-block synthetics) are excluded from the bare-preference tie-break
- Both engines now attribute `f(method) → g(method)` for the `receiver-callee-mixup` jelly-micro fixture, matching the ground-truth expected-edges

## Test plan

- [ ] `npx vitest run tests/benchmarks/resolution/jelly-micro.test.ts` — 64 tests pass including `receiver-callee-mixup` at recall=100%
- [ ] `npm test` — all 180 test files pass (Erlang parser failures are pre-existing, unrelated to this change)
- [ ] Confirm native engine now produces `f(method) → g(method)` instead of `o1.f(function) → g(method)` for the fixture

Closes #1510

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
@@ -712,12 +712,14 @@ fn find_enclosing_caller<'a>(defs: &[DefWithId<'a>], call_line: u32, file_node_i
             let span = def.end_line.saturating_sub(def.line);
             if is_callable_kind(def.kind) {
                 // On a strict span improvement always take the new candidate.
-                // On a tie, prefer bare names over qualified names (dot-containing, no angle
-                // brackets) so native matches WASM: both pick `f(method)` over `o1.f(function)`
-                // when an object-literal method is extracted under both names at the same line.
+                // On a tie, prefer bare names over qualified names so native matches WASM:
+                // both pick `f(method)` over `o1.f(function)` when an object-literal method
+                // is extracted under both names at the same line. Synthetic angle-bracket
+                // nodes (e.g. `B.<static:36:2>`) are excluded on both sides of the comparison.
                 let is_improvement = span < fn_caller_span;
                 let is_tie_prefer_bare = span == fn_caller_span
                     && !def.name.contains('.')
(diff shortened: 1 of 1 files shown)


carlos-alm:
Fixed — added `&& !def.name.contains('<')` to the `is_tie_prefer_bare` condition, making the angle-bracket exclusion guard symmetric: both the incoming candidate and the existing candidate are now required to be free of synthetic angle-bracket names. Updated the comment to document that both sides are excluded.

carlos-alm:
@greptileai

PR #1536, opened 2026-06-14
fix(native): resolve this-dispatch in func-prop methods (#1512)

### LINKED ISSUE #1512: fix: native misses this-dispatch for function-as-object-property methods (this.js f.h→f.g)

carlos-alm:
## Summary

- `runPostNativeThisDispatch` had three bugs preventing `this.g()` inside `f.h = function(){}` (func-prop methods) from resolving on the native engine path
- The guard required `extends` edges before even starting — but func-prop `this`-dispatch uses the dot-prefix of the caller name (`f` from `f.h`) as the class prefix, needing no class inheritance at all
- Full-build file selection only unioned files touched by `extends` edges, so func-prop-only files were never re-parsed
- Empty `parents` map caused an early-return, even though `resolveThisDispatch` works correctly with an empty parents map (direct class-prefix lookup)

## Root Cause

`runPostNativeThisDispatch` was designed around class inheritance (`extends` edges) and never accounted for the func-prop pattern. `resolveThisDispatch` already handles it correctly: for caller `f.h`, it extracts class prefix `f`, then looks up `f.g` via `lookup.byName()` — no parents traversal needed. The post-pass just never gave it the right files or the chance to run.

## Fix (`src/domain/graph/builder/stages/native-orchestrator.ts`)

1. **Guard** — fire when `extends` edges exist OR when dot-named `method` nodes exist (`INSTR(name, '.') > 0`); skip the costly `extends` JOIN entirely when no `extends` edges are present
2. **parents map** — remove the `parents.size === 0` early return; an empty parents map is valid and correct for the pure func-prop case
3. **File selection** — add a third `UNION` arm to include files containing any dot-named `method` node

Also adds `this-dispatch-func-prop` to `TECHNIQUE_MAP` in `resolution-benchmark.test.ts` so the mode is correctly bucketed as `cha-rta`.

## Test plan

- `npm run lint` — clean (757 files, no fixes)
- `npm test` — 660 tests pass; the 2 failures are pre-existing from a concurrent open worktree for #1513 (untracked test file, unrelated)
- `jelly-micro/this` fixture (`f.h → f.g`, recall floor `1.0`) tests WASM-only and already passes; native path fix is validated by parity

Closes #1512

### DIFF
--- a/src/domain/graph/builder/stages/native-orchestrator.ts
+++ b/src/domain/graph/builder/stages/native-orchestrator.ts
@@ -701,8 +701,13 @@ const THIS_DISPATCH_EXTS = new Set(['.js', '.ts', '.tsx', '.jsx', '.mjs', '.cjs'
  * `this`/`super` receivers, then resolves them through the class hierarchy stored
  * in DB `extends` edges — mirroring what `buildChaPostPass` does on the WASM path.
  *
- * Only runs when `extends` edges exist in the DB; if there is no inheritance
- * hierarchy there is nothing to resolve via `this`/`super` dispatch.
+ * Also handles function-as-object-property methods (`f.h = function() { this.g() }`):
+ * these use `this` to reference sibling properties on the same object (`f`), so
+ * `resolveThisDispatch` resolves them by treating the dot-prefix of the caller name
+ * (`f` from `f.h`) as the class and looking up `f.g` directly — no `extends` edge needed.
+ *
+ * Runs when either `extends` edges exist (class inheritance) OR dot-named `method`
+ * nodes exist (func-prop assignments); skips only when neither is present.
  */
 async function runPostNativeThisDispatch(
   db: BetterSqlite3Database,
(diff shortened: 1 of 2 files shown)


carlos-alm:
Acknowledged — the `INSTR(name, '.') > 0` arm could include class-method files without extends edges in very large repos, partially diluting the re-parse optimization. This is a performance concern (not a correctness issue) — the DISTINCT and `THIS_DISPATCH_EXTS` filter still bound the set, and real-world impact depends on the ratio of class-only vs func-prop files. Tracked in #1550 for a follow-up that narrows the arm (e.g. via NOT EXISTS guard) after profiling on a large repo confirms the actual overhead.

carlos-alm:
@greptileai

Issue #1583, opened 2026-06-17
bug(roles): interface properties classified as dead-leaf when containing interface is exported and used


carlos-alm:
## Context
Discovered during Titan audit (phase: gauntlet, date: 2026-06-17).

## Description
Interface properties are classified as `dead-leaf` by `codegraph roles --role dead` even when the containing interface is exported and actively used. Example: `ExtractParametersOptions` in `src/extractors/helpers.ts` is exported and used on lines 395 and 416, but its property methods are flagged as dead-leaf.

The role classifier does not recognize TypeScript interface property usage sites as references — it only sees the interface symbol itself, not its member properties.

## Impact
- False positives in dead-code detection for any exported interface with property members
- Makes `codegraph roles --role dead` unreliable for TypeScript interface-heavy files
- Grind phase cannot safely remove symbols flagged by this pattern

## Additional Context
Observed for `ExtractParametersOptions` in `src/extractors/helpers.ts:356`. The interface is exported at the module boundary and used as a parameter type in two separate functions. The property `methods` inside the interface is flagged as `dead-leaf` despite being accessed by callers.

## Source
- **Titan phase:** gauntlet
- **Severity:** bug
- **Category:** codegraph

PR #1601, opened 2026-06-18
fix(roles): classify Commander.js execute/validate methods as entry, not dead-entry

### LINKED ISSUE #1585: bug(roles): Commander.js action callbacks (command.execute, command.validate) classified as dead-entry
## Context
Discovered during Titan audit (phase: sync, date: 2026-06-17).

## Description
`codegraph roles --role dead` returns 12,653 "dead" symbols when run on this codebase, but ~99% are false positives. The primary cause is the Commander.js dispatch pattern used in `src/cli/commands/*.ts`.

Every CLI command file exports an object like:
```ts
export const command = {
  execute(args, opts, ctx) { ... },
  validate(args) { ... }
}
```

These `execute` and `validate` methods are dispatched by the Commander.js CLI framework via dynamic property access at runtime. Codegraph cannot resolve this dynamic dispatch, so all `command.execute` and `command.validate` symbols across ~30 CLI command files are marked as `dead-entry`.

## Impact
- Dead-code cleanup phases (grind) cannot safely operate on this codebase
- 12,653 false positives drown out the ~100-200 genuinely dead symbols
- Role classifier output is not actionable for Commander.js-based CLIs

## Reproduction
```
codegraph roles --role dead -T --json | grep "command\."
```

## Source
- **Titan phase:** sync
- **Severity:** bug
- **Category:** codegraph

carlos-alm:
## Summary
- Methods named `execute` and `validate` in files matching `ENTRY_PATH_PATTERNS` (`cli/commands/`, `routes/`, `handlers/`, `mcp/`, `middleware/`) are well-known Commander.js dispatch entry points. They were previously classified as `dead-entry`, causing them to appear in `--role dead` output.
- This produced ~12,000+ false positives in dead-code analysis output, making `codegraph roles --role dead` unactionable on Commander.js-based CLIs.
- The fix adds `COMMANDER_DISPATCH_NAMES` (`execute`, `validate`) and promotes matching methods directly to `entry` when their file path confirms they are in a framework dispatch directory. A method named `execute` in a non-framework directory (e.g. `src/utils/`) is still classified as `dead-unresolved`.
- Both the TypeScript classifier (`src/graph/classifiers/roles.ts`) and the mirrored Rust classifier (`crates/codegraph-core/src/graph/classifiers/roles.rs`) are updated for engine parity.

## Test plan
- [ ] 2 new unit tests added to `tests/unit/roles.test.ts`:
  - `classifies Commander.js execute/validate methods in cli/commands/ as entry (#1585)` — verifies promotion to `entry`
  - `does not classify execute/validate as entry when not in a framework directory (#1585 boundary)` — verifies the name-only guard doesn't over-classify
- [ ] All 14 roles unit tests pass
- [ ] Lint clean (`npm run lint` passes)

Closes #1585


carlos-alm:
Both Greptile P2 findings addressed:

1. **Stale JSDoc on `hasActiveFileSiblings`** — updated in `src/graph/classifiers/roles.ts` to accurately document that annotation-only kinds = `TYPE_DEF_KINDS` ∪ {`constant`}, and explain why the circular-dependency concern requires excluding them from the active-file count.

2. **Inlined `annotation_only` slice in Rust `compute_active_files`** — extracted as a named module-level constant `ANNOTATION_ONLY_KINDS` in `crates/codegraph-core/src/graph/classifiers/roles.rs`, replacing the inline literal that duplicated `TYPE_DEF_KINDS + constant`. This stays in sync automatically when `TYPE_DEF_KINDS` changes.

Merge conflict in `tests/unit/roles.test.ts` also resolved: the new Commander.js entry tests (#1585) and the incremental-path exported-interface test added to main (#1583) are both kept.

@greptileai

PR #1602, opened 2026-06-18
fix(roles): classify methods and functions with active siblings as leaf

### LINKED ISSUE #1586: bug(roles): three untraced call-site patterns cause dead-unresolved false positives (handler-table, conditional-interface-dispatch, logical-or default)
## Context
Discovered during Titan audit (phase: grind, date: 2026-06-17).

## Description
Three call-site resolution patterns are not traced by the role classifier, causing false-positive dead-code detection:

### Pattern 1: Handler-table object property callbacks
```ts
const STATEMENT_HANDLERS = [
  { match: isReturnStatement, handle: handleReturn },
  { match: isThrowStatement, handle: handleThrow },
];
STATEMENT_HANDLERS.forEach(h => { if (h.match(node)) h.handle(node, ctx); });
```
`handleReturn` and `handleThrow` are marked `dead-unresolved` even though they are called via `h.handle(...)`. Codegraph cannot trace that `handle` property of the handler-table object references these functions.

### Pattern 2: Conditional interface method dispatch
```ts
if (v.enterFunction) v.enterFunction(fn, ctx);
if (v.enterNode) v.enterNode(node, ctx);
```
Methods implementing `Visitor` interface (`enterFunction`, `exitFunction`, `enterNode`, `exitNode`, `finish`) are marked `dead-unresolved` even though the visitor walker dispatches them via conditional property access. Affects all three visitor files: `cfg-visitor.ts`, `complexity-visitor.ts`, `dataflow-visitor.ts`.

### Pattern 3: Logical-or function defaults
```ts
const fn = options._fetchLatest || fetchLatestVersion;
```
`fetchLatestVersion` is marked `dead-unresolved` even though it's used as the fallback operand in a logical-or expression. Function references as fallback values in `||` expressions are not treated as call sites.

## Impact
All three patterns produce `dead-unresolved` roles for actively-used functions, making grind automation unreliable.

## Affected Files
- `src/ast-analysis/visitors/cfg-visitor.ts` (patterns 1 + 2)
- `src/ast-analysis/visitors/complexity-visitor.ts` (pattern 2)
- `src/ast-analysis/visitors/dataflow-visitor.ts` (pattern 2)
- `src/infrastructure/update-check.ts` (pattern 3)

## Source
- **Titan phase:** grind
- **Severity:** bug
- **Category:** codegraph

carlos-alm:
## Summary

- Extends the `hasActiveFileSiblings` heuristic in the role classifier to cover `method` and `function` kinds, eliminating three false-positive `dead-unresolved` patterns discovered during Titan audit phase (grind):

  **Pattern 2 — Interface dispatch via conditional property access**: Methods implementing interfaces like `Visitor` are dispatched via `if (v.enterFunction) v.enterFunction(fn, ctx)`. Codegraph resolves the call to the property accessor, not the concrete method implementation, so the method gets `fanIn=0`. Affected: `enterFunction`, `exitFunction`, `enterNode`, `exitNode`, `finish` in `cfg-visitor.ts`, `complexity-visitor.ts`, `dataflow-visitor.ts`.

  **Pattern 3 — Logical-or function defaults**: `const fn = options._fetchLatest || fetchLatestVersion` — a function reference as a fallback value is not a call site, so no call edge is produced. Affected: `fetchLatestVersion` in `update-check.ts`. A `fanOut > 0` guard ensures only non-trivial helpers (those that call something) are promoted, preserving detection of genuinely inert dead functions.

  **Pattern 1 — Handler-table object property callbacks** (partially covered): Functions stored as object property values and dispatched via `h.handle(...)` also benefit when the file has active siblings.

- Both the TypeScript classifier (`src/graph/classifiers/roles.ts` + `src/features/structure.ts`) and the mirrored Rust classifier (`crates/codegraph-core/src/graph/classifiers/roles.rs`) are updated symmetrically.

- Also fixes a stale test that expected `dead-entry` for `execute()` in CLI command files — corrected to `entry` by #1585 but the test was not updated.

## Test plan

- [x] All 45 `classifyRoles` unit tests pass (30 existing + 5 new + 1 updated for #1585 correction)
- [x] All 15 `classifyNodeRoles` integration tests pass
- [x] Rust code compiles cleanly
- [x] Lint passes (Biome)
- [x] New tests explicitly cover both the positive (leaf) and negative (dead-unresolved when no siblings) cases for each pattern

Closes #1586


carlos-alm:
**Addressed: P1 self-sibling false-negative in both TS and Rust classifiers**

Both bugs identified in the Greptile review have been fixed:

**P1 — `buildActiveFilesSet` / `src/features/structure.ts`**

The function now returns two sets instead of one:
- `activeFiles` (fan_in>0 || fan_out>0) — used for annotation-only kinds (struct/enum/trait/type/interface/record), which need a loose check since they have no callers by design
- `calledActiveFiles` (fan_in>0 only) — used for method/function kinds to prevent the self-sibling loop

A function or method with fanIn=0, fanOut>0 as the sole callable in its file no longer adds its own file to the active set, so it stays `dead-unresolved` instead of being promoted to `leaf`.

**P1 — `compute_active_files` / `crates/codegraph-core/src/graph/classifiers/roles.rs`**

Same fix mirrored symmetrically: `compute_active_files` now returns `(active_files, called_active_files)`. `classify_rows` receives both and uses `called_active_files` for the method/function `has_active_siblings` check.

**New boundary tests**

Two new integration-path tests added to `tests/unit/roles.test.ts`:
- `does not promote sole function with fanIn=0, fanOut>0 to leaf via self-sibling (#1586 boundary)`
- `does not promote sole method with fanIn=0, fanOut>0 to leaf via self-sibling (#1586 boundary)`

All 63 roles tests pass (unit + graph classifier + integration).

@greptileai

PR #1616, opened 2026-06-19
feat(dataflow): P4 incremental re-stitch on native engine path

### LINKED ISSUE #1614: feat(dataflow): P4 incremental re-stitch on native engine path
## Context

P6 (PR #1612) wired vertex extraction into the native bulk-insert fast path, but skipped P4 (incremental re-stitch of unchanged callers).

## What P4 does on the JS path

When file A (callee) changes and file B (caller) does not:
- A is re-parsed → A's vertices rebuilt  
- B is NOT in `fileSymbols` → B's arg_in edges (pointing to A's old param vertices) are deleted but never recreated
- `collectCallerStitchCandidates` re-reads B from disk using WASM parsers and rebuilds the stitch candidates

## Why P4 is skipped on native

`collectCallerStitchCandidates` requires WASM parsers (`parsers` + `getParserFn`), which are not initialized in the native fast path. Caller files not in `fileSymbols` have no `symbols.dataflow` available.

## Impact

On incremental native-engine builds where only callee files changed: arg_in inter-procedural edges for callers of changed functions will be missing until the next full rebuild.

## Possible fix

Option A: Initialize WASM parsers in the native path and call `collectCallerStitchCandidates` for unchanged callers.  
Option B: Request Rust to re-parse the identified caller files (requires new Rust API).  
Option C: Detect stale arg_in edges on query and trigger a partial rebuild.

The v19 migration sentinel ensures a one-time full rebuild, so this only affects subsequent incremental builds.

carlos-alm:
## Summary

Fixes #1614. When only a callee file changes on an incremental native-engine build, `runDataflowVertexPass` (P6) rebuilds the callee's vertices — but unchanged caller files are never re-parsed, leaving their `arg_in` edges pointing at the deleted old param vertices.

**`buildDataflowP4ForNative`** (new export in `features/dataflow.ts`) closes this gap:

1. Finds caller files that have `flows_to` edges targeting the changed functions
2. Parses them with WASM (lazy init — no parser startup cost unless needed)
3. Calls `buildInterproceduralStitch` with the resulting stitch candidates

Called from `runDataflowVertexPass` after the vertex pass, so arg_in edges are correct on every incremental native build — not just JS-engine builds.

**Guard**: skips when `changedFiles.length >= totalFilesInDb` (full build — all callers already in the changed set, no re-stitch needed).

This PR also includes the P6 commits that missed the main merge of PR #1612 (`ecd47b61`, `02e55f70`) plus the README update.

## Test plan

- [x] `npx vitest run tests/integration/dataflow*.test.ts` — **44 passing** (40 existing + 4 new P4-native)
- [x] New tests cover: arg_in creation, scope=inter, full-build guard
- [x] `npm run lint` — clean


carlos-alm:
Closing as superseded by #1615, which merged the same P4 re-stitch fix with additional improvements (FK constraint fix, full-build optimization, benchmark fixes). The only unique file here () tests an export () that was replaced by a better approach in #1615.

PR #1658, opened 2026-06-21
fix: make dataflow vertex write and inter-procedural stitch atomic

### LINKED ISSUE #1640: follow-up: make buildDataflowVerticesFromMap vertex write and inter-procedural stitch atomic
Deferred from PR #1635 review.

**Original reviewer comment:** https://github.com/optave/ops-codegraph-tool/pull/1635#issuecomment-4759903488

**Context:** `buildDataflowVerticesFromMap` commits vertex rows inside `tx()` then calls `buildInterproceduralStitch` outside that transaction. If the stitch throws or the process is killed between the two, the DB is left with `dataflow_vertices` rows but no `arg_in`/`return_out` edges. The same split exists in the pre-existing `buildDataflowEdges` P6 block — fixing it requires unifying both paths under a single transaction boundary, which is an architectural change beyond this PR's scope.

carlos-alm:
## Summary

- `buildDataflowVerticesFromMap` committed vertex rows inside `tx()` then called `buildInterproceduralStitch` outside that transaction boundary. The same split existed in the native P6 block of `buildDataflowEdges` and the JS fallback path.
- If the stitch threw or the process was killed between vertex writes and the stitch, the DB would be left with `dataflow_vertices` rows but no `arg_in`/`return_out` edges — an inconsistent half-written state with no way to detect or recover from it.
- The fix extracts the stitch body into `runInterproceduralStitch` (no transaction wrapper), enabling outer callers to call it directly from within their own transaction. `buildInterproceduralStitch` wraps it in its own transaction and is now used only by `buildDataflowP4ForNative`, which runs as a standalone operation. All three affected paths now perform vertex writes and inter-procedural stitch under a single `db.transaction()` boundary. The async P4 disk-read step (`collectCallerStitchCandidates`) is hoisted above the transaction open since better-sqlite3 transactions are synchronous.

## Test plan

- [x] `tests/integration/dataflow-vertices.test.ts` — P1 vertex creation + P2 arg_in/return_out inter-procedural edges
- [x] `tests/integration/dataflow.test.ts` — full dataflow edge insertion
- [x] `tests/integration/dataflow-p4-native.test.ts` — P4 native re-stitch (uses `buildDataflowP4ForNative` which still calls `buildInterproceduralStitch` as a standalone step)
- [x] `tests/integration/dataflow-incremental.test.ts` — incremental build dataflow correctness
- [x] `tests/engines/dataflow-parity.test.ts` — WASM/native engine parity
- [x] Full `npm test` — 3245 tests pass (3 pre-existing failures on `main` unrelated to this change)
- [x] `npm run lint` — clean

Closes #1640

### DIFF
--- a/src/features/dataflow.ts
+++ b/src/features/dataflow.ts
@@ -477,7 +477,15 @@ function buildDataflowVerticesAndEdges(
  *  - If B's summary shows B.param[j] reaches B's return: emits 'return_out'
  *    inter edge: B.return → A's capture local (if any)
  */
-function buildInterproceduralStitch(
+/**
+ * Core stitch logic — must be called inside an already-open transaction.
+ *
+ * Separated from buildInterproceduralStitch so callers that manage their own
+ * outer transaction (buildDataflowVerticesFromMap, buildDataflowEdges) can
+ * call this directly without starting a nested transaction, which better-sqlite3
+ * does not support.
+ */
+function runInterproceduralStitch(
   db: BetterSqlite3Database,
   candidates: StitchCandidate[],
   captures: ReturnCapture[],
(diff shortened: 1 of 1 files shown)


carlos-alm:
Addressed both Greptile findings:

**Issue 1 (P1) — Atomicity gap in `buildDataflowP4ForNative`:** The separate autocommit DELETE followed by `buildInterproceduralStitch` opening its own transaction has been replaced with a single `db.transaction()` that wraps both the DELETE and the `runInterproceduralStitch` call. A crash between the two steps can no longer leave `arg_in` edges permanently removed but never re-inserted.

**Issue 2 (P2) — Stale duplicate JSDoc above `runInterproceduralStitch`:** The old dangling JSDoc block (describing the original `buildInterproceduralStitch`) was removed, leaving only the correct JSDoc that describes `runInterproceduralStitch`.

As a side effect, `buildInterproceduralStitch` became dead code (its only call site was `buildDataflowP4ForNative`, now refactored) and was removed.

carlos-alm:
@greptileai

Issue #1687, opened 2026-06-21
fix(parity): jelly-micro bind.js: WASM emits dyn=1 for f.call/bind aliases, native emits dyn=0


carlos-alm:
## Summary

When running `node scripts/parity-compare.mjs --langs jelly-micro` with grammars properly available, the WASM engine produces `dyn=1` edges for `f.call`/`f.bind` patterns in `tests/benchmarks/resolution/fixtures/jelly-micro/bind/bind.js`, while the native engine produces `dyn=0`.

## Divergence

```
[edge] [calls] bind/bind.js:bind/bind.js(file) -> arguments/arguments.js:f(function) conf=0.5 dyn=0  wasm=0 native=1
[edge] [calls] bind/bind.js:bind/bind.js(file) -> arguments/arguments.js:f(function) conf=0.5 dyn=1  wasm=1 native=0
```

22 similar edge diffs across different target files.

## Root cause

The `bind/bind.js` fixture content:
```js
const f = function () {}.bind({});
f();
f.call({});
```

`extractCallbackReferenceCalls` on `f.call({})` emits `{ name: 'f', dynamic: true }` (no `dynamicKind`). This results in a pts-resolved edge with `dyn=1` being emitted before the direct `f()` call (which should produce `dyn=0`) can suppress it.

## Note

This divergence was previously masked because `bind.js` failed to parse in test environments where the JavaScript grammar file was not properly available. The issue surfaced when running parity compare with grammars properly available.

## Impact

Parity gate fails for `jelly-micro`. All other 41 fixtures pass.

PR #1697, opened 2026-06-22
fix(wasm): apply plain-identifier guard to query path for .call/.apply/.bind

### LINKED ISSUE #1694: fix(wasm): query path still emits dynamic=true for .call/.apply alias calls after #1687
## Bug

After PR #1687 fixed the walk extractor to emit \`dyn=0\` for \`f.call()\`/\`f.apply()\`/\`f.bind()\` on plain identifiers, the query path (\`extractSymbolsQuery\` / \`dispatchQueryMatch\`) was not updated.

The \`query-walk-parity.test.ts\` fixture \`JS — dynamic call patterns (.call/.apply)\` now diverges:

- Walk path: no \`dynamic\` flag on \`fn\`, \`obj\`, \`method\` (correct, fixed in #1687)
- Query path: still emits \`dynamic: true\` (stale)

## Reproduction

\`\`\`js
fn.call(null, arg);
obj.apply(undefined, args);
method.bind(ctx);
\`\`\`

Run: \`npx vitest run tests/engines/query-walk-parity.test.ts\`

Expected: query and walk results are identical (no \`dynamic\` flag on plain-identifier callee).
Got: query result has \`dynamic: true\` on all three calls.

## Root cause

The WASM extractor has two code paths (documented in MEMORY). The #1687 fix updated \`extractMemberExprCallInfo\` (walk path) but the query path (\`callfn_node\`/\`callmem_node\` branches in \`dispatchQueryMatch\`) still applies \`dynamic: true\` for `.call`/`.apply`/`.bind` member expressions without checking whether the object is a plain identifier.

## Fix

Apply the same plain-identifier guard added in #1687 to the \`dispatchQueryMatch\` query path in the WASM extractor.

carlos-alm:
Closes #1694

## Summary
- The walk path fix from #1687 (`extractMemberExprCallInfo`) excluded `.call`/`.apply`/`.bind` from `dynamic=true` when the object is a plain identifier.
- The query path (`dispatchQueryMatch`) was not updated — this PR applies the same guard there.
- Fixes the `JS — dynamic call patterns (.call/.apply)` divergence in `query-walk-parity.test.ts`.

### DIFF
--- a/src/extractors/javascript.ts
+++ b/src/extractors/javascript.ts
@@ -317,6 +317,9 @@ function dispatchQueryMatch(
     if (callfnInfo) calls.push(callfnInfo);
     calls.push(...extractCallbackReferenceCalls(c.callfn_node));
   } else if (c.callmem_node) {
+    // extractCallInfo → extractMemberExprCallInfo applies the plain-identifier guard for
+    // .call/.apply/.bind: when the object is a bare identifier (e.g. `fn.call(ctx)`),
+    // the call is emitted as static (no dynamic flag), matching the walk path and native engine.
     const callInfo = extractCallInfo(c.callmem_fn!, c.callmem_node);
     if (callInfo) calls.push(callInfo);
     const cbDef = extractCallbackDefinition(c.callmem_node, c.callmem_fn);


carlos-alm:
@greptileai

PR #1791, opened 2026-07-04
refactor: address warnings in benchmark tracer tooling and ast-analysis naming

carlos-alm:
## Summary
- Address warn-level GAUNTLET findings in the benchmark tracer tooling (`loader-hooks.mjs`, `lua-tracer.lua`, `native-tracer.sh`).
- Address warn-level naming/style findings in `ast-analysis` and `extractors/helpers.ts` (rename for clarity, un-export a type with zero external references).

## Titan Audit Context
- **Phase:** warning
- **Domain:** tests/tracer, ast-analysis, extractors
- **Commits:** 2
- **Depends on:** PR #6 (quality fixes — ast-analysis/presentation/extractors/cli)

## Changes
- `tests/benchmarks/resolution/tracer/{loader-hooks.mjs,lua-tracer.lua,native-tracer.sh}`
- `src/ast-analysis/visitors/cfg-shared.ts`, `src/extractors/helpers.ts`

## Note
The tracer-tooling commit in this PR was redone from scratch after a session interruption during the FORGE phase: an earlier attempt at the `native-tracer.sh` fix was left staged but uncommitted when the session crashed mid-edit. On resume, a fresh session diff-reviewed the staged work before continuing and found that draft fix actually increased the script's complexity rather than reducing it, so it was discarded and redone correctly (the commit included in this PR). No partial/broken state from the interruption was carried into this branch.

## Metrics Impact
Warn-level polish pass — no functional behavior change.

## Test plan
- [ ] CI passes (lint + build + tests)
- [ ] codegraph check --cycles --boundaries passes
- [ ] No new functions above complexity thresholds

### DIFF
--- a/src/ast-analysis/visitors/cfg-conditionals.ts
+++ b/src/ast-analysis/visitors/cfg-conditionals.ts
@@ -6,7 +6,7 @@ import type {
   LoopCtx,
   ProcessStatementsFn,
 } from './cfg-shared.js';
-import { getBodyStatements, isCaseNode, isIfNode, nn } from './cfg-shared.js';
+import { getBodyStatements, isCaseNode, isIfNode, requireNode } from './cfg-shared.js';
 
 /**
  * Create a branch block off `condBlock`, wire the `branchKind` edge into it,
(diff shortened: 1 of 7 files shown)


carlos-alm:
Addressed both findings from the Greptile review (commit 2ea50ef7):

1. **Unquoted glob expansion in `inject_trace_calls`** — switched `for srcfile in $glob_pattern` to `while IFS= read -r srcfile; do ... done < <(compgen -G "$glob_pattern" || true)`. Verified the original code silently skips every file when the glob's base path contains a space (word-splitting breaks the pattern before globbing), while `compgen -G` expands the pattern as a single argument and instruments correctly. Confirmed no behavior change on the existing swift/c/cpp fixtures (byte-identical trace output) and the zero-match case no longer risks tripping `set -e`.

2. **`maybe_*` helpers depend on implicit dynamic scope** — expanded the existing design-rationale comment above the helpers to spell out the concrete failure mode you described (an unset `decl_regex` from `maybe_inject_declaration` matching every line) so future maintainers immediately see the risk if the call chain invariant (`inject_trace_calls -> instrument_one_file -> maybe_*`) is ever broken.

Issue #1825, opened 2026-07-05
Renamed import used as a receiver (import { X as Y }; Y.method()) does not resolve — qualified-name fallback doesn't know about the rename


carlos-alm:
## Summary

#1730 fixed call-edge resolution for renamed imports used as a *bare* call (`import { collectFiles as collectFilesUtil } from './helpers'; collectFilesUtil()`). While verifying that fix I found a distinct, still-broken variant: a renamed import used as a **receiver** (`import { X as Y } from './mod'; Y.method()`) does not resolve at all — no `calls` edge is created.

This is empirically confirmed (not speculative) — see repro below with raw DB inspection.

## Repro

```js
// helpers.js
export const NamespaceObj = {
  doThing() { return 1; },
};

// consumer.js
import { NamespaceObj as NsAlias } from './helpers.js';
export function useReceiver() {
  return NsAlias.doThing();
}
```

After `codegraph build`, querying edges shows `useReceiver` has **no** outgoing `calls` edge to `doThing` / `NamespaceObj.doThing` at all — the call is silently dropped. (Verified with sqlite3 directly against `edges`/`nodes` tables — `useReceiver` produces zero rows in `edges`.)

## Root cause

This does **not** go through `resolveCallTargets`'s import-aware branch (the one #1730 fixed) — `doThing` itself was never imported, only `NsAlias` was, so `importedNames.get('doThing')` is never populated. The call falls through to `resolveByMethodOrGlobal` → `resolveByReceiver` (`src/domain/graph/resolver/strategy.ts`) → `resolveViaDirectQualifiedMethod`:

```ts
function resolveViaDirectQualifiedMethod(
  lookup: StrategyLookup,
  effectiveReceiver: string,
  call: { name: string },
  relPath: string,
): ReadonlyArray<{ id: number; file: string }> {
  const qualifiedName = `${effectiveReceiver}.${call.name}`;
  return lookup.byName(qualifiedName) ...
}
```

`effectiveReceiver` here is the literal call-site receiver text (`NsAlias`), so it builds the qualified name `NsAlias.doThing` — but the actual definition is stored as `NamespaceObj.doThing` (the object literal's *original* name). The lookup misses because `resolveByReceiver`/`resolveViaDirectQualifiedMethod` has no way to know `NsAlias` is a renamed alias for `NamespaceObj`.

[...]

Issue #1824, opened 2026-07-05
Dynamic import destructuring rename (const { a: b } = await import(...)) records the wrong binding name


carlos-alm:
## Summary

While fixing #1730 (renamed static `import { X as Y }` specifiers not resolving to the correct call edge), I found the same class of bug in the *dynamic* import extraction path: `const { a: b } = await import('./mod')` records the original name (`a`) instead of the local binding actually used at call sites (`b`), so a subsequent call to `b(...)` will not resolve against the exported symbol `a` in `./mod`.

## Root cause

`extractDynamicImportNames` in `src/extractors/javascript.ts` (`pair_pattern` branch, handling `{ a: b } = await import(...)`) currently does:

```js
} else if (child.type === 'pair_pattern') {
  // { a: localName } → use localName (the alias) for the local binding,
  // but use the key (original name) for import resolution
  const key = child.childForFieldName('key');
  if (key) names.push(key.text);
}
```

The comment describes wanting *both* pieces of information, but only `key.text` (`a`, the original exported name) is ever pushed into the flat `names` array — the local alias (`b`) is dropped entirely. Downstream, `buildImportedNamesMap`/`buildImportedNamesForNative` use `imp.names` entries as the *local* binding name to key the `importedNames` map against call-site text (this is exactly the invariant #1730 restored for static imports). Since `names` holds `a` instead of `b` here, a call site `b(...)` never matches this import's map entry.

I have not checked whether the Rust extractor's `extract_dynamic_import_names` (`crates/codegraph-core/src/extractors/javascript.rs`) has the identical issue — likely, given the mirrored-engine convention, but not yet confirmed.

## Repro

```js
async function run() {
  const { realName: alias } = await import('./mod.js');
  alias();
}
```

`alias()` should resolve to `realName` exported from `./mod.js`, but does not.

## Expected

`extractDynamicImportNames` needs the same local/original split #1730 introduced for static `import_specifier` handling (`Import.renamedImports` / `RenamedImport`), so `names` carries the local alias `b` and a rename pair records `{ local: 'b', imported: 'a' }` for the resolver to consult.

## Scope note

Deliberately not fixed as part of #1730 — different extraction function (dynamic `import()` destructuring, not static `import` specifiers), or a distinct enough sub-case to warrant its own fix + tests.

Issue #1823, opened 2026-07-05
Barrel re-export with rename (export { X as Y } from ...) is not tracked, breaking barrel resolution


carlos-alm:
## Summary

While fixing #1730 (renamed *import* specifiers not resolving to the correct call edge), I found a distinct-but-related gap in barrel/re-export handling: a barrel file that re-exports a symbol under a new name (`export { X as Y } from './foo'`) does not track the rename at all, so resolution through that barrel silently breaks for consumers of the renamed name.

## Root cause

`extractImportNames` in `src/extractors/javascript.ts` (and the mirrored `scan_import_names_depth` in `crates/codegraph-core/src/extractors/javascript.rs`) handles `export_specifier` nodes by always reading the tree-sitter `name` field (the *local* declaration being re-exported, i.e. `X`), and never the `alias` field (the external name it's re-exported as, i.e. `Y`). This part of the code is intentionally unchanged by #1730's fix — see the comment left in place in both files.

Given that, `resolveBarrelExport` in `src/domain/graph/builder/stages/resolve-imports.ts` (and `resolve_barrel_export` in `crates/codegraph-core/src/domain/graph/builder/barrel_resolution.rs`) look up a barrel's re-export list by the name a *consumer* of the barrel would import (`Y`), then call `sourceDefinesSymbol(ctx, re.source, symbolName)` to check whether the underlying module actually defines that name. Since `re.names` only ever contains `X` (never `Y`) for a renamed re-export, both the initial `re.names.includes(symbolName)` check and the `sourceDefinesSymbol` check fail whenever a consumer imports the barrel's renamed external name.

## Repro

```js
// underlying.js
export function realName() {}

// barrel.js
export { realName as friendlyName } from './underlying.js';

// consumer.js
import { friendlyName } from './barrel.js';
friendlyName();
```

`friendlyName()` in consumer.js does not resolve to `realName` in underlying.js — the call edge is dropped, and `codegraph exports underlying.js` reports zero consumers for `realName` even though it's genuinely used through the barrel.

## Expected

`resolveBarrelExport`/`resolve_barrel_export` should be able to map the barrel's external name (`Y`) back to the original name (`X`) it re-exports, the same way #1730 taught direct-import resolution to do via `Import.renamedImports`/`RenamedImport`. The export_specifier branch of the extractor needs a parallel rename-pair mechanism (or reuse of the same `renamedImports` field, extended to `export_specifier`) plus barrel-resolution changes to consult it.

## Scope note

[...]

PR #1826, opened 2026-07-05
fix: resolve call edges through renamed import specifiers

### LINKED ISSUE #1730: codegraph exports does not credit consumers that import via a rename (import { X as Y })
## Summary

`codegraph exports <file>` reports zero consumers for an exported function that is genuinely imported and used elsewhere, when the importing file renames it at the import site (`import { X as Y } from '...'`). This is a distinct root cause from the already-filed #1724 (type-only imports not credited) and #1728 (extraction omitting some exports entirely) — here the export IS listed, but its `consumers`/`consumerCount` silently drop any call site reached through a renamed binding.

## Repro

```
codegraph exports src/domain/graph/builder/helpers.ts -T --json
```

Result for `collectFiles` (helpers.ts:230):
```json
{"name": "collectFiles", "kind": "function", "line": 230, "endLine": 258, "role": "utility", "consumers": [], "consumerCount": 0}
```

But `collectFiles` is genuinely imported (and used) via a rename in two files:

```
$ grep -n "collectFiles as" src/domain/graph/builder/stages/*.ts
src/domain/graph/builder/stages/collect-files.ts:17:  collectFiles as collectFilesUtil,
src/domain/graph/builder/stages/native-orchestrator.ts:57:  collectFiles as collectFilesUtil,
```

It's also re-exported through `src/domain/graph/builder.ts` and directly imported/tested in 3 test files.

## Impact

This makes `codegraph exports` / `codegraph roles --role dead` produce false "dead export" positives for any function whose only import sites use a local alias — a common pattern to avoid name collisions (e.g. `collectFiles as collectFilesUtil` to distinguish from a same-named import elsewhere). Combined with #1723/#1724/#1728, this means Rule-4 (dead-code) auditing via `codegraph exports`/`roles --role dead` currently requires manual grep verification of every flagged "dead" export before it can be trusted.

## Expected

Consumer resolution should follow the binding through a renamed import specifier and still credit the call site against the original exported symbol.

## Context

Found while running the Titan GAUNTLET quality-audit pass (`/titan-gauntlet`) over `src/domain/graph/builder/helpers.ts`.

carlos-alm:
## Summary
- `import { X as Y }` extraction in both engines preferred the tree-sitter `name` field over `alias` for `import_specifier` nodes. `name` is always present (the source-declared name); `alias` only exists for renames and holds the local binding actually used at call sites. This meant renamed imports recorded the *original* name instead of the local alias, so calls like `collectFilesUtil(...)` never matched any known imported name and were dropped unresolved.
- Fixing extraction alone wasn't enough — the resolver also needed a local-alias → original-name map to look up the correct symbol in the target file. Added `Import.renamedImports` and threaded it through call resolution, points-to alias resolution, the `Object.defineProperty` post-pass, barrel/type-only edge emission, and cross-file return-type propagation, in both the full-build and incremental paths.

Closes #1730

## Test plan
- `npm test` — full suite green (3398 passed, 0 failed)
- `npm run lint` — clean
- `cargo test --lib` — 427 passed
- Repro confirmed: `codegraph exports src/domain/graph/builder/helpers.ts -T --json` now credits `collectFiles` with `consumerCount: 2`, identical on native and WASM

Filed #1823 (barrel re-export rename `export { X as Y } from ...`), #1824 (dynamic `import()` destructuring rename), #1825 (renamed import used as a call receiver `Y.method()`) — same bug class, different resolution code paths, out of scope for this fix.

Stacked on #1822 (base branch `fix/issue-1729-ast-string-type-annotation`) — only the resolver/extractor/test diff is this issue's change.

### DIFF
--- a/crates/codegraph-core/src/db/repository/graph_read.rs
+++ b/crates/codegraph-core/src/db/repository/graph_read.rs
@@ -28,6 +28,41 @@ fn escape_like(s: &str) -> String {
     out
 }
 
+/// Append an `AND (col LIKE ? [OR col LIKE ? ...])` fragment for a multi-value
+/// file filter and push the corresponding escaped LIKE params. Mirrors
+/// `buildFileConditionSQL()` / `NodeQuery.fileFilter()` in
+/// `src/db/query-builder.ts`, which is why callers must accept `Vec<String>`
+/// (repeatable `-f/--file`) rather than a single `String`. No-op when `files`
+/// is empty. Advances `*idx` past the placeholders it consumes.
+fn push_file_filter(
+    sql: &mut String,
+    params_v: &mut Vec<Box<dyn rusqlite::types::ToSql>>,
+    idx: &mut usize,
+    column: &str,
+    files: &[String],
+) {
+    if files.is_empty() {
+        return;
+    }
+    let start = *idx;
+    if files.len() == 1 {
+        sql.push_str(&format!(" AND {column} LIKE ?{start} ESCAPE '\\'"));
+        params_v.push(Box::new(format!("%{}%", escape_like(&files[0]))));
+        *idx += 1;
+        return;
+    }
+    let clauses: Vec<String> = files
+        .iter()
+        .enumerate()
+        .map(|(i, _)| format!("{column} LIKE ?{} ESCAPE '\\'", start + i))
+        .collect();
+    sql.push_str(&format!(" AND ({})", clauses.join(" OR ")));
+    for f in files {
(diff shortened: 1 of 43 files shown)


carlos-alm:
Addressed the duplication concern from the review summary — extracted `importNamePairs` into a new shared `src/domain/graph/builder/import-utils.ts`, imported by both `stages/build-edges.ts` and `incremental.ts` (the Rust mirror doesn't need this since `import_edges.rs` has no separate incremental copy).

carlos-alm:
@greptileai

PR #1860, opened 2026-07-06
refactor: dedupe busy/locked error detection into isBusyOrLockedError

carlos-alm:
## Summary
- Of the two cleanups in #1749, the `DEFAULTS.db.busyTimeoutMs` registration was already done by an earlier commit in this batch (PR #1788) — confirmed no hardcoded `'busy_timeout = 5000'` literal remains.
- The remaining duplication — `/\b(busy|locked|SQLITE_BUSY|SQLITE_LOCKED)\b/i` inline in both `tryOpenRepoNative` and `openReadonlyWithNative`'s catch blocks — extracted into a shared `isBusyOrLockedError(msg)` helper. Regex behavior unchanged.

Closes #1749

## Test plan
- `npm test` — full suite green (3490 passed, 0 failed)
- `npm run lint` — clean
- `codegraph diff-impact --staged -T` — 3 functions changed, 27 callers affected, consistent with a non-behavioral helper extraction

For #1763's benefit: confirmed `crates/codegraph-core/src/db/connection.rs` has the identical hardcoded `busy_timeout = 5000` literal duplicated at two call sites (`open_read_write`, `open_readonly`). No Rust equivalent of the busy/locked regex exists — rusqlite surfaces `SQLITE_BUSY`/`SQLITE_LOCKED` as typed error variants, not string matching.

Stacked on #1859 (base branch `fix/issue-1748-connection-halstead-effort`) — only the connection.ts diff is this issue's change.

PR #1861, opened 2026-07-06
fix: log statSync failures in findDbPath instead of silently swallowing them

### LINKED ISSUE #1750: db/connection.ts: findDbPath silently swallows all fs.statSync errors (not just ENOENT)
GAUNTLET (Titan phase 15 audit, rule 10) flagged a comment-only catch block in `findDbPath` (lines ~277-279) that silently swallows *all* `fs.statSync` errors, not just the expected "path doesn't exist yet" case:

```ts
try {
  if (fs.statSync(resolved).isDirectory()) { ... }
} catch {
  // Path doesn't exist yet — return as-is (e.g. a future custom DB path).
}
```

Every other catch block in this file (18 of them) calls `debug(...)` with the caught error; this one doesn't, so a genuine unexpected error (e.g. EACCES, a symlink loop) is indistinguishable from the expected "not found yet" case and leaves no diagnostic trail.

Out of scope for phase 15 (narrowly scoped to the openReadonlyWithNative leak-ordering fix + engine-resolution dedup per sync.json's phase-15 label).

Recommendation: add a `debug(...)` call matching the file's existing convention, e.g. `debug(\`findDbPath: statSync failed for ${resolved}, treating as non-existent: ${toErrorMessage(e)}\`)`.

carlos-alm:
## Summary
- `resolveCustomDbPath`'s catch block silently swallowed all `fs.statSync` errors, unlike every other catch block in this file (18 of them), leaving no diagnostic trail for a genuine unexpected error (EACCES, symlink loop) vs. the expected "not found yet" case.
- Added a `debug(...)` call matching the file's existing convention. Behavior unchanged — still falls through to treat the path as not-yet-existing.

Closes #1750

## Test plan
- `npm test` — full suite green (3490 passed, 0 failed)
- `npm run lint` — clean

Stacked on #1860 (base branch `fix/issue-1749-connection-busytimeout-regex-dedupe`) — only the connection.ts diff is this issue's change.

### DIFF
--- a/src/db/connection.ts
+++ b/src/db/connection.ts
@@ -278,8 +278,11 @@ function resolveCustomDbPath(customPath: string): string {
       }
       return path.join(resolved, '.codegraph', 'graph.db');
     }
-  } catch {
+  } catch (e) {
     // Path doesn't exist yet — return as-is (e.g. a future custom DB path).
+    debug(
+      `findDbPath: statSync failed for ${resolved}, treating as non-existent: ${toErrorMessage(e)}`,
+    );
   }
   return resolved;
 }


carlos-alm:
Recreated as #1903 — this PR was inadvertently closed when its stacked base branch was deleted after #1899 merged. Continuing there.

Issue #1865, opened 2026-07-06
reconnectReverseDepEdges still mis-wires same-named siblings when sibling count also changes (incremental rebuild)


carlos-alm:
## Summary

The fix for #1752 (`reconnectReverseDepEdges` in `build-edges.ts` / `reconnect_reverse_dep_edges` in `detect_changes.rs`) correctly resolves the common case where an incremental rebuild shifts a group of same-`(name, kind)` sibling declarations in a file (e.g. several object-literal `close() {}` methods) without changing how many siblings exist, by matching on each target's saved ordinal rank instead of nearest-line proximity.

However, when a same-named/same-kind sibling is **also added or removed** in the same file edit that shifts the group, the sibling-count check in `pickReconnectTarget`/`pick_reconnect_target` intentionally falls back to the pre-#1752 nearest-line heuristic (documented in-code as "a genuinely ambiguous case"). That fallback is exactly the heuristic #1752 proved unreliable once a same-named group shifts far enough — so this specific compound scenario (shift + sibling count change, in the same incremental build) still produces reverse-dep call edges that diverge from a full rebuild of the identical final source.

This is a **pre-existing gap**, not a regression from the #1752 fix — verified by reproducing the identical divergence against the pre-#1752 code on the same scenario (both produce the same wrong result; the #1752 fix does not make this case worse, it just doesn't cover it).

## Reproduction

Using a fixture with 4 functions in one file, each returning an object with its own `close() {}` method (`openA`..`openD`), and 4 reverse-dep caller files each calling `close()` off one of them:

1. Full build.
2. Edit the file: insert an unrelated helper function above all 4 (shifts every `close()` sibling's line down), **and** rename one sibling's method away from `close` (e.g. `openB`'s `close` → `shutdown`), changing the sibling count for `(name="close", kind="method")` in that file from 4 to 3 in the same edit.
3. Incremental rebuild (default).
4. Compare the resulting `close()` call edges (or `fn-impact close --json`) against a full rebuild (`--no-incremental`) of the identical final source — they diverge (some callers end up wired to the wrong remaining sibling, or drop an edge entirely).

The same divergence reproduces with a sibling **added** instead of removed (4 → 5), and reproduces identically on both the WASM/JS engine and the native Rust engine.

## Suggested direction

[...]

PR #1862, opened 2026-07-06
test: add direct unit coverage for closeDbPair/closeDbPairDeferred/closeDbDeferred

carlos-alm:
## Summary
- `closeDbPair`, `closeDbPairDeferred`, and `closeDbDeferred` had zero direct test coverage despite being core resource-lifecycle primitives in a fanIn-55 file — exactly the kind of gap an earlier phase's real resource-leak bug slipped through.
- Added 10 tests to `tests/unit/db.test.ts` (existing home for this file's connection tests): close ordering (native before better-sqlite3, survives native-close failure), deferred-close timing (native immediate, better-sqlite3 deferred to next tick), and advisory-lock release/flush semantics (immediate lock release, `flushDeferredClose()` closes synchronously with no double-close).
- No production code changes — everything needed was already exported.

Closes #1751

## Test plan
- `npm test` — full suite green (3500 passed, 0 failed), all 10 new tests confirmed running and passing (not skipped)
- `npm run lint` / `npm run typecheck` — clean
- Non-vacuousness confirmed via 3 temporary mutation checks (swapped close order, removed double-close guard, moved lock release inside the deferred callback) — each correctly caught by the corresponding test, then reverted

Stacked on #1861 (base branch `fix/issue-1750-finddbpath-swallowed-errors`) — only the test diff is this issue's change.

Issue #1740, opened 2026-07-02
titan-gate: flat blast-radius threshold false-positives on additive-only changes to high-fan-in core functions


carlos-alm:
## Found during Titan FORGE (phase 7, 2026-07-02)

**Severity:** Low-Medium (process friction, not incorrect analysis)

## Problem

\`.claude/skills/titan-gate/SKILL.md\` Step 8 fails any staged change whose transitive blast radius exceeds 30 callers, computed from \`codegraph diff-impact --staged\`. This is a flat, diff-agnostic threshold: it fires identically whether the diff adds new call edges (real risk) or is purely additive/behavior-preserving (e.g., replacing inline literals with named \`DEFAULTS\` constants, no new/changed/removed call edges).

Concretely: \`src/infrastructure/config.ts\`'s \`resolveSecrets\` is reachable only through \`loadConfig\`, which nearly every CLI command calls — so it structurally has ~87 transitive callers regardless of what changes inside it. Any future edit to this function (or others like it: single entry points on the "spine" of the call graph) will always fail Step 8, even risk-free ones.

## Suggestion

Differentiate blast-radius severity by whether the diff changes the call graph shape:
- If the diff adds/removes/rewires call edges → keep current threshold behavior (this is exactly what blast-radius should catch).
- If the diff is edge-neutral (same call sites, same signatures, e.g. literal→constant substitution, internal-only logic changes with no new callees) → the existing 87-caller fan-in is pre-existing risk already accepted by the codebase, not risk introduced by this commit. Consider computing "new blast radius introduced by this diff" (delta vs. the pre-diff graph) rather than raw absolute transitive-caller count.

## Impact

Blocks legitimate low-risk refactors to core/widely-used files (config loading, and likely other central utilities) with a FAIL that requires manual override every time, rather than surfacing only for changes that actually add risk.

PR #1866, opened 2026-07-06
fix: correct blast-radius/fn-impact computation for line-shifted declarations on incremental rebuild

### LINKED ISSUE #1752: Incremental rebuild leaves stale blast-radius/fn-impact numbers for line-shifted declarations (same root cause as #1738, #1743)
While validating phase 15 of the Titan pipeline (a pure ordering/dedup fix in `src/db/connection.ts` that inserts a new ~23-line helper function above several unrelated, unchanged declarations later in the file — shifting their line numbers but not their bodies), `codegraph check --staged --blast-radius 30` reported a FAIL (`maxFound: 46`, violation on `close` at `src/db/connection.ts:368`) immediately after the file was edited (graph updated incrementally, presumably by the `update-graph.sh` post-edit hook).

Rebuilding the same staged state with `codegraph build --no-incremental` and re-running the identical `codegraph check --staged --blast-radius 30` command produced a **different, passing result**: `maxFound: 28`. This was stable across repeated re-runs on the unchanged, cleanly-rebuilt state.

Root-cause hypothesis: this looks like the same incremental-staleness bug family already tracked as #1738 (stale `codegraph structure` directory metrics) and #1743 (stale cyclomatic-complexity values), but manifesting in a third subsystem: blast-radius / fn-impact / diff-impact computation. When a file changes such that unrelated, byte-identical declarations shift line numbers (e.g. inserting a new function above them), the incremental rebuild path appears to leave stale line-keyed edge/impact data for those declarations, producing an inflated and non-reproducible blast-radius count until a full (`--no-incremental`) rebuild is performed.

Reproduction:
1. In a repo with an existing graph.db, edit a file to insert N new lines of a self-contained function above several existing, otherwise-untouched functions (shifting their line numbers only).
2. Let the incremental rebuild run (e.g. via the post-edit hook, or `codegraph build` without `--no-incremental`).
3. Run `codegraph check --staged --blast-radius <threshold>` (or `codegraph fn-impact <name>`) against one of the shifted declarations — observe an inflated/incorrect blast-radius number.
4. Run `codegraph build --no-incremental`, then repeat step 3 — observe a different (correct, lower) number that is stable across repeated runs.

This is a real correctness bug in the incremental build path, not a documentation-worthy "acceptable gap" — filing per project policy that codegraph must produce correct results when analyzing itself.

carlos-alm:
## Summary
- `reconnectReverseDepEdges`/`reconnect_reverse_dep_edges` re-attach a reverse-dep caller's edge to its purged-and-reinserted target using `(name, kind, file)` plus nearest-old-line as a tiebreak. When unrelated code shifts a whole group of same-named/same-kind siblings (e.g. `connection.ts` has four object-literal `close() {}` methods), an old reference line can end up numerically closer to the *wrong* sibling's new line, silently reattaching the edge to it — losing edges on one sibling while duplicating them on another.
- `checkMaxBlastRadius`/`fnImpactData` are pure node-ID graph traversal with no line-keyed logic — confirmed the defect is entirely upstream in edge reconnection, no interaction with #1740's diff-handling changes in the same file.
- Fixed by recording each target's 1-based ordinal rank among same-(name,kind) siblings before purge (`computeNodeOrdinals`/`compute_ordinals`), and reconnecting using that saved ordinal instead of line proximity (`pickReconnectTarget`/`pick_reconnect_target`), falling back to nearest-line only if the sibling count itself changed.

Closes #1752

## Test plan
- `npm test` — full suite green (3502 passed, 0 failed)
- `npm run lint` — clean
- `cargo test --release` — 443 passed
- A/B proof on both engines: reverted to pre-fix source, rebuilt, reproduced the exact bug signature; restored fix, reconfirmed pass
- Scenario matrix: 3 shift positions × 4 touched-sibling variants × 2 engines = 24 combinations, all pass against full-rebuild ground truth
- 5 repeated cycles + a 3-step chained incremental sequence (no full rebuild in between) — stable

Filed #1865 for a genuinely separate, pre-existing (not a regression) gap found during stress-testing: when a same-named/same-kind sibling is also added or removed in the same edit that shifts the group, the ordinal-based fix's own line-proximity fallback still diverges from a full rebuild — suggests per-declaration content hashing as the fix.

Stacked on #1862 (base branch `fix/issue-1751-closedbpair-test-coverage`) — only the builder/detect-changes/test diff is this issue's change.

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs
@@ -369,6 +369,54 @@ pub struct SavedReverseDepEdge {
     pub edge_kind: String,
     pub confidence: f64,
     pub dynamic: i64,
+    /// 1-based rank of the target (by ascending line) among nodes sharing its
+    /// (name, kind) within `tgt_file`, computed at save time — see #1752.
+    pub tgt_ordinal: i64,
+    /// Size of that (name, kind) sibling group at save time.
+    pub tgt_sibling_count: i64,
+}
+
+/// Computes each node's 1-based ordinal rank (by ascending line) among nodes
+/// sharing its (name, kind) within `file`, plus the sibling-group size,
+/// keyed by `(name, kind, line)`.
+///
+/// A file can contain multiple distinct symbols with the identical name and
+/// kind — e.g. several object-literal `close() {}` methods returned from
+/// different functions in the same file. `(name, kind, file)` alone is not a
+/// unique identity for such symbols, so `reconnect_reverse_dep_edges` cannot
+/// safely tell them apart by nearest-line matching once unrelated code
+/// shifts the candidates unevenly (#1752). The ordinal recorded here — the
+/// target's rank among same-named siblings at save time — lets reconnection
+/// map an old target to its new node correctly as long as the sibling count
+/// is unchanged, regardless of how far the whole group has shifted.
+fn compute_ordinals(conn: &Connection, file: &str) -> HashMap<(String, String, i64), (i64, i64)> {
+    let mut by_group: HashMap<(String, String), Vec<i64>> = HashMap::new();
+    let mut result = HashMap::new();
+    let mut stmt = match conn.prepare("SELECT name, kind, line FROM nodes WHERE file = ?1") {
+        Ok(s) => s,
+        Err(_) => return result,
+    };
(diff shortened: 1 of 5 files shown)


carlos-alm:
Recreated as #1916 — this PR was inadvertently closed when its stacked base branch was deleted after #1910 merged. Continuing there.

Issue #1871, opened 2026-07-06
leiden/partition.ts: makePartition exceeds cyclomatic threshold (cyc=10)


carlos-alm:
## What

\`codegraph complexity --file src/graph/algorithms/leiden/partition.ts --health -T --json\` flags \`makePartition\` (the factory function that wires \`PartitionState\` to the public \`Partition\` interface) as exceeding the cyclomatic complexity threshold:

\`\`\`
cyclomatic: 10   (threshold: warn >= 10, see DEFAULTS.complexity in src/infrastructure/config.ts)
cognitive: 13
maintainabilityIndex: 32.5
exceeds: ["cyclomatic"]
\`\`\`

Found incidentally while working on #1755 (a sibling complexity fix for \`computeDeltaModularityDirected\` in the same file) — out of scope for that issue, filing separately per repo scope-discipline convention.

## Likely source of complexity

\`makePartition\` builds an object literal with many getter methods, several of which have their own bounds-check ternaries inline (e.g. \`getCommunityTotalSize\`, \`getCommunityNodeCount\`, \`getNeighborEdgeWeightToCommunity\` and friends with \`|| 0\` fallbacks). Each inline ternary/\`||\` in the returned object literal adds to the enclosing function's cyclomatic count even though they're trivial one-liners.

## Suggested approach

Decompose by extracting the getter closures (or groups of them) into small named functions that close over \`s: PartitionState\`, similar to how \`initAggregates\`, \`accumulateNeighborWeights\`, \`moveNode\`, and \`compactIds\` are already extracted and merely referenced in the returned object literal. This should pull the branchy one-liners out of \`makePartition\` itself.

Note: while decomposing, keep in mind #1755's finding that the \`|| 0\` fallback after \`fget\`/bounds-check reads in this file is currently a no-op given Float64Array invariants (dense, zero-initialized, NaN-scrubbed at adapter.ts ingestion) — a shared \`fgetOrZero\` helper was added to \`typed-array-helpers.ts\` in that fix and could be reused here instead of hand-rolling the same ternary+\`||\` pattern again.

PR #1872, opened 2026-07-06
refactor: reduce cyclomatic complexity of computeDeltaModularityDirected

### LINKED ISSUE #1755: leiden/partition.ts: computeDeltaModularityDirected exceeds cyclomatic threshold (cyc=11)
## Context

Found while doing phase-22 complexity remediation (titan-forge) on `src/graph/algorithms/leiden/partition.ts` and `adapter.ts`. Fixed the functions explicitly named in the gauntlet recommendation (`moveNode`, `makeGraphAdapter`/`populateUndirectedEdges`) plus the two other cognitive/cyclomatic-exceeding functions caught by the same violation (`buildSortedCommunityIds`, `computeDeltaCPM`).

`computeDeltaModularityDirected` (partition.ts) was **not** part of that recommendation and predates this phase (present before phase 3's shared-helper extraction and before phase 22). Current metrics:

```
computeDeltaModularityDirected: cognitive=10, cyclomatic=11 (exceeds cyclomatic, threshold is 10)
```

It has no directed/undirected branch to split (it's already the directed-only variant). Its complexity comes from four repeated `newC < arr.length ? fget(arr, newC) [|| 0] : 0` bounds-checked-getter expressions. A generic `fgetOrZero(arr, c)` helper could reduce this, but two of the four call sites omit the `|| 0` fallback (`totalInStrengthNew`/`totalOutStrengthNew`) while the other two include it (`inFromNew`/`outToNew`) — unifying them risks a subtle behavior change (masking a NaN case the current code doesn't mask) that needs its own careful review, not a same-phase drive-by fix.

## Ask

Decompose `computeDeltaModularityDirected` (or otherwise reduce its cyclomatic complexity below the threshold) in a dedicated pass, verifying the `||0` fallback inconsistency intentionally or fixing it with an explicit behavior note.

## Scope note

Filed per project scope-discipline rule — out of scope for phase 22, not silently left undocumented.

carlos-alm:
## Summary
- `computeDeltaModularityDirected` (cyc=11, threshold 10) had a documented `||0` asymmetry across its 4 bounds-checked getter call sites, flagged as needing investigation before any unification (risk of masking a real NaN case).
- Confirmed the asymmetry is behaviorally inert: all arrays involved are dense, zero-initialized `Float64Array`s, and every value entering them is pre-scrubbed through `+x || 0` at ingestion (`adapter.ts`) — NaN/undefined can never reach these arrays. Cross-checked 3 ways: the same inconsistency exists within `computeDeltaCPM` on the same array, `modularity.ts`'s independently-written `diffModularityDirected` reproduces the identical split, and the asymmetry already existed unchanged in the original vendored `ngraph.leiden` source at the initial vendoring commit.
- Added a shared `fgetOrZero` helper (documented with this finding), applied uniformly to all 8 reads, and extracted `computeDirectedEdgeWeightTerms`/`computeDirectedStrengthTerms`. Cyclomatic 11→3, cognitive 10→2.

Closes #1755

## Test plan
- `npm test` — full suite green (3519 passed, 0 failed)
- `npm run lint` — clean
- Exact (not approximate) behavior-preservation check: 12 combinations (file/function-level × 3 seeds × 2 resolutions) on this repo's own 701-file/8833-function dependency graph, pre-fix vs. post-fix — byte-for-byte identical quality scores and community assignments, after controlling for an unrelated native-vs-JS-engine resolution difference in the test environment (not a regression)
- New tests pin `fgetOrZero`'s bounds/zero-fill contract and exact-value regression tests for `computeDeltaModularityDirected`'s output

Filed #1871 for an out-of-scope finding: `makePartition` (cyc=10, pre-existing, unrelated) also exceeds the threshold.

Stacked on #1870 (base branch `fix/issue-1754-domain-search-console-log`) — only the leiden/test diff is this issue's change.

### DIFF
--- a/src/graph/algorithms/leiden/partition.ts
+++ b/src/graph/algorithms/leiden/partition.ts
@@ -9,7 +9,7 @@
 
 import type { GraphAdapter } from './adapter.js';
 import { accumulateInternalEdgeWeights, accumulateNodeAggregates } from './aggregate-helpers.js';
-import { fget, iget, u8get } from './typed-array-helpers.js';
+import { fget, fgetOrZero, iget, u8get } from './typed-array-helpers.js';
 
 export interface CompactOptions {
   keepOldOrder?: boolean;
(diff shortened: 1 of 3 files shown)


carlos-alm:
Recreated as #1925 — this PR was inadvertently closed when its stacked base branch was deleted after #1870 merged. Continuing there.

Issue #1876, opened 2026-07-06
Rust resolver misses receiver-typed calls on locally-typed variables and trait-dispatch through local vars


carlos-alm:
## Summary

While adding `tests/benchmarks/resolution/fixtures/rust/main.rs` (#1759), I confirmed that both the WASM and native Rust engines agree on the same set of unresolved call edges — this is **not** an engine-parity divergence (native and WASM produce identical results), it's a genuine resolution gap for two related patterns:

1. **Receiver-typed calls on a local variable typed by another function's return value**, e.g.:
   ```rust
   let service = build_service();   // build_service() -> UserService
   service.add_user(1, "Alice", "alice@example.com");  // not resolved
   service.get_user(1);                                 // not resolved
   service.remove_user(1);                              // not resolved
   ```
   Also applies to `self.field.method()` inside a struct's own methods (e.g. `UserService::get_user`'s `self.repo.find_by_id(id)`, where `repo: UserRepository` is a struct field) — these were already-documented false negatives in the fixture before this change.

2. **Trait-dispatch through a local variable**, e.g.:
   ```rust
   let email_v = EmailValidator::new();
   email_v.validate(user)?;   // Validator::validate — not resolved
   ```

## Evidence

`tests/benchmarks/resolution/fixtures/rust/expected-edges.json` documents these as real, hand-verified call edges (`mode: "receiver-typed"` / `mode: "trait-dispatch"`) that neither engine currently resolves — they show up as false negatives in `resolution-benchmark.test.ts`'s rust suite:

```
validate_all -> EmailValidator.validate   (trait-dispatch)
validate_all -> NameValidator.validate    (trait-dispatch)
UserService.get_user -> UserRepository.find_by_id   (receiver-typed, self.field)
UserService.add_user -> UserRepository.save         (receiver-typed, self.field)
UserService.remove_user -> UserRepository.delete    (receiver-typed, self.field)
main -> UserService.add_user       (receiver-typed, local var typed by return value)
main -> UserService.get_user       (receiver-typed, local var typed by return value)
main -> UserService.remove_user    (receiver-typed, local var typed by return value)
main -> User.display_name          (receiver-typed, local var typed by return value)
direct_repo_access -> UserRepository.save  (receiver-typed, local var typed by constructor)
```

Verified via `buildGraph(fixtureDir, { engine: 'wasm' | 'native', ... })` on the fixture — both engines produce the identical 14 resolved edges, missing all 10 edges above.

[...]

PR #1877, opened 2026-07-06
fix: add main.rs driver to rust dynamic tracer fixture

### LINKED ISSUE #1759: Rust dynamic tracer fixture has no main.rs — trace_rust always fails before producing edges
## Bug

`tests/benchmarks/resolution/fixtures/rust/` contains `models.rs`, `repository.rs`, `service.rs`, `validator.rs` but no `main.rs` (and no `fn main` anywhere in the fixture). `native-tracer.sh`'s `trace_rust()` unconditionally does:

```bash
sedi '1s/^/mod trace_support;\n/' "$TMP_DIR/src/main.rs"
```

which fails immediately (`set -euo pipefail`) because `main.rs` doesn't exist among the copied fixture files, so the tracer never reaches the injection/compile/run steps and produces no output at all.

## Impact

`run-tracer.mjs`/`tracer-validation.test.ts` catches the resulting non-zero exit and treats it as "toolchain not available", silently skipping the rust same-file recall assertion (threshold 0.5) rather than actually validating it — even on machines with `cargo` installed. This masks the fact that the rust dynamic tracer path has never actually been exercised in CI or locally.

## Repro

```bash
bash tests/benchmarks/resolution/tracer/native-tracer.sh tests/benchmarks/resolution/fixtures/rust rust
```

Fails with `sed: .../src/main.rs: No such file or directory` (both on `origin/main` and unrelated to any in-flight branch work — verified against a clean checkout of HEAD).

## Suggested fix

Add a `main.rs` to `tests/benchmarks/resolution/fixtures/rust/` (mirroring the `main.swift`/`main.dart`/`main.zig` driver files that already exist in the sibling swift/dart/zig fixtures) that exercises the models/repository/service/validator call chain, so `trace_rust` actually compiles and runs.

carlos-alm:
## Summary
- `tests/benchmarks/resolution/fixtures/rust/` had no `main.rs`, so `native-tracer.sh`'s `trace_rust()` failed immediately on the injection step, before ever reaching compile/run — silently skipped as "toolchain not available" by the consuming test, masking the fact this tracer path had never actually been exercised.
- Added `main.rs` exercising the full models/repository/service/validator chain. Getting it to actually run end-to-end (the issue's real impact concern) surfaced 3 more latent bugs in the same never-exercised path, all fixed as necessary prerequisites: a GNU-only sed form in `native-tracer.sh` that BSD sed (macOS) rejects, an impl-block context regex that captured the trait name instead of the concrete type for `impl Trait for Type`, and the fixture itself never compiling as a real crate (missing `Clone` derive, missing trait import).

Closes #1759

## Test plan
- `npm test` — full suite green (3519 passed, 0 failed)
- `npm run lint` — clean (`.rs`/`.sh` outside Biome's scope, no other format gate applies)
- `native-tracer.sh` now compiles and runs end-to-end, producing 23/24 expected edges (the one miss is legitimately unreachable at runtime — a repo stub's `save()` never persists into its `HashMap`, pre-existing fixture behavior)
- `tracer-validation.test.ts`'s rust case now actually runs (not skipped) at 100% same-file recall against the 0.5 threshold
- `resolution-benchmark.test.ts`'s rust suite: 100% precision, 58.3% recall, both above thresholds

Filed #1876 for a genuinely separate finding surfaced by this work: both engines miss receiver-typed calls on locally-typed variables and trait-dispatch through local variables — confirmed as a real resolution gap, not an engine-parity divergence.

Stacked on #1875 (base branch `fix/issue-1757-token-benchmark-dedupe-persist`) — only the rust fixture/tracer diff is this issue's change.

### DIFF
--- a/tests/benchmarks/resolution/fixtures/rust/expected-edges.json
+++ b/tests/benchmarks/resolution/fixtures/rust/expected-edges.json
@@ -100,6 +100,76 @@
       "kind": "calls",
       "mode": "same-file",
       "notes": "Same-file associated function call (UserService::new)"
+    },
+    {
+      "source": { "name": "main", "file": "main.rs" },
+      "target": { "name": "build_service", "file": "service.rs" },
+      "kind": "calls",
+      "mode": "module-function",
+      "notes": "Cross-module function call via use crate::service::build_service"
+    },
+    {
+      "source": { "name": "main", "file": "main.rs" },
+      "target": { "name": "UserService.add_user", "file": "service.rs" },
+      "kind": "calls",
+      "mode": "receiver-typed",
+      "notes": "service.add_user() — service is typed as UserService (returned by build_service)"
+    },
+    {
+      "source": { "name": "main", "file": "main.rs" },
+      "target": { "name": "UserService.get_user", "file": "service.rs" },
+      "kind": "calls",
+      "mode": "receiver-typed",
+      "notes": "service.get_user() — service is typed as UserService (returned by build_service)"
+    },
+    {
+      "source": { "name": "main", "file": "main.rs" },
+      "target": { "name": "User.display_name", "file": "models.rs" },
+      "kind": "calls",
+      "mode": "receiver-typed",
+      "notes": "user.display_name() — user is typed as User (returned by UserService::get_user)"
+    },
+    {
+      "source": { "name": "main", "file": "main.rs" },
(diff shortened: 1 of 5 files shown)


carlos-alm:
Recreated as #1935 — this PR was inadvertently closed when its stacked base branch was deleted after #1934 merged. Continuing there.

PR #1878, opened 2026-07-06
fix: scope diff file-header detection to between-hunk positions

### LINKED ISSUE #1761: check.ts parseDiffOutput: header detection isn't hunk-scoped, can misclassify body lines
## Bug (minor, latent)

Noticed while fixing #1760 (\`checkNoSignatureChanges\` false positives from diff context lines).

\`parseDiffOutput\` in \`src/features/check.ts\` runs \`isSourceFileHeaderLine\` (checks \`line.startsWith('--- ')\`) and \`extractNewFileName\` (checks \`^\+\+\+ b\/\`) against **every** line in the diff output unconditionally — not just lines expected to be file headers between hunks. There's no parser state distinguishing "between files, expecting a header" from "inside a hunk body."

If a hunk body contains a **removed** line whose original content starts with \`-- \` (so the diff-prefixed line becomes \`--- ...\`), it gets misclassified as a file-header line and silently dropped — never reaching the old-line-cursor tracking added in #1760's fix. This desyncs the cursor for the rest of that hunk, producing incorrect (too-low) old-side line numbers for every subsequent line, which could cause \`checkNoSignatureChanges\` to miss a real signature change or misattribute one to the wrong declaration.

Same latent issue applies symmetrically to \`+++ b/...\`-prefixed added lines via \`extractNewFileName\`.

## Likelihood

Very low for typical source files (TS/Rust/etc. rarely have lines starting with \`-- \` or \`++ \`), more plausible in diffs touching Markdown/YAML/reST files with \`---\` horizontal rules or frontmatter delimiters.

## Suggested fix

Make the parser hunk-scoped: only attempt file-header matching when not currently inside a hunk (i.e., after a hunk's line budget — tracked via old/new counts — is exhausted, or immediately after a \`diff --git\` line), rather than testing every line against the header regexes regardless of position.

carlos-alm:
## Summary
- `parseDiffOutput` tested every diff line against file-header regexes (`--- `, `+++ b/`) unconditionally. A removed line whose original content starts with `-- ` becomes `--- ...` once diff-prefixed — indistinguishable from a real file header by content alone, causing it to be silently dropped and desyncing the old-line cursor for the rest of the hunk. Same latent issue for added lines starting with `++ b/`.
- Added `oldHunkEnd`/`newHunkEnd` tracking (computed from the most recent `@@ -a,b +c,d @@` header) and an `insideHunk()` check; file-header detection now only runs when not currently inside a hunk body. Hunk-header detection itself is untouched (unambiguous by construction).

Closes #1761

## Test plan
- `npm test` — full suite green (3521 passed, 0 failed)
- `npm run lint` — clean
- Two new regression tests (removed-line and added-line cases), diff format verified against real `git diff --unified=0` output on temp fixtures — confirmed both fail with the exact predicted corruption against pre-fix code, pass post-fix
- Along the way, fixed an existing "handles multiple files" test fixture that had hunk headers declaring zero actual body lines (unrealistic, happened to work under the old unconditional matching but would break under correct position-aware parsing) — added matching body content, preserved original headers/assertions

Stacked on #1877 (base branch `fix/issue-1759-rust-tracer-fixture-mainrs`) — only the check.ts/test diff is this issue's change.

### DIFF
--- a/src/features/check.ts
+++ b/src/features/check.ts
@@ -74,6 +74,9 @@ function isDevNullTargetLine(line: string): boolean {
 class DiffLineTracker {
   private oldLineCursor = 0;
   private newLineCursor = 0;
+  /** End-exclusive old/new bounds declared by the current hunk's `@@ -a,b +c,d @@` header — see `insideHunk`. */
+  private oldHunkEnd = 0;
+  private newHunkEnd = 0;
   private removedRunStart: number | null = null;
   private removedRunEnd: number | null = null;
   private addedRunStart: number | null = null;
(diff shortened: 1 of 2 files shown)


carlos-alm:
Recreated as #1940 — this PR was inadvertently closed when its stacked base branch was deleted after #1935 merged. Continuing there.

Issue #1884, opened 2026-07-06
Computed string-literal keys not unwrapped in several typeMap/prototype extraction sites (both engines)


carlos-alm:
## Bug

While fixing #1764 (`extractObjectLiteralFunctions`'s `pair` branch emitting `obj.['foo']` instead of `obj.foo` for computed string-literal keys), the same "only unwrap plain `string`-typed keys, fall back to raw bracket/quote text for `computed_property_name`" pattern was found in several other object-key extraction sites across both engines. These are separate from the `#1764` fix (which only covers `extractObjectLiteralFunctions` / `extract_object_literal_functions` + `match_js_objlit_qualified_method_defs`, the functions that determine `Definition.name`) — the sites below all feed `typeMap` seeding or other adjacent extraction paths, not `Definition.name` directly, so they don't affect `codegraph where`'s reported name, but they do affect resolution correctness (dangling/garbled typeMap keys) for the affected call patterns.

None of these unwrap `computed_property_name` → inner `string`/`string_fragment` the way `resolveMethodDefinitionName`/`resolve_method_def_name` (and now the fixed `pair` branches) do.

### TypeScript (`src/extractors/javascript.ts`)

- `handleObjectLiteralTypeMap`'s `pair` branch — same string-only gap for typeMap seeding (used for `this.x()` resolution via object-literal aliasing).
- `handleObjectLiteralTypeMap`'s `method_definition` branch — worse: doesn't call `resolveMethodDefinitionName` at all, uses raw `nameNode.text` unconditionally (would even mishandle a `[Symbol.iterator]`-style key, not just string literals).
- `handleDefinePropertyTypeMap`'s `defineProperties` pair-iteration branch — same string-only gap for `Object.defineProperties(obj, { ['foo']: descriptor })`.
- `seedProtoProperties`'s `pair` branch — same string-only gap (seeds pts keys from `Object.create({ ['foo']: identifier })`-style prototype object literals).
- `collectObjectRestParams`'s `pair` branch — different shape: explicitly excludes *all* computed keys via `keyN.type !== 'computed_property_name'`, including resolvable string literals (under-extraction rather than a garbage name, but same root cause — worth unwrapping resolvable ones instead of blanket-skipping).
[...]

PR #1885, opened 2026-07-06
fix: resolve computed string-literal keys in object-literal extraction

### LINKED ISSUE #1764: Object literal computed method keys extract as `obj.['foo']` instead of `obj.foo`
## Bug

In \`src/extractors/javascript.ts\`, \`extractObjectLiteralFunctions\`'s \`pair\` branch (around line 1272-1273) only strips quotes for plain \`string\`-typed keys:

\`\`\`ts
const keyName =
  keyNode.type === 'string' ? keyNode.text.replace(/^['"]|['"]$/g, '') : keyNode.text;
\`\`\`

For a computed string-literal key like \`{ ['foo']: () => {} }\`, \`keyNode.type\` is \`computed_property_name\`, not \`string\`, so this falls through to \`keyNode.text\`, producing the literal bracket/quote text. The emitted definition name becomes \`obj.['foo']\` instead of \`obj.foo\`, so \`obj.foo()\` call sites can never resolve to it.

Reproduced identically on both engines (not a parity gap — both are wrong the same way):

\`\`\`js
const obj = {
  ['foo']: () => { return 1; },
  bar: () => { return 2; },
};
obj.foo();
obj.bar();
\`\`\`

\`codegraph where --file sample.js -T --json\` (both \`--engine native\` and \`--engine wasm\`) reports:
\`\`\`json
{ "name": "obj.['foo']", "kind": "function", "line": 2 }
\`\`\`
instead of \`obj.foo\`.

## Fix direction

The adjacent \`method_definition\` branch in the same function already has a helper, \`resolveMethodDefinitionName\` (src/extractors/javascript.ts, extracted in the Titan phase 10 decomposition), that correctly unwraps \`computed_property_name\` -> inner \`string\`/\`string_fragment\` -> stripped text, returning \`''\` for unresolvable computed keys (e.g. \`[Symbol.iterator]\`). The \`pair\` branch's \`keyName\` computation should use the same unwrapping logic (skip the pair if the computed key isn't a resolvable string, mirroring the existing \`method_definition\` branch's \`continue\` behavior) instead of falling back to raw text. The mirrored Rust extractor (\`crates/codegraph-core/src/extractors/\`) should be checked for the same gap.

Found while auditing forge phase 10 (javascript.ts decomposition) during a Titan grind pass — out of scope for that phase's helper-adoption work, filed separately per project convention.

carlos-alm:
## Summary
- `extractObjectLiteralFunctions`'s `pair` branch only stripped quotes for plain `string`-typed keys; a computed string-literal key like `{ ['foo']: () => {} }` fell through to raw `keyNode.text`, emitting `obj.['foo']` instead of `obj.foo`, so `obj.foo()` call sites could never resolve to it. Reproduced identically on both engines.
- Extracted the existing `computed_property_name`-unwrapping logic (already correct in the adjacent `method_definition` branch's `resolveMethodDefinitionName`) into a shared `resolveComputedKeyName` helper, reused by the `pair` branch. Rust mirror needed two fix sites (const path and let/var path, which Rust splits into separate functions where TS uses one).

Closes #1764

## Test plan
- `npm test` — full suite green (3533 passed, 0 failed)
- `npm run lint` / `npm run typecheck` — clean
- `cargo test --release` — 448 passed
- Exact repro confirmed fixed on both engines: `obj.foo`/`obj.bar` now resolve correctly (verified via `where` and `query` showing identical caller edges on both engines)
- `[Symbol.iterator]` computed key confirmed skipped (no garbage definition) on both engines, matching `method_definition`'s existing precedent; let/var parity confirmed separately

Filed #1884 for an out-of-scope finding: the same unwrapping gap exists in several adjacent typeMap-seeding functions (affecting `typeMap` seeding, not `Definition.name`, so they don't reproduce this issue's symptom).

Stacked on #1883 (base branch `fix/issue-1763-busytimeout-readonly-wiring`) — only the extractor/test diff is this issue's change.


carlos-alm:
Auto-closed when its stacked base PR #1943 was merged with --delete-branch (GitHub closes rather than retargets dependents in this case). No commits lost — recreated against `main` directly as #1954.

Issue #1944, opened 2026-07-07
Quoted (non-computed) method/property keys extract with literal quotes, e.g. `A.'foo'` instead of `A.foo`


carlos-alm:
## Bug

While fixing #1764 (computed property key unwrapping), I found a related but distinct bug: plain **quoted** (non-computed) string keys used as method/property names are not stripped of their quote characters, producing broken definition names.

\`resolveMethodDefinitionName\` (\`src/extractors/javascript.ts\`) only strips quotes for the *computed* branch:

\`\`\`ts
function resolveMethodDefinitionName(nameNode: TreeSitterNode): string {
  if (nameNode.type !== 'computed_property_name') return nameNode.text;
  const inner = nameNode.child(1);
  if (!inner || (inner.type !== 'string' && inner.type !== 'string_fragment')) {
    return '';
  }
  return inner.text.replace(/^['"]|['"]$/g, '');
}
\`\`\`

For a plain quoted method key (not inside \`[...]\`), \`nameNode.type\` is \`'string'\` (not \`'computed_property_name'\`), so it falls through to \`return nameNode.text\`, which includes the surrounding quote characters.

Repro:

\`\`\`js
class A {
  'foo'() { return 1; }
}
\`\`\`

\`codegraph where --file sample.js -T --json\` reports:
\`\`\`json
{ "name": "A.'foo'", "kind": "method", "line": 2 }
\`\`\`
instead of \`A.foo\`.

Same issue for object-literal quoted method shorthand, e.g. \`{ 'quoted'() { return 1; } }\` → \`obj.'quoted'\`.

## Fix direction

\`resolveMethodDefinitionName\`'s non-computed branch should strip quotes when \`nameNode.type === 'string'\` (matching the quote-stripping already applied to computed string keys, and to the sibling \`resolvePropertyKeyName\` helper added in #1764's fix). The mirrored Rust extractor (\`crates/codegraph-core/src/extractors/javascript.rs\`, \`resolve_method_def_name\`) has the identical gap and should be checked too.

Found while auditing the \`extractObjectLiteralFunctions\`/\`resolveMethodDefinitionName\` family of key-resolution helpers during the #1764 fix — out of scope for that PR, filed separately per project convention.

PR #1945, opened 2026-07-07
fix(extractors): unwrap computed property keys in object-literal key resolution

### LINKED ISSUE #1764: Object literal computed method keys extract as `obj.['foo']` instead of `obj.foo`

carlos-alm:
Fixes #1764

## Bug

`extractObjectLiteralFunctions`'s `pair` branch only stripped quotes for plain `string`-typed keys. For a computed string-literal key like `{ ['foo']: () => {} }`, the key node type is `computed_property_name`, so it fell through to the raw bracket/quote text, producing `obj.['foo']` instead of `obj.foo`. Call sites like `obj.foo()` could never resolve to it. Reproduced identically on both engines.

## Fix

Auditing the codebase turned up the identical bug pattern duplicated across five other key-resolution call sites in the JS/TS extractor (typeMap seeding for object literals, `Object.defineProperties`, `Object.create`/prototype-literal seeding, prototype assignment), and mirrored in the native Rust extractor. Per the project's "best architecture over smallest diff" guidance, I added one shared helper per engine — `resolvePropertyKeyName` (WASM/TS) and `resolve_property_key_name` (native/Rust) — that unwraps `computed_property_name` string literals the same way the existing `resolveMethodDefinitionName` / `resolve_method_def_name` already did for `method_definition` name nodes, and switched all six affected call sites in each engine to use it. This fixes the whole bug class identically in both engines instead of leaving four latent copies of the same defect behind.

Non-string computed keys (e.g. `[Symbol.iterator]`) are skipped rather than producing a garbage name, matching the existing `method_definition` behavior.

## Out of scope

Found a separate, distinct bug while auditing this code: `resolveMethodDefinitionName`/`resolve_method_def_name` don't strip quotes for **non-computed** quoted string keys (`{ 'foo'() {} }` → `A.'foo'` instead of `A.foo`). Filed as #1944, not fixed here.

## Validation

- New unit tests in `tests/parsers/javascript.test.ts` covering the `pair` branch (arrow/function values, non-string computed keys).
- New integration test `tests/integration/issue-1764-computed-objlit-key.test.ts` exercising both WASM and native engines end-to-end (build graph, verify plain node name, verify call edge resolves, verify no bracketed names leak).
- New Rust unit tests in `crates/codegraph-core/src/extractors/javascript.rs` covering `extract_object_literal_functions`, `extract_js_prototype_object_literal`, and `seed_descriptor_object` with computed keys.
- Manually reproduced the issue's exact repro (`codegraph where --file sample.js -T --json`) with both `--engine native` and `--engine wasm` before and after the fix — both now report `obj.foo` identically.
- Full suite: `npm test` (214 files, 3550 passed), `cargo test --release` (451 passed), `npm run lint` clean, `npm run build` clean.
- `codegraph diff-impact --staged -T` confirmed the blast radius matched intent: the 5 fixed call sites plus the new shared helper, 21 transitive callers, no unexpected files touched.

### DIFF
--- a/crates/codegraph-core/src/extractors/javascript.rs
+++ b/crates/codegraph-core/src/extractors/javascript.rs
@@ -399,12 +399,7 @@ fn seed_object_create_entries(var_name: &str, call_node: &Node, source: &[u8], s
                 let Some(key_n) = child.child_by_field_name("key") else { continue };
                 let Some(val_n) = child.child_by_field_name("value") else { continue };
                 if val_n.kind() != "identifier" { continue; }
-                let key = if key_n.kind() == "string" {
-                    extract_string_fragment(&key_n, source).map(|s| s.to_string())
-                } else {
-                    Some(node_text(&key_n, source).to_string())
-                };
-                let Some(key) = key else { continue };
+                let Some(key) = resolve_property_key_name(&key_n, source) else { continue };
                 symbols.type_map.push(TypeMapEntry {
                     name: format!("{}.{}", var_name, key),
                     type_name: node_text(&val_n, source).to_string(),
@@ -423,12 +418,7 @@ fn seed_descriptor_object(obj_name: &str, obj_node: &Node, source: &[u8], symbol
         if child.kind() != "pair" { continue; }
         let Some(key_n) = child.child_by_field_name("key") else { continue };
         let Some(val_n) = child.child_by_field_name("value") else { continue };
-        let key = if key_n.kind() == "string" {
-            extract_string_fragment(&key_n, source).map(|s| s.to_string())
-        } else {
-            Some(node_text(&key_n, source).to_string())
-        };
-        let Some(key) = key else { continue };
+        let Some(key) = resolve_property_key_name(&key_n, source) else { continue };
         let Some(target) = find_descriptor_value(&val_n, source) else { continue };
         symbols.type_map.push(TypeMapEntry {
             name: format!("{}.{}", obj_name, key),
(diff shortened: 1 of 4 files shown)


carlos-alm:
Closing — this duplicates already in-progress work on #1885 for the same issue (#1764). Continuing the existing stacked chain (#1885 → #1939) instead of a competing branch.

Issue #1946, opened 2026-07-07
Incremental rebuild missing Kotlin-reflection and reflection-keyExpr call fallbacks (full-build/incremental parity gap)


carlos-alm:
## What

While fixing #1765 (missing same-class bare-call fallback in `incremental.ts`), found two more full-build call-resolution fallback strategies present in `src/domain/graph/builder/stages/build-edges.ts`'s `resolveFallbackTargets` that have no equivalent in `src/domain/graph/builder/incremental.ts`'s call-edge resolution path:

1. `resolveKotlinReflectionPreQualified` — Kotlin member callable reference (`Greeter::greet`) pre-qualification, applied *before* the primary `resolveCallTargets` pass.
2. `resolveReflectionKeyExprFallback` — JVM reflection with a literal method-name argument (`getMethod("name")` / `invokeMethod("name")`), keyed off `call.dynamicKind === 'reflection'` and `call.keyExpr`.

`incremental.ts` has zero references to `reflection`, `dynamicKind`, or `keyExpr` — a single-file incremental rebuild (watch mode) of a Kotlin/Java/Scala/Groovy file using either pattern will silently fail to resolve (or resolve differently) the call edge, compared to a full `codegraph build`.

## Where

- `src/domain/graph/builder/stages/build-edges.ts` — `resolveKotlinReflectionPreQualified`, `resolveReflectionKeyExprFallback` (full-build path, has both strategies)
- `src/domain/graph/builder/incremental.ts` — call-edge resolution (incremental path, missing both strategies)

## Suggested fix

Mirror both strategies into the incremental call-resolution path (or, per the architectural fix applied in #1765, hoist the shared logic into `src/domain/graph/builder/call-resolver.ts` so both the full-build and incremental paths call the exact same functions and cannot drift again).

## Found during

Working #1765 — scoped out to keep that PR to a single concern.

carlos-alm:
Duplicate of #1993 — the broader points-to/CHA/dynamic-sink gap this issue also touched on was already fixed by #1852, and the narrower remaining Kotlin-reflection pre-qualify + reflection-keyExpr fallback gap is explicitly tracked by #1993 (referenced directly in `incremental.ts`'s `applyCallFallbacks` doc comment: "tracked separately (#1993), not by this function"). Closing this one to avoid tracking the same remaining work under two issue numbers — #1993 is still open and is the one to pick up.

PR #1947, opened 2026-07-07
fix: add missing same-class bare-call fallback to incremental rebuild

### LINKED ISSUE #1765: Incremental rebuild missing same-class bare-call fallback (full-build/incremental parity gap)
## What

The full-build call-resolution path (`resolveFallbackTargets` in `src/domain/graph/builder/stages/build-edges.ts`, decomposed into named helpers in the phase-11 grind/forge pass) has a same-class bare-call fallback strategy (`resolveSameClassBareCallFallback`): for class-scoped languages (non-JS/TS), a bare call with no receiver that fails global resolution retries qualified as `<CallerClass>.<callName>` — this is how e.g. C# static sibling calls (`IsValidEmail()` inside `Validators.ValidateUser` → `Validators.IsValidEmail`) get resolved.

`src/domain/graph/builder/incremental.ts`'s `applyThisReceiverFallbacks` only implements two strategies:
1. same-class `this.method()` fallback (`resolveThisSameClassTarget`)
2. `Object.defineProperty` accessor fallback (`resolveDefinePropertyTarget`)

There is no bare-call (non-`this`, class-scoped) fallback anywhere in `incremental.ts`, and it doesn't even import `isModuleScopedLanguage`. This means: a C# (or other class-scoped-bare-call language) static sibling call that resolves correctly on a full `codegraph build` will silently fail to resolve (or resolve differently) after an incremental single-file rebuild (watch mode).

## Where

- `src/domain/graph/builder/stages/build-edges.ts` — `resolveSameClassBareCallFallback` (full-build path, has the strategy)
- `src/domain/graph/builder/incremental.ts` — `applyThisReceiverFallbacks` (incremental path, missing the strategy)

## Suggested fix

Add a bare-call fallback branch to `applyThisReceiverFallbacks` (or its call site) that mirrors `resolveSameClassBareCallFallback`, using the same `resolveSameClassQualifiedMethod`-shaped lookup, guarded by `isModuleScopedLanguage(relPath)`.

## Found during

Titan grind phase 11 (adopting forge's `build-edges.ts` decomposition) — codebase-wide duplicate-logic scan surfaced this while comparing the newly-named `resolveSameClassQualifiedMethod`/`resolveSameClassThisFallback`/`resolveSameClassBareCallFallback` trio against incremental.ts's pre-existing `resolveThisSameClassTarget`.

carlos-alm:
Fixes #1765

The full-build call-resolution pipeline (`resolveFallbackTargets` in `src/domain/graph/builder/stages/build-edges.ts`) retries a no-receiver call as `<CallerClass>.<callName>` when global resolution fails — this is how C#/Java-style static sibling calls (e.g. `IsValidEmail()` inside `Validators.ValidateUser`) resolve to `Validators.IsValidEmail`. The incremental single-file rebuild path (`src/domain/graph/builder/incremental.ts`, used by watch mode) only implemented the same-class `this.method()` and `Object.defineProperty` fallbacks — the class-scoped bare-call fallback, and the `isModuleScopedLanguage` import it needs, were entirely missing.

## Fix

Rather than duplicating the fallback logic into `incremental.ts` (which is exactly how it drifted out of sync with `build-edges.ts` in the first place), this consolidates the two same-class fallback strategies into the shared `src/domain/graph/builder/call-resolver.ts` module — already the home of the underlying `resolveSameClassQualifiedMethod` helper both pipelines call:

- `resolveSameClassThisFallback` and `resolveSameClassBareCallFallback` move from `build-edges.ts` (private) to `call-resolver.ts` (exported), unchanged in behavior.
- `build-edges.ts`'s `resolveFallbackTargets` now imports and calls the shared versions instead of its own local copies.
- `incremental.ts`'s fallback dispatcher (renamed `applyThisReceiverFallbacks` → `applyCallResolutionFallbacks`, since it now covers more than `this`-receiver calls) wires in the bare-call strategy between the existing this-fallback and the `defineProperty` accessor fallback, mirroring `resolveFallbackTargets`'s strategy order exactly.

Both pipelines now call the identical exported functions, so they cannot drift apart again.

## Scope note

While investigating, I found two more full-build fallback strategies (`resolveKotlinReflectionPreQualified`, `resolveReflectionKeyExprFallback`) that are also missing from the incremental path. Filed as #1946 — out of scope for this PR to keep it to a single concern.

## Validation

- `npx tsc --noEmit` — clean.
- `npm run lint` — clean.
- New unit tests (`tests/unit/call-resolver.test.ts`) directly exercise `resolveSameClassThisFallback`/`resolveSameClassBareCallFallback` in isolation with a lookup mock scoped to `byNameAndFile` only, proving each fallback resolves via the file-scoped lookup independent of the primary resolution path's own embedded same-class handling.
- New integration test (`tests/integration/issue-1765-incremental-bare-call-fallback.test.ts`) with a C# fixture (two classes, colliding method names) verifies an incremental rebuild produces byte-identical `calls` edges to a full build, including a same-class bare-call resolution and a check that it does **not** false-positive across the sibling class.
- Ran the full test suite after building all 34 WASM grammars locally (this worktree started with an incomplete grammar cache unrelated to this change): 3253 passed, 339 skipped, 0 regressions. The only remaining failures (10 files) require the prebuilt native addon, which isn't installed in this worktree — a pre-existing environment gap, not something this change touches (no Rust/native code changed).
- `tests/benchmarks/resolution/resolution-benchmark.test.ts` (precision/recall across all 34 languages): 206/206 passed.
- `codegraph diff-impact --staged -T` shows a contained blast radius: the two relocated/renamed functions and their transitive callers in `incremental.ts`, plus the new fixture.

### DIFF
--- a/src/domain/graph/builder/call-resolver.ts
+++ b/src/domain/graph/builder/call-resolver.ts
@@ -60,6 +60,52 @@ export function resolveSameClassQualifiedMethod(
     .filter((n) => n.kind === 'method');
 }
 
+/**
+ * Same-class `this.method()` fallback: when the call receiver is `this` and
+ * the primary resolution pass found nothing, derive the enclosing class name
+ * from the caller (e.g. `Logger.info` → class prefix `Logger`) and retry with
+ * the qualified method name `Logger._write`. This mirrors what the native
+ * Rust engine does implicitly via its class-scoped symbol table.
+ *
+ * NOTE: restricted to `this` only — `super.method()` targets a parent class,
+ * not the enclosing class, so qualifying with the child class name would
+ * produce a false edge when the child also defines a same-named method.
+ *
+ * Shared by the full-build (`stages/build-edges.ts`) and incremental
+ * (`incremental.ts`) call-resolution pipelines so the two cannot drift out of
+ * sync (#1765).
+ */
+export function resolveSameClassThisFallback(
+  call: { name: string; receiver?: string | null },
+  callerName: string | null,
+  relPath: string,
+  lookup: CallNodeLookup,
+): Array<{ id: number; file: string; kind?: string }> {
+  if (call.receiver !== 'this' || callerName == null) return [];
+  return resolveSameClassQualifiedMethod(call.name, callerName, relPath, lookup);
+}
+
+/**
+ * Same-class bare-call fallback: when a no-receiver call can't be resolved
+ * globally, try the caller's own class as a qualifier. Handles C# static
+ * sibling calls: `IsValidEmail()` inside `Validators.ValidateUser` resolves
+ * to `Validators.IsValidEmail`. Skipped for JS/TS where bare calls are
+ * module-scoped, not class-scoped.
+ *
+ * Shared by the full-build (`stages/build-edges.ts`) and incremental
+ * (`incremental.ts`) call-resolution pipelines so the two cannot drift out of
+ * sync (#1765).
+ */
+export function resolveSameClassBareCallFallback(
+  call: { name: string; receiver?: string | null },
+  callerName: string | null,
+  relPath: string,
+  lookup: CallNodeLookup,
+): Array<{ id: number; file: string; kind?: string }> {
+  if (call.receiver || callerName == null || isModuleScopedLanguage(relPath)) return [];
+  return resolveSameClassQualifiedMethod(call.name, callerName, relPath, lookup);
+}
+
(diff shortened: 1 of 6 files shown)


carlos-alm:
Closing — this duplicates already in-progress work on #1886 for the same issue (#1765). Continuing the existing stacked chain (#1885 → #1939) instead of a competing branch.

PR #1951, opened 2026-07-07
refactor(structure): share chunk-size statement cache with builder helpers

### LINKED ISSUE #1767: Consider adopting chunk-size statement-cache helper in features/structure.ts::batchUpdateRoles
## Context

Titan grind phase 19 (forge commit 0e83ba03 extracted `getOrCreateBatchStmt`/`runBatchInsert` in `src/domain/graph/builder/helpers.ts`, deduping the WeakMap<db, Map<chunkSize, stmt>> caching shape between `getNodeStmt`/`getEdgeStmt`). The mandatory codebase-wide duplicate scan (grind Step 2c) also adopted the same pattern for a second, pre-existing duplicate between `insertDefinitionsAndExports` (insert-nodes.ts) and `insertBackfilledNodes` (native-orchestrator.ts) — see the `markExportedSymbols` helper added in that commit.

## Finding

`src/features/structure.ts::batchUpdateRoles` (around line 553) has the same shape of chunk-size-keyed statement cache:

```ts
const roleStmtCache = new Map<number, SqliteStatement>();
...
let stmt = roleStmtCache.get(chunkSize);
if (!stmt) {
  const placeholders = Array.from({ length: chunkSize }, () => '?').join(',');
  stmt = db.prepare(`UPDATE nodes SET role = ? WHERE id IN (${placeholders})`);
  roleStmtCache.set(chunkSize, stmt);
}
```

This is structurally identical to the pattern just deduped in `builder/helpers.ts`, but `features/structure.ts` is a different domain (features/, not db/ or builder/), so it was left out of that PR's scope to keep the change focused (one PR = one concern).

## Suggested follow-up

Evaluate whether `batchUpdateRoles` should call a shared chunk-cache primitive too — either by exporting `getOrCreateBatchStmt`-style helper from `builder/helpers.ts` for reuse, or by extracting a standalone `chunkedStmtCache` utility (e.g. in `shared/`) that both `builder/helpers.ts` and `features/structure.ts` can depend on without crossing domain boundaries awkwardly.

Not urgent — this is a code-organization/dedup opportunity, not a bug.

carlos-alm:
Fixes #1767

`batchUpdateRoles` in `src/features/structure.ts` duplicated the `WeakMap<db, Map<chunkSize, stmt>>` cache-getter shape already extracted as `getOrCreateBatchStmt` in `src/domain/graph/builder/helpers.ts` during a prior Titan grind pass.

Since `features/` is a different domain than `domain/graph/builder/`, this extracts the generic helper into `src/db/repository/cached-stmt.ts` as `cachedChunkStmt` — a natural sibling of the existing single-statement `cachedStmt` helper already used across `db/`, `domain/analysis/`, etc. Both `builder/helpers.ts` and `features/structure.ts` now depend on this lower-level `db/` module instead of crossing feature/domain boundaries directly. A matching `ChunkStmtCache<TRow>` type was added next to `StmtCache<TRow>` in `types.ts`.

As a side effect, the role statement cache in `batchUpdateRoles` is now module-scoped and keyed per-db (previously a fresh `Map` was allocated on every call), matching the semantics of the node/edge/export batch caches — repeated `classifyNodeRoles` calls against the same long-lived db connection (watch mode, MCP server) now reuse compiled statements instead of recompiling them every time.

No Rust changes: the native engine's role classification (`crates/codegraph-core/src/graph/classifiers/roles.rs`) already uses rusqlite's built-in `prepare_cached`, so this duplicate-cache pattern only existed on the TypeScript side.

## Validation
- `npx vitest run tests/unit/roles.test.ts` — added two regression tests covering (1) a single role spanning multiple statement chunk sizes (500 + remainder), and (2) cache isolation across two separate db instances — all 22 tests pass
- `npm test` — full suite: 215 files / 3562 tests passed
- `npm run lint` — clean
- `npm run build` — clean (tsc)
- `codegraph diff-impact --staged -T` — blast radius confined to the 6 changed statement-cache/role-classification functions, 21 transitive callers across 5 files
- `codegraph check --cycles` — no cycles introduced

### DIFF
--- a/src/db/repository/cached-stmt.ts
+++ b/src/db/repository/cached-stmt.ts
@@ -1,4 +1,9 @@
-import type { BetterSqlite3Database, SqliteStatement, StmtCache } from '../../types.js';
+import type {
+  BetterSqlite3Database,
+  ChunkStmtCache,
+  SqliteStatement,
+  StmtCache,
+} from '../../types.js';
 
 /**
  * Resolve a cached prepared statement, compiling on first use per db.
@@ -15,6 +20,33 @@ export function cachedStmt<TRow = unknown>(
   if (!stmt) {
     stmt = db.prepare<TRow>(sql);
     cache.set(db, stmt);
+  }
+  return stmt;
+}
+
+/**
+ * Resolve a cached prepared statement for a multi-value INSERT/UPDATE whose
+ * SQL text depends on a chunk size (e.g. the number of `?` placeholders in
+ * an `IN (...)` clause), compiling on first use per db + chunk size.
+ *
+ * `buildSql` is only invoked on a cache miss; subsequent calls with the same
+ * `db`/`chunkSize` pair return the cached statement without re-invoking it.
+ */
+export function cachedChunkStmt<TRow = unknown>(
+  cache: ChunkStmtCache<TRow>,
+  db: BetterSqlite3Database,
+  chunkSize: number,
+  buildSql: (chunkSize: number) => string,
+): SqliteStatement<TRow> {
+  let perDb = cache.get(db);
+  if (!perDb) {
+    perDb = new Map();
+    cache.set(db, perDb);
+  }
+  let stmt = perDb.get(chunkSize);
+  if (!stmt) {
+    stmt = db.prepare<TRow>(buildSql(chunkSize));
+    perDb.set(chunkSize, stmt);
   }
   return stmt;
 }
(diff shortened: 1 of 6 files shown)


carlos-alm:
Closing — this duplicates already in-progress work on #1890 for the same issue (#1767). Continuing the existing stacked chain (#1885 → #1939) instead of a competing branch.

Issue #1990, opened 2026-07-08
Flaky cycle-node ordering in checkNoNewCycles full-suite run (tests/integration/check.test.ts)


carlos-alm:
## Summary

`tests/integration/check.test.ts > checkNoNewCycles > excludeSpeculative has no effect on cycles formed entirely of static import edges` fails consistently when the full suite runs (`npm test`), but passes every time when run in isolation (`npx vitest run tests/integration/check.test.ts -t "..."`).

## Failure

```
AssertionError: expected [ { …(2) } ] to deeply equal [ { nodes: [ …(2) ], …(1) } ]

- Expected
+ Received

  [
    {
      "nodes": [
-       "src/utils.js",
        "src/math.js",
+       "src/utils.js",
      ],
      "speculative": false,
    },
  ]
```

Only the *order* of the two nodes within the reported cycle differs — the cycle itself (same 2 files) is detected correctly both ways. This points to non-deterministic ordering somewhere in cycle detection (e.g. `Map`/`Set` iteration order affected by state left over from earlier tests in the same worker process), not an incorrect result.

## Repro

Run the full suite twice in a row:

```
npm test
```

Both runs fail the same test with the same node-order swap; running the file alone always passes:

```
npx vitest run tests/integration/check.test.ts -t "excludeSpeculative has no effect on cycles formed entirely of static import edges"
```

## Where found

Found while validating an unrelated fix (#1848, native/WASM barrel-only import parity) — the two Rust files touched by that fix (`pipeline.rs`, `import_edges.rs`) have no relationship to cycle detection or `check.test.ts`, and the failure reproduces identically with and without that change. Filing as a separate, pre-existing issue per this repo's scope-discipline policy rather than fixing it as part of #1848.

## Suggested next step

Look for the source of node-order non-determinism in cycle detection (`src/domain/graph/cycles.ts` / `graph/algorithms/tarjan.rs` and however `checkNoNewCycles` orders `cycle.nodes`) — likely a `Set`/`Map`/`HashMap` iteration order that isn't stable across worker-process test execution order. The fix should make node order within a reported cycle deterministic (e.g. sort, or preserve traversal order consistently) rather than just making the test order-insensitive, since inconsistent ordering could also affect real diff/CI output stability.

carlos-alm:
Already resolved on `main`. Investigated as part of a `/fixer` pass, expecting to fix the ordering non-determinism, but found the test itself already fixed.

[...]

Issue #1684, opened 2026-06-21
Engine parity: pts-javascript dispatch-table edges — native emits file-sink instead of resolved dtFn1/dtFn2 targets


carlos-alm:
## Summary

WASM resolves dispatch-table dynamic edges to specific function targets; native emits a sink edge to the file node instead.

## Divergence

Fixture: `pts-javascript` — `dispatch-table.js`

```
wasm=0 native=1: dispatch-table.js:runDispatch(function) -> dispatch-table.js:dispatch-table.js(file) conf=0 dyn=1
wasm=1 native=0: dispatch-table.js:runDispatch(function) -> dispatch-table.js:dtFn1(function) conf=0.9 dyn=1
wasm=1 native=0: dispatch-table.js:runDispatch(function) -> dispatch-table.js:dtFn2(function) conf=0.9 dyn=1
```

WASM correctly resolves the dispatch-table pattern `{a:fnA,b:fnB}[k]()` to the actual target functions (`dtFn1`, `dtFn2`) with conf=0.9, dyn=1 via PTS wildcard resolution. Native emits a single unresolved sink edge to the file node (conf=0, dyn=1).

## Root cause (localization)

Pattern is `wasm=A, native=B` → Rust pipeline missing the dispatch-table PTS resolution. The `dispatch-table` `DynamicKind` is classified in the JS extractor and then resolved in `build-edges.ts` via points-to analysis. The Rust `build_edges.rs` either doesn't receive the `dynamicKind` field (serialization boundary) or doesn't implement dispatch-table resolution.

## Fix

Two steps:
1. Ensure the `dynamic_kind` field from the JS extractor flows through the `NativeFileEntry` → `build_edges.rs` boundary (check `SerializedExtractorOutput` and the `FileEdgeInput` structs).
2. Add dispatch-table resolution in `crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs` to mirror the JS PTS wildcard resolution for `DynamicKind::DispatchTable`.

carlos-alm:
Fixed in PR #1690. The native JS extractor now handles `({a:fnA,b:fnB})[key]()` dispatch-table calls by seeding `arrayElemBindings` under a synthetic `<dt_line_col>` name and emitting `<dt_line_col>[*]`, mirroring the WASM `extractSubscriptCallInfo` RES-2 branch. Parity result: `pts-javascript: wasm vs native OK`.

Issue #1665, opened 2026-06-21
feat(dynamic-calls): RES-2 — closed dispatch table resolution ({a:fnA,b:fnB}[k]())


carlos-alm:
## Problem

`{a:fnA, b:fnB}[k]()` where k is statically resolved to 'a' should emit an edge to fnA.
Currently this is flagged as computed-key and dropped as a sink edge.

## Approach

Over-approximation modeled like CHA:
- When a subscript call's `keyExpr` resolves to a pts literal that points into an object literal, emit edges to all values of that object
- Confidence = baseline - CHA_DISPATCH_PENALTY (0.1)
- Use the existing arr[*] wildcard pattern in points-to.ts: seed pts('<table>[*]')={fnA,fnB}

## Implementation in points-to.ts

Extend `buildPointsToMap` to detect:
```js
const handlers = {a: fnA, b: fnB};
handlers[key]()  // → emit edges to fnA AND fnB at penalized confidence
```

Key: this avoids inventing a new subsystem — the existing Andersen solver can handle it.

## Fixture to update

`tests/benchmarks/resolution/pts-javascript` (precision 1.0 is the guardrail; dispatch table edges go in a separate fixture)

PR #2035, opened 2026-07-08
fix(extractors): resolve inline object-literal dispatch tables on WASM

### LINKED ISSUE #1897: Engine parity: WASM doesn't resolve inline object-literal dispatch tables ({a:fnA,b:fnB}[k]())
## Summary

Discovered while investigating #1771. The native Rust engine has a dispatch-table resolution path for **inline** object-literal subscript calls that the WASM/TS engine does not implement, so the two engines produce different edges for the same source pattern.

## Divergence

```js
function runDispatch(k) {
  return ({ a: fnA, b: fnB })[k]();
}
```

- **Native**: `crates/codegraph-core/src/extractors/javascript.rs`'s `extract_dispatch_table_call` (invoked from `handle_call_expr` for RES-2) detects that the subscript callee's `object` is an inline `object` node (optionally unwrapped from `parenthesized_expression`), synthesizes a `<dt_LINE_COL>` name, records each property's identifier value as an `ArrayElemBinding`, and emits a `<dt_LINE_COL>[*]` call with `dynamic_kind: "dispatch-table"` — which the PTS/wildcard resolver expands into real edges to `fnA`/`fnB` (confidence-penalized, `dynamic=1`).
- **WASM/TS**: `src/extractors/javascript.ts`'s `extractSubscriptCallInfo` never checks whether the subscript's `object` is an `object`-literal node (inline or parenthesized) — it only special-cases string/template-literal and identifier index types. For an inline object-literal target, extraction falls through to a generic `computed-key`/`unresolved-dynamic` classification with no receiver, so no `ArrayElemBinding`/PTS-wildcard wiring occurs and the call resolves to nothing (or a sink edge), not to `fnA`/`fnB`.

Note: the **named-variable** variant of this pattern (`const handlers = {a:fnA,b:fnB}; handlers[k]()`) already works correctly on WASM (via the existing `ArrayElemBinding` + PTS-wildcard mechanism) and was the subject of the now-closed engine-parity issue #1684, which fixed native's handling of *that* case. This new issue is specifically about the **inline** (not variable-bound) object-literal subscript variant, which native handles (RES-2 / `extract_dispatch_table_call`) but WASM does not.

## Impact

For inline dispatch-table subscript calls, native produces real `dtFn1`/`dtFn2`-style edges while WASM produces none/a sink edge — a genuine engine-parity gap per this repo's dual-engine correctness requirement (both engines must produce identical graphs).

## Suggested fix

Add the WASM/TS mirror of `extract_dispatch_table_call` to `src/extractors/javascript.ts`'s `extractSubscriptCallInfo` (or a sibling helper called from the same site): when the subscript object is an `object` node (optionally unwrapped from a `parenthesized_expression`) and the index is a bare identifier, synthesize the same `<dt_LINE_COL>` naming scheme, populate `arrayElemBindings`, and emit a `<dt_LINE_COL>[*]` call with `dynamicKind: 'dispatch-table'` so the existing points-to wildcard resolution path (already used for the named-variable case) picks it up identically to native.

## Where found

Found during root-cause investigation for issue #1771 (dispatch-table function references as object-literal property values) while comparing native's `extract_dispatch_table_call`/RES-2 machinery against the WASM/TS `extractSubscriptCallInfo` implementation. Not fixed as part of #1771 — different syntactic pattern (immediately-subscript-invoked inline object literal vs. a bare identifier stored as a property value and dispatched elsewhere).

carlos-alm:
## Summary

- WASM's `extractSubscriptCallInfo` had no equivalent of native's RES-2 `extract_dispatch_table_call`, so `({a:fnA,b:fnB})[key]()` classified as a generic flagged `computed-key` sink on WASM while native resolved it to `fnA`/`fnB` via the points-to wildcard solver — a genuine engine-parity gap.
- Adds `extractDispatchTableCall` in `src/extractors/javascript.ts`, mirroring `extract_dispatch_table_call` in `crates/codegraph-core/src/extractors/javascript.rs`: when a subscript call's object is an object literal (optionally unwrapped from a parenthesized expression) and the index is a bare identifier, seeds `arrayElemBindings` under a synthetic `<dt_line_col>` name and emits a `<dt_line_col>[*]` call with `dynamicKind: 'dispatch-table'` — the existing points-to wildcard resolution path (already used for array-literal dispatch, e.g. `const arr=[f1,f2]; arr[i]()`) then resolves it to each concrete target identically on both engines.
- Threaded through both WASM extraction paths per this repo's two-path extractor architecture: `dispatchQueryMatch` (used by the WASM worker) and `handleCallExpr` (direct/test callers) both now pass `arrayElemBindings` into `extractCallInfo` → `extractSubscriptCallInfo`.
- Adds `'dispatch-table'` to the `DynamicKind` union in `src/types.ts`, intentionally excluded from `FLAG_ONLY_DYNAMIC_KINDS` (matching native's sink-edge exclusion) so a resolved table never produces a spurious sink edge.

### Note on history

While investigating, I found this exact fix (types.ts union member, `extractSubscriptCallInfo` dispatch-table branch, the `pts-javascript/dispatch-table.js` fixture, and `points-to.test.ts` unit tests) was already implemented and merged once before (commit `9a6b882f`, closing #1665) but silently disappeared from the tree afterward — presumably lost in a merge/rebase — while the native-side mirror (added later in PR #1690, closing #1684) survived. This PR re-implements the WASM side from scratch against the current architecture (the extraction pipeline has changed substantially since then — single-pass collector walk, `runCollectorWalk`, etc.) and restores the dropped fixture/tests, plus adds new query-vs-walk and cross-engine parity regression coverage that didn't exist before.

## Test plan

- `npx vitest run tests/parsers/javascript.test.ts tests/unit/points-to.test.ts tests/engines/query-walk-parity.test.ts` — new unit tests pass (297/297)
- `npx vitest run tests/engines/parity.test.ts` — new cross-engine parity case passes against a locally rebuilt native binary (37 passed, 8 skipped)
- `npx vitest run tests/benchmarks/resolution/resolution-benchmark.test.ts -t "pts-javascript"` — 5/5 pass (restored `dispatch-table.js` + expected-edges entries)
- `npm test` — 3978 passed, 30 skipped, 2 todo, 1 pre-existing known-flaky failure unrelated to this change (`checkNoNewCycles` node-ordering, tracked in #1990)
- `cargo test --release -p codegraph-core` — 591/591 pass (native side untouched; rebuilt binary confirms parity)
- `npx tsc --noEmit` / `npm run lint` — clean (no new issues)

Closes #1897


carlos-alm:
Addressed the one gap named in the summary (Confidence Score 4/5): `TECHNIQUE_MAP` in `tests/benchmarks/resolution/resolution-benchmark.test.ts` had no entry for the new `pts-dispatch-table` mode, so `rollupByTechnique` was falling back to `other` for the two new dispatch-table edges instead of folding them into `points-to`.

Added `'pts-dispatch-table': 'points-to'` to `TECHNIQUE_MAP` (commit 0003dba3). Verified locally with the full 34-language resolution benchmark — the `pts-javascript` fixture's per-technique breakdown now reads:

```
By technique (edges contributed):
    points-to         :  12 resolved /  12 expected  (100.0% recall)
    ts-native         :   3 resolved /   3 expected  (100.0% recall)
```

No `other` bucket remains, confirming the two `runDispatch -> dtFn1/dtFn2` edges are now correctly attributed. As noted in the summary, this only affects the report table — precision/recall thresholds were never gated on it.

Separately: this branch had drifted far behind `main` (several stacked PRs in the same chain were squash-merged one at a time), which surfaced as a large `git merge` conflict unrelated to this PR's own diff. Resolved by merging `origin/main` and reconstructing this PR's own commit (`1d957a40`) on top of main's evolved state — verified the result is diff-identical to the original commit against the new base (`git diff origin/main` reproduces the same 8 files / 274 insertions / 8 deletions as `git show 1d957a40`).

carlos-alm:
@greptileai

Issue #2063, opened 2026-07-08
TypeScript resolution fixture missing expected-edges entries for new X() -> X.constructor attribution (#1892)


carlos-alm:
## Summary

While validating the fix for #1929 (bare `super(...)` extraction), I noticed `tests/benchmarks/resolution/fixtures/typescript/hierarchy.ts`'s constructor-instantiation call sites resolve to **both** the class node and the class's `.constructor` method node:

```
makeCircle@hierarchy.ts -> Circle@hierarchy.ts             (documented in expected-edges.json)
makeCircle@hierarchy.ts -> Circle.constructor@hierarchy.ts (NOT documented)
makeEllipse@hierarchy.ts -> Ellipse.constructor@hierarchy.ts (NOT documented)
makeRectangle@hierarchy.ts -> Rectangle.constructor@hierarchy.ts (NOT documented)
```

The `X.constructor` edges come from the existing constructor-call attribution feature (#1892) — they're correct, pre-existing behavior, unrelated to #1929's bare-`super(...)` fix. They just aren't listed in `expected-edges.json`, so the benchmark's recall metric doesn't credit them and precision headroom is slightly eaten by "extra" resolved edges the test doesn't know about.

## Suggested fix

Add the missing `X.constructor` target entries (mode: `constructor`) alongside the existing `X` class-node entries in `tests/benchmarks/resolution/fixtures/typescript/expected-edges.json`, and audit other language fixtures for the same gap if the #1892 attribution applies there too.

Found opportunistically while working on #1929 — filed separately per this repo's scope-discipline convention rather than folded into that PR.

carlos-alm:
Already resolved — no action needed.

PR #2028 (\`fix(resolver): attribute new ClassName() calls to the constructor method\`, merged 2026-07-10) implemented the underlying #1892 attribution comprehensively and, as part of that, added the exact \`X.constructor\` target entries this issue asked for to \`tests/benchmarks/resolution/fixtures/typescript/expected-edges.json\` (\`makeCircle -> Circle.constructor\`, \`makeEllipse -> Ellipse.constructor\`, \`makeRectangle -> Rectangle.constructor\`, all with \`mode: "constructor"\`), plus the equivalent entries in the javascript, python, csharp, dynamic-typescript, java, php, and dart fixtures — covering the "audit other language fixtures for the same gap" part of the suggested fix too.

Verified: the full resolution benchmark suite (206 tests across all fixtures) passes cleanly on current \`main\`.

[...]

Issue #2062, opened 2026-07-08
resolveThisDispatch resolves this/super dispatch to the wrong file when a same-named base class exists elsewhere with no same-file match


carlos-alm:
## Bug

`resolveThisDispatch` (`src/domain/graph/builder/cha.ts`) resolves `this.method()`/`super.method()`/bare `super(...)` calls by walking the CHA parents map and looking up `${current}.${methodName}` **by name only** (`lookup.byName(qualified)`), across the whole project. It has same-file preference logic, but that logic only helps when a same-file candidate exists *alongside* cross-file candidates:

```ts
const found = lookup.byName(qualified).filter((n) => n.kind === 'method');
if (found.length > 0) {
  if (callerFile && found.some((n) => n.file === callerFile)) {
    return found.filter((n) => n.file === callerFile);
  }
  return found; // <-- BUG: no same-file candidate exists, but a same-named class
                //     in a completely unrelated file is returned anyway, instead
                //     of continuing the walk or returning no match.
}
```

When the qualified name (e.g. `Shape.constructor`) exists in the project **only** in a different file than the caller — because the caller's own file's same-named base class has no explicit definition of that method (e.g. an implicit default constructor) — the walk returns the wrong file's match instead of either continuing up the hierarchy or giving up.

## Repro (observed empirically while validating PR for #1929)

Two independent fixture files, each defining an unrelated class named `Shape`:

`hierarchy.ts`:
```ts
export class Shape {
  area(): number { return 0; }
}
export class Circle extends Shape {
  constructor(private radius: number) { super(); }
  ...
}
```

`super-dispatch.ts` (a different, unrelated hierarchy):
```ts
export class Shape {
  constructor(public name: string) {}
}
export class Polygon extends Shape {
  constructor(name: string, public sides: number) { super(name); }
}
```

Building this project produces:
```
Circle.constructor@hierarchy.ts -> Shape.constructor@super-dispatch.ts       (WRONG — cross-file, unrelated class)
Rectangle.constructor@hierarchy.ts -> Shape.constructor@super-dispatch.ts    (WRONG)
Polygon.constructor@super-dispatch.ts -> Circle.constructor@hierarchy.ts    (WRONG)
Polygon.constructor@super-dispatch.ts -> Rectangle.constructor@hierarchy.ts (WRONG)
Polygon.constructor@super-dispatch.ts -> Ellipse.constructor@hierarchy.ts   (WRONG)
```

[...]

Issue #1892, opened 2026-07-06
Constructor calls (new X()) resolve to the class declaration, not X.constructor


carlos-alm:
## Summary

`new ClassName(...)` call sites never attribute a caller to `ClassName.constructor` — they resolve to the class declaration node itself instead. This is a **universal, pre-existing gap affecting every class in the codebase**, not specific to any one class. Discovered while fixing #1769 (CodeGraph method-attribution bug); the confidence-heuristic fix in that issue does not touch this — it's a distinct name-shape mismatch, confirmed present even at same-file/maximum-confidence distance.

## Root cause

- `handleNewExpr` (`src/extractors/javascript.ts:1417-1433`) emits a call record with `name: ctor.text` — the **bare class name** (e.g. `"CodeGraph"`), never `"CodeGraph.constructor"`.
- Since it's a no-receiver call, resolution goes through `resolveByGlobal` (`src/domain/graph/resolver/strategy.ts:317-336`), which does `lookup.byName(call.name)` with no kind filter — just a confidence check.
- `handleMethodDef` (`src/extractors/javascript.ts:996-1015`) stores the constructor under the *qualified* name `"ClassName.constructor"`, while the class itself is stored under the bare name `"ClassName"`.
- Net effect: `new X()` always resolves the bare name `"X"` to the **class node**, never the `"X.constructor"` **method node** — a name-shape mismatch, not a confidence/proximity problem.

## Evidence

Verified against 3 unrelated classes with real multi-site constructors (`src/shared/errors.ts`), via `codegraph fn-impact <name> -T --json`:

| Class | class-node (direct/transitive/total, role) | `.constructor` (direct/transitive/total, role) |
|---|---|---|
| ConfigError | 23/147/170, core | 0/0/0, dead |
| DbError | 14/183/197, core | 0/0/0, dead |
| EngineError | 6/50/56, core | 0/0/0, dead |

100% consistent — every class's `constructor` method node is structurally unreachable from `new X()` call sites, regardless of file/directory proximity.

## Impact

Any class's `constructor` method always shows `totalDependents: 0` / role `dead`/`dead-unresolved` via `codegraph fn-impact`/`codegraph roles --role dead`, even when the class is constructed extensively throughout the codebase. This makes constructor-specific dead-code signals from `codegraph roles --role dead` unreliable (though most constructors are trivial enough that this hasn't caused practical harm yet — no one is deleting constructors based on this signal).

## Suggested fix direction

[...]

PR #2065, opened 2026-07-08
fix(extractors): extract bare super(...) constructor calls as call edges

### LINKED ISSUE #1929: Bare super(...) constructor calls are never extracted as call edges
## Bug

Bare `super(...)` calls inside a subclass constructor (calling the parent class's constructor) are never extracted as `calls` edges by either engine, unlike `super.method()` dispatch (fixed in #1377/#1315).

## Repro (this repo's own source)

`src/shared/errors.ts` defines `CodegraphError` and 7 subclasses (`ParseError`, `DbError`, `ConfigError`, `ResolutionError`, `EngineError`, `AnalysisError`, `BoundaryError`), each of which calls `super(message, { code: ..., ...opts })` in its constructor:

```ts
export class ParseError extends CodegraphError {
  constructor(message: string, opts: CodegraphErrorOpts = {}) {
    super(message, { code: 'PARSE_FAILED', ...opts });
    this.name = 'ParseError';
  }
}
```

Despite 7 call sites, `CodegraphError.constructor` has zero inbound `calls` edges in the graph:

```
codegraph roles --role dead -T --json | jq '.symbols[] | select(.name == "CodegraphError.constructor")'
# → { "role": "dead-unresolved", ... }
```

Querying the DB directly confirms no `calls` edge targets the node — only structural `contains`/`parameter_of` edges exist.

## Root cause

`extractCallInfo` in `src/extractors/javascript.ts` branches on the call expression's `function` field type (`identifier`, `member_expression`, `subscript_expression`) but has no branch for `fnType === 'super'` (the bare `super(...)` keyword-callee case, distinct from `super.method()` which is a `member_expression` with object `super`). The native Rust extractor (`crates/codegraph-core/src/extractors/javascript.rs`) has the equivalent gap.

## Impact

Any base class whose constructor is only ever reached via `super(...)` (never directly `new`'d) is misclassified `dead-unresolved`, even when actively used throughout the codebase via subclassing.

## Suggested fix

Extend `extractCallInfo` (and the native mirror) to handle `fnType === 'super'`: resolve to the immediate parent class's constructor via the existing extends/CHA chain (typeMap or `classHierarchy`-style lookup), emitting a `calls` edge to `ParentClass.constructor`. Mirror in both WASM/TS and native Rust engines per this repo's dual-engine requirement.

Found while investigating #1784 (instanceof-checks not crediting consumers) — same file (`src/shared/errors.ts`), different mechanism (bare `super()` callee shape vs. `instanceof` binary expression), so filed separately rather than folded into that fix.

carlos-alm:
## Summary

Bare `super(...)` constructor calls (distinct from `super.method()`) were never extracted as call edges by either engine — `extractCallInfo` (WASM/TS) and `extract_call_info` (native Rust) had no branch for the `super` keyword-callee shape, so a class whose constructor is only ever reached via `super(...)` (never directly `new`'d) was misclassified `dead-unresolved`.

- **WASM/TS** (`src/extractors/javascript.ts`): `extractCallInfo` now returns `{ name: 'constructor', receiver: 'super' }` for `fnType === 'super'`, reusing the existing this/super hierarchy dispatch (`resolveThisDispatch` in `cha.ts`) that already resolves `super.method()` — no new resolution path needed. Added the matching `(call_expression function: (super) @callsuper_fn) @callsuper_node` query pattern (mirrored in `src/domain/parser.ts` and `src/domain/wasm-worker-entry.ts`) and a `dispatchQueryMatch` branch, so both the walk path and the query path (used by the WASM worker pool) extract it identically.
- **Native Rust** (`crates/codegraph-core/src/extractors/javascript.rs`): mirrored `extract_call_info`'s new `"super"` arm, and split `handle_call_expr`'s combined `this`/`super` early-return guard so `super(...)` now routes through `extract_call_info` while still skipping callback-reference-call extraction on the arguments (unchanged behavior, still guards against the false cross-file edges that guard was added to prevent).

Both engines now produce the same edge: `Subclass.constructor -> ParentClass.constructor`.

## Out-of-scope findings filed separately

While adding fixture coverage I found a genuine pre-existing bug in the shared `resolveThisDispatch` resolver: when a qualified name like `Shape.constructor` only exists in an unrelated file (no same-file candidate), it wrongly returns the cross-file match instead of continuing the walk or giving up — filed as #2062 with a concrete repro (this is why the TypeScript fixture below uses `Container`/`Box` rather than reusing the existing `Shape` name from `hierarchy.ts`). Also filed #2063 for an unrelated pre-existing fixture-documentation gap (missing `new X() -> X.constructor` expected-edges entries from #1892).

## Test plan

- `tests/parsers/javascript.test.ts` — new describe block covering plain class, class-expression, and callback-reference-arg suppression for bare `super(...)`
- `tests/engines/query-walk-parity.test.ts` — new case asserting the walk and query extraction paths agree on bare `super(...)`
- `tests/integration/phase-8.5-cha-dispatch.test.ts` (`describe.each(['wasm', 'native'])`) — new fixture (`Vehicle.ts`/`Car.ts`) and assertion that `Car.constructor -> Vehicle.constructor` is emitted on **both** engines
- `tests/benchmarks/resolution/fixtures/{javascript,typescript}` — extended `inheritance.js`/`super-dispatch.ts` with bare-`super(...)` constructor chains and their expected edges; also credited a correctly-resolved `Ellipse.constructor -> Circle.constructor` edge in the pre-existing `hierarchy.ts` fixture that this fix now produces
- `crates/codegraph-core/src/extractors/javascript.rs` — new `bare_super_call_extracted_as_constructor_call` Rust unit test; existing `super_call_args_do_not_emit_callback_reference_calls` still passes unchanged
- `npm run lint` — clean
- `npm run build` — clean
- Native addon rebuilt locally (`napi build --platform --release`, codesigned) and verified directly against the fixture

Closes #1929


carlos-alm:
Superseded by #2102 — recreated from a clean branch off current main (see #2101 for why the original branch couldn't be conflict-resolved in place).

Issue #2067, opened 2026-07-08
test: checkNoNewCycles reports the same cycle with non-deterministic node order across calls


carlos-alm:
## Summary

`tests/integration/check.test.ts`'s `checkNoNewCycles > excludeSpeculative has no effect on cycles formed entirely of static import edges` fails intermittently/consistently (reproduced 3/3 runs locally) with a node-order mismatch, not a content mismatch:

```
- Expected
+ Received

  [
    {
      "nodes": [
-       "src/utils.js",
        "src/math.js",
+       "src/utils.js",
      ],
      "speculative": false,
    },
  ]
```

The test:

```ts
test('excludeSpeculative has no effect on cycles formed entirely of static import edges', () => {
  const withSpeculative = checkNoNewCycles(db, new Set(['src/math.js']), false, false);
  const withoutSpeculative = checkNoNewCycles(db, new Set(['src/math.js']), false, true);
  expect(withoutSpeculative.cycles).toEqual(withSpeculative.cycles);
});
```

`checkNoNewCycles` calls `findCycles()` (`src/domain/graph/cycles.ts`) internally and `excludeSpeculative` only applies a `.filter()` afterward (`return opts.excludeSpeculative ? cycles.filter((c) => !c.speculative) : cycles;`) — it never reorders anything. Since both calls query the exact same DB fixture with no writes in between, the two calls should be producing byte-for-byte identical `cycles` arrays (same node order within each cycle), but the node order inside the single reported cycle differs between the two invocations.

## Suspected root cause

Likely a missing `ORDER BY` somewhere in the query that feeds node/edge data into the cycle-detection (Tarjan) algorithm, allowing SQLite to return rows in a different order across two separate query executions of the same underlying SQL — which then changes the traversal order in Tarjan's algorithm and thus the order nodes are pushed into a discovered cycle. Should be investigated in `src/domain/graph/cycles.ts` and `src/graph/algorithms/tarjan.ts`.

## Impact

Cosmetic for callers that only care about which files are in a cycle (most callers), but:
- Makes the affected test flaky/consistently-red depending on execution timing.
- Any caller that compares cycle results for equality (as this test does) or displays them to users could see non-deterministic, confusing output across runs on unchanged code.

## Repro

```
npx vitest run tests/integration/check.test.ts -t "excludeSpeculative has no effect on cycles formed entirely of static import edges"
```

[...]

Issue #2066, opened 2026-07-08
bug: native engine's init_schema() never mirrors migration v20 (edges.dynamic_kind), but do_insert_edges always references it


carlos-alm:
## Summary

`src/db/migrations.ts` migration v20 (added in #1629, commit 5907d588) adds `edges.dynamic_kind`:

```sql
ALTER TABLE edges ADD COLUMN dynamic_kind TEXT;
CREATE INDEX IF NOT EXISTS idx_edges_dynamic_kind ON edges(dynamic_kind) WHERE dynamic_kind IS NOT NULL;
```

`crates/codegraph-core/src/db/connection.rs`'s `MIGRATIONS` array — the Rust mirror of `migrations.ts`, per the file's own doc comment ("IMPORTANT: Migration DDL is mirrored from src/db/migrations.ts. Any changes there MUST be reflected here (and vice-versa)") — only goes up to `version: 19`. The v20 migration was never mirrored; that commit touched only `src/db/migrations.ts` (confirmed via `git show 5907d588 --stat`).

## Why this matters

`crates/codegraph-core/src/db/repository/edges.rs`'s `do_insert_edges` — used for **every single edge insert** on the native path — unconditionally includes `dynamic_kind` in its INSERT column list:

```rust
let sql = format!(
    "INSERT OR IGNORE INTO edges (source_id, target_id, kind, confidence, dynamic, dynamic_kind) VALUES {}",
    placeholders.join(",")
);
```

`NativeDatabase::init_schema()` (`connection.rs`) is the actual native schema-migration runner (mirrors `initSchema()` from `migrations.ts`) — it runs its own `MIGRATIONS` array directly against the native `rusqlite::Connection`, independent of the TS migration runner. Since its `MIGRATIONS` array stops at v19, any database whose schema was initialized purely through the native path (schema_version <= 19, never previously touched by the TS `initSchema()`) will be missing the `dynamic_kind` column entirely — and every native edge insert will fail with `no such column: dynamic_kind`.

This is likely masked in most real-world usage because `db/index.ts`'s `openDb()` calls the TS `initSchema()` (which does include v20) before native code ever touches most databases — but any code path that creates/migrates a fresh DB purely through the native `NativeDatabase.init_schema()` (e.g. a native-first proxy path) would hit this immediately.

## Fix

Add the missing migration to `crates/codegraph-core/src/db/connection.rs`'s `MIGRATIONS` array:

```rust
Migration {
    version: 20,
    up: r#"
[...]

Issue #1806, opened 2026-07-05
follow-up: checkNoSignatureChanges cannot detect exported-symbol loss from full file deletion


carlos-alm:
Found while addressing Greptile review feedback on PR #1792 (fix/titan-check-signature-scope).

**Context:** `checkNoSignatureChanges` in `src/features/check.ts` detects signature changes by matching diff `oldRanges` line numbers against exported symbol declaration lines queried from the current (post-change) DB state: `SELECT ... FROM nodes WHERE file = ? AND kind IN (...) AND exported = 1`.

When a file is deleted entirely, the graph builder purges all of its rows from `nodes` (`DELETE FROM nodes WHERE file = ?` in `src/db/repository/build-stmts.ts`) before `check` ever runs. So even if `parseDiffOutput` correctly tracked the deleted file's removed line ranges, the `WHERE file = ?` lookup against the post-change DB would always return zero rows for that file — the check can never flag an exported symbol that was removed by deleting its entire containing file. This is a false-negative gap: deleting a file that still has external callers elsewhere in the codebase silently passes the signature-change gate.

This is a pre-existing limitation of the DB-driven query approach (it predates PR #1792's `DiffLineTracker` refactor — the old `pushHunkRanges` implementation had the same gap, since the lookup is against the current DB regardless of how ranges are computed). Fixing it requires comparing against symbol data captured *before* the deletion (e.g. a pre-change DB snapshot or re-parsing the pre-image blob via `git show`), which is an architectural change beyond the scope of a diff-parsing bug fix.

**Suggested fix:** capture the set of exported symbols in files slated for deletion (from the pre-change DB state, before the incremental rebuild purges them) and check whether any had callers outside the deleted file, before running the incremental rebuild in `codegraph check`'s pipeline.

PR #2068, opened 2026-07-08
fix: persist deleted-export advisories so check survives purge ordering

### LINKED ISSUE #1938: follow-up: checkNoDeletedExportsInUse misses violations once a rebuild has already purged the deleted file
**Context:** #1806 added `checkNoDeletedExportsInUse` (`src/features/check.ts`) to detect an exported symbol lost via full-file deletion, when that symbol still has an external consumer. It works by querying the *current* DB for the deleted file's exported symbols and their cross-file callers — which only works while the DB still reflects pre-purge state.

**The remaining gap:** `codegraph check` never triggers a rebuild itself (`checkData` only opens the DB read-only). Whether `checkNoDeletedExportsInUse` can see a deleted file's exports depends entirely on whether some *other*, separate `codegraph build` invocation has already run and purged that file's `nodes` rows (`DELETE FROM nodes WHERE file = ?`, run unconditionally by `detectChanges` for any file no longer found on disk — see `src/domain/graph/builder/stages/detect-changes.ts`) by the time `check` runs.

In this repo's own hook-driven workflow specifically: a file deletion via `rm`/`git rm` (a Bash tool call) does not itself trigger `update-graph.sh` (which only fires on Edit/Write PostToolUse). But if *any other* file is edited afterward in the same session before the commit — extremely common — `update-graph.sh` fires, runs `codegraph build`, and that build scans the whole working tree, notices the deleted file is gone, and purges it well before `pre-commit.sh`'s check ever runs. More generally, any pipeline that runs `codegraph build && codegraph check` (a very natural sequence) hits this ordering on every single deletion.

Once purged, `checkNoDeletedExportsInUse` finds zero rows for that file and silently reports no violation — the exact false-negative gap #1806 set out to close, just narrowed to "only when check runs before the next rebuild" instead of "always."

**Suggested fix (deferred from #1806 as out of scope for that PR):** capture the pre-purge state durably, inside the build pipeline itself, so detection survives regardless of invocation order. Candidate approach: at the point `detectChanges` computes `removed: string[]` (before `purgeFilesFromGraph`/`purgeFilesData` deletes the rows), record each removed file's exported symbols plus whether each has an external (cross-file) consumer, into a small persistent side-channel (e.g. a dedicated table, or a `.codegraph/` journal file) that `codegraph check` can read in addition to live DB rows. Needs:
- A lifecycle/invalidation story (advisories should not accumulate forever or leak across unrelated commits once the deletion is actually committed or otherwise resolved).
- Mirroring in the native Rust engine (`crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs`, `db/connection.rs`) per this repo's dual-engine parity requirement, since purging happens on both paths.

This is meaningfully larger than a diff-parsing fix — it changes what the build pipeline persists, not just how `check` reads existing data — hence tracked separately rather than folded into #1806.

carlos-alm:
## Summary

`checkNoDeletedExportsInUse` (#1806) detects an exported function/method/class lost when a file is deleted in its entirety, but only by querying the *current* DB for that file's rows. Once *any* `codegraph build` purges the deleted file's `nodes`/`edges` rows — `detectChanges` does this unconditionally for any file no longer found on disk, regardless of whether `codegraph check` has run yet — the predicate has nothing left to query and the violation silently goes undetected. This is purely a function of external invocation ordering (e.g. this repo's own `update-graph.sh` hook firing a plain `codegraph build` before `pre-commit.sh`'s `check` runs), not something the predicate itself can control.

Closes #1938

## Fix

Capture a durable, purge-order-independent snapshot instead of relying on live rows surviving until `check` runs:

- New `deleted_export_advisories` table (migration v21, mirrored in both `src/db/migrations.ts` and `crates/codegraph-core/src/db/connection.rs`) stores, per deleted file, each lost export's cross-file consumers.
- `detectChanges` (`domain/graph/builder/stages/detect-changes.ts`) captures this snapshot at the exact point it computes the removed-file set — **before** purging — via new `recordDeletedExportAdvisories`/`clearDeletedExportAdvisories` in `db/repository/deleted-export-advisories.ts`. A file that reappears under the same path has its stale advisory cleared before its nodes are (re)inserted, so a resolved deletion never misattributes a stale violation.
- `checkNoDeletedExportsInUse` now queries live `nodes` first (unchanged behavior when nothing has purged yet) and falls back to the persisted advisory table only when a deleted file's live rows are already gone.
- Extracted the shared "exported function/method/class definitions" and "cross-file consumers of a node" queries into `db/repository/nodes.ts`/`edges.ts` (`findExportedDefinitions`, `findExternalConsumers`) so both the live-DB path and the advisory-capture path use identical semantics.

**Native engine parity:** the native Rust orchestrator's incremental fast path (`crates/codegraph-core/.../pipeline.rs`'s `save_and_purge_changed`) bypasses the JS `detectChanges` stage entirely for a plain incremental build — this is the *default* code path when the native engine is available. Mirrored the capture/clear logic there (`record_deleted_export_advisories`/`clear_deleted_export_advisories` in `detect_changes.rs`) so both engines produce identical results. Verified via a real `buildGraph()` end-to-end regression test that runs against both engines.

While implementing, found and fixed a real bug this surfaced: the Rust `table_queryable` existence-probe used `query_row(...).is_ok()`, which returns `Err(QueryReturnedNoRows)` — not `Ok` — for a table that exists but is still empty, causing the whole advisory capture to silently no-op on a freshly-migrated table. Caught by the new Rust unit tests; fixed by treating `QueryReturnedNoRows` as "exists".

## Test plan

- `cargo test --lib` (crates/codegraph-core) — 649 passed, including 8 new tests for `record_deleted_export_advisories`/`clear_deleted_export_advisories`
- `npm test` — 4128 tests, 1 pre-existing unrelated failure (node-order non-determinism in `checkNoNewCycles`, filed as #2067), everything else green including 30 new tests across `tests/unit/deleted-export-advisories.test.ts`, the new `checkNoDeletedExportsInUse` advisory-fallback describe block in `tests/integration/check.test.ts`, and a new end-to-end regression test (`tests/integration/issue-1938-deleted-export-advisory-persistence.test.ts`) exercising both WASM and native engines via a real `buildGraph()` → delete → rebuild → `checkData` flow
- `npm run lint` — clean (only a pre-existing unrelated warning in an untouched fixture file)
- `codegraph diff-impact --staged -T` — 28 symbols changed, 13 callers affected across 6 files, no cycles, no dead exports

## Out-of-scope findings filed separately

- #2066 — native engine's `init_schema()` never mirrors TS migration v20 (`edges.dynamic_kind`), even though `do_insert_edges` always references that column
- #2067 — `checkNoNewCycles` reports the same cycle with non-deterministic node order across two calls against the same DB state

Stacked on #1937/#1929 (base branch `fix/issue-1929-bare-super-constructor-calls-are-never`) — only the check/detect-changes/migrations/test diff described above is this issue's change.

### DIFF
--- a/crates/codegraph-core/src/db/connection.rs
+++ b/crates/codegraph-core/src/db/connection.rs
@@ -337,6 +337,25 @@ const MIGRATIONS: &[Migration] = &[
         // dataflow_vertices and dataflow_summary on the next `codegraph build`.
         up: "SELECT 1",
     },
+    // NOTE: TS migration v20 (edges.dynamic_kind) has no Rust counterpart yet — see #2066.
+    // v21 below is independent of it (a standalone new table), so this gap does not block it.
+    Migration {
+        version: 21,
+        up: r#"
+      CREATE TABLE IF NOT EXISTS deleted_export_advisories (
+        id INTEGER PRIMARY KEY AUTOINCREMENT,
+        file TEXT NOT NULL,
+        name TEXT NOT NULL,
+        kind TEXT NOT NULL,
+        line INTEGER NOT NULL,
+        consumer_name TEXT NOT NULL,
+        consumer_file TEXT NOT NULL,
+        consumer_line INTEGER NOT NULL,
+        deleted_at INTEGER NOT NULL
+      );
+      CREATE INDEX IF NOT EXISTS idx_deleted_export_advisories_file ON deleted_export_advisories(file);
+    "#,
+    },
 ];
 
 // ── napi types ──────────────────────────────────────────────────────────
--- a/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
@@ -253,6 +253,13 @@ fn save_and_purge_changed(
     let removed_file_neighbors =
         detect_changes::capture_removed_file_neighbors(conn, &change_result.removed);
 
+    // A file about to be (re)inserted can no longer be "deleted" — clear any
+    // stale advisory left over from a prior removal at this same path before
+    // capturing this build's actual removals, and before purging deletes the
+    // live evidence `record_deleted_export_advisories` reads (#1938).
+    detect_changes::clear_deleted_export_advisories(conn, &changed_paths);
+    detect_changes::record_deleted_export_advisories(conn, &change_result.removed);
+
     let files_to_purge: Vec<String> = change_result
         .removed
         .iter()
--- a/crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs
@@ -814,6 +814,138 @@ pub fn capture_removed_file_neighbors(conn: &Connection, removed_files: &[String
     neighbors.into_iter().collect()
 }
 
+/// True if `table` can be queried without error. Used for
+/// `deleted_export_advisories` (added in migration v21), which an older,
+/// not-yet-migrated schema may not have.
+fn table_queryable(conn: &Connection, table: &str) -> bool {
+    // `query_row` returns `Err(QueryReturnedNoRows)` for a query that is
+    // syntactically/semantically valid but matches zero rows — e.g. a table
+    // that exists but is still empty — which is NOT the same as the table
+    // being absent (a real `SqliteFailure`/"no such table" error). Unlike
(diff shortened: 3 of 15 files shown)


carlos-alm:
Superseded by #2103 — recreated from a clean branch off current main (see #2101 for why the original branch couldn't be conflict-resolved in place).

PR #2069, opened 2026-07-09
fix(perf): resolve config once in withRepo to avoid double loadConfig

### LINKED ISSUE #1941: perf: withRepo/fn-impact.ts has the same double-loadConfig pattern as withReadonlyDb
Follow-up from PR #1883's Greptile review (https://github.com/optave/ops-codegraph-tool/pull/1883#discussion_r3528... — comment on `src/domain/analysis/query-helpers.ts:19`).

That PR fixed `withReadonlyDb` + its callers (`exports.ts`, `context.ts`) to resolve config once via the new `resolveDbConfig()` export and pass it through to `resolveAnalysisOpts({ ...opts, config: opts.config ?? config })`, avoiding a redundant `loadConfig()` call per query.

`fn-impact.ts` has the identical pattern but through the `withRepo` path instead:
- `withRepo(customDbPath, fn)` → `openRepo(customDbPath)` → `resolveDbSettings(customDbPath, opts.engine)` → `loadConfig(deriveRootDirFromDbPath(customDbPath))`
- `fn-impact.ts:280` then separately calls `resolveAnalysisOpts(opts)`, which calls `loadConfig()` again (cwd-rooted) when `opts.config` is absent.

Fixing this is out of scope for #1883 because `openRepo`/`resolveDbSettings` in `db/connection.ts` are used by every native-engine call site (not just read-only analysis queries), so widening `withRepo`'s callback signature to also pass `config` has a much larger blast radius than `withReadonlyDb`'s change. Worth doing as its own focused PR.

carlos-alm:
## Summary

- `withRepo()` (backing `fn-impact.ts`, `brief.ts`, `dependencies.ts`, `implementations.ts`) opens its `Repository` via `openRepo()`, which internally resolves config via `resolveDbSettings()` for engine/busy-timeout selection. Its two callers that also needed analysis-tuning config — `fnImpactData` via `resolveAnalysisOpts()`, and `briefData` via a direct `loadConfig()` call — triggered a *second*, independent `loadConfig()` for the same DB path.
- Beyond the redundant work, that second call resolved its rootDir from `process.cwd()` instead of the resolved `--db` path, so `--db /other/repo/.codegraph/graph.db` would silently read the *wrong* project's `.codegraphrc.json` for `analysis.fnImpactDepth`/`analysis.briefCallerDepth` — the same correctness gap already fixed for `withReadonlyDb()`'s callers (`exports.ts`, `context.ts`).
- Fix: widen `withRepo()` to resolve config once and thread it through to both `openRepo()` (via a new, additive `opts.config` — every other `openRepo()` call site is unaffected) and the callback, mirroring `withReadonlyDb`'s `(db, config) => T` signature. `fnImpactData` and `briefData` now reuse `opts.config ?? dbConfig` instead of calling `loadConfig()` again.
- `dependencies.ts` and `implementations.ts` — the other two `withRepo` callers — don't read config, so they're unaffected; TypeScript allows their existing `(repo) => T` callbacks to ignore the new second parameter.

No native/Rust changes: this is a TS-only config-resolution optimization internal to `db/connection.ts` / `domain/analysis/query-helpers.ts` orchestration — the native crate has no equivalent concept to mirror.

Fixes #1941

## Test plan

- New `tests/unit/withrepo-config-resolution.test.ts`: spies on `loadConfig` to assert `withRepo`, `fnImpactData`, and `briefData` each call it exactly once (previously twice for the latter two); also asserts `analysis.fnImpactDepth`/`analysis.briefCallerDepth` are read from the `--db` path's rootDir even when `process.cwd()` points elsewhere. Verified these tests fail against the pre-fix code (2 loadConfig calls / config not threaded) and pass after the fix.
- `npx tsc --noEmit` — clean
- `npm run lint` — clean
- `codegraph diff-impact --staged -T` — confirms only the intended 5 functions changed (`resolveDbSettings`, `openRepo`, `briefData`, `fnImpactData`, `withRepo`), 43 callers across 33 files, all backward-compatible
- `npm test` — 225/259 files and 3688/4134 tests pass. The remaining 20 failing files (6 individual test failures) all share one root cause unrelated to this change: `Cannot find module '@optave/codegraph-darwin-arm64'` — this worktree has no built native addon, a pre-existing, already-documented environment gap for isolated worktrees, not something this diff touches.

### DIFF
--- a/src/db/connection.ts
+++ b/src/db/connection.ts
@@ -441,6 +441,9 @@ export function resolveConfigForDbPath(customDbPath?: string): CodegraphConfig {
  * Derives rootDir from the resolved DB path so loadConfig reads the right project config.
  * Shared by openRepo() and openReadonlyWithNative() so the two call sites can't drift.
  *
+ * Accepts an optional pre-resolved `config` for callers that already loaded it for the
+ * same `customDbPath` (e.g. withRepo()), avoiding a second findDbPath()/loadConfig() call.
+ *
  * MUST be called before opening any DB handle: loadConfig can throw (e.g. ConfigError
  * via resolveSecrets on a malformed llm.apiKeyCommand config), and an already-open
  * handle at that point would never be closed.
@@ -448,8 +451,8 @@ export function resolveConfigForDbPath(customDbPath?: string): CodegraphConfig {
 function resolveDbSettings(
   customDbPath: string | undefined,
   engineOpt: 'native' | 'wasm' | 'auto' | undefined,
+  config: CodegraphConfig = resolveConfigForDbPath(customDbPath),
 ): ResolvedDbSettings {
-  const config = resolveConfigForDbPath(customDbPath);
   // config.build.engine is already populated from CODEGRAPH_ENGINE env by applyEnvOverrides,
   // so this covers both the env-var path and the .codegraphrc.json config-file path.
   return {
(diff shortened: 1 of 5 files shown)


carlos-alm:
Superseded by #2104 — recreated from a clean branch off current main (see #2101 for why the original branch couldn't be conflict-resolved in place).

PR #2097, opened 2026-07-11
fix(tracer): preserve target file permissions and surface copy failures in sedi()

carlos-alm:
## Summary
- Follow-up to #2047, which merged before Greptile's re-review of its final push (the Windows cross-device-rename fix) had posted. That re-review flagged two real robustness gaps in the rewritten `sedi()` helper in `tests/benchmarks/resolution/tracer/tracer-common.sh` — both confirmed and fixed here.
- `mv "$tmp" "$file"` replaces the target's inode with the scratch file's, so the target picks up `mktemp`'s restrictive `0600` mode instead of keeping its own permissions.
- A failed `mv` was not checked, so `sedi()` reported success (and leaked the scratch file) even when the edit never landed.
- Switches to `cp "$tmp" "$file"`: POSIX `cp` writing into an existing destination overwrites its content while leaving the destination's own inode — and therefore its permission bits — untouched, and the exit status is now checked so a failed copy surfaces as a failure. The scratch file is removed unconditionally afterward instead of only in the failure branch, since `cp` (unlike `mv`) never consumes it.

## Test plan
- [x] `npx vitest run tests/benchmarks/resolution/tracer/` — 47/47 pass
- [x] Local isolated bash tests confirming: permission bits preserved across an edit (0755 stays 0755, not reset to mktemp's 0600), a failed copy (read-only target) is reported as a failure with no leaked scratch file and the original content left untouched
- [x] `shellcheck tests/benchmarks/resolution/tracer/tracer-common.sh` — clean
- [x] Real end-to-end run of `native-tracer.sh` for rust and swift fixtures via `run-tracer.mjs` — correct edge counts, no regression


carlos-alm:
Both CI failures were unrelated to this PR's `sedi()` fix:

- **Pre-publish benchmark gate**: failing because the `dev` branch's benchmark comparisons against the pinned `3.15.0` baseline had drifted (source file count grew ~60% since the 3.15.0 release), tripping the `Full build` / `1-file rebuild` regression thresholds for every PR. This was already root-caused and fixed separately in #2107 (merged 2026-07-16), which adds the appropriate `KNOWN_REGRESSIONS` exemptions.
- **CI Testing Pipeline**: just the aggregator check, which failed as a consequence of the benchmark gate failure above.

Merged `main` into this branch to pick up #2107's fix — no conflicts, `tests/benchmarks/resolution/tracer/tracer-common.sh` is untouched by the merge. All 22 checks are now green, and 47/47 tracer tests + shellcheck pass locally.

PR #2114, opened 2026-07-16
fix(ci): allow-list Publish trigger events in benchmark.yml instead of excluding 'push'

### LINKED ISSUE #2109: fix: benchmark.yml's push-exclusion filter will break once Publish's dev-build trigger changes
## Correction to original writeup

The original version of this issue claimed `benchmark.yml` runs the full expensive benchmark suite on every Publish completion (~8/day, tracking merge frequency). That's **not accurate** — both jobs already gate on `github.event.workflow_run.event != 'push'`:

```yaml
record-benchmarks:
  if: >-
    github.event_name == 'workflow_run' &&
    github.event.workflow_run.conclusion == 'success' &&
    github.event.workflow_run.event != 'push'

embedding-benchmark:
  if: >-
    github.event_name == 'workflow_dispatch' ||
    (github.event.workflow_run.conclusion == 'success' &&
     github.event.workflow_run.event != 'push')
```

So today, a push-triggered Publish completion already short-circuits both jobs — the workflow "runs" (shows up in run counts) but does essentially no work for dev builds. No fix was needed for the original concern.

## The real problem

optave/ops-codegraph-tool#2108 changes Publish's dev-build trigger from `push` to `schedule`. `!= 'push'` does **not** exclude `'schedule'` — so once #2108 merges, a scheduled dev-build Publish completion would newly satisfy both jobs' conditions:

- `record-benchmarks` would attempt to download the `benchmark-files`/`benchmark-results-json` artifacts, but `pre-publish-benchmark` in `publish.yml` only runs for `release`/`workflow_dispatch` events (not `schedule`) — so those artifacts won't exist, and the download step will fail.
- `embedding-benchmark` would kick off a redundant 2.5+ hour embedding benchmark run against the latest **stable** npm version (its `workflow_run` branch always resolves to the latest non-dev tag) once a day, for no reason since nothing changed on npm.

## Fix

Replace the negative exclusion (`!= 'push'`) with a positive allow-list of the event names that actually produce the upstream artifacts/warrant a benchmark run (`release`, `workflow_dispatch`), so the filter doesn't silently drift out of sync the next time Publish's trigger set changes.

Depends on optave/ops-codegraph-tool#2108 (same underlying change).

carlos-alm:
## Summary
- Both jobs in `benchmark.yml` gated on `github.event.workflow_run.event != 'push'`. That's a negative exclusion, not a positive allow-list of the event types that actually warrant a benchmark run.
- #2113 changes Publish's dev-build trigger from `push` to `schedule` — `!= 'push'` doesn't exclude `'schedule'`, so once that merges, scheduled dev-build Publish completions would newly pass this filter, and:
  - `record-benchmarks` would try to download `benchmark-files`/`benchmark-results-json` artifacts that don't exist for schedule-triggered Publish runs (only `release`/`workflow_dispatch` go through `pre-publish-benchmark`), failing the job
  - `embedding-benchmark` would kick off a redundant 2.5+ hour run against the latest stable npm version once a day, for nothing
- Fix: replace the exclusion with an explicit allow-list (`event == 'release' || event == 'workflow_dispatch'`) so it can't silently drift out of sync with Publish's trigger set again

## Why
Direct correctness dependency of #2113 (Publish dev-build scheduling). Also corrects my own earlier issue #2109, which incorrectly assumed this workflow ran the full benchmark suite on every Publish completion — it already didn't, this fix just keeps that true after #2113.

Closes #2109
Depends on #2113 (should merge together or #2113 first)

## Test plan
- [x] YAML parses, `actionlint` shows no new findings (pre-existing shellcheck notes only)
- [ ] After both PRs merge: verify the first scheduled dev-build Publish completion does *not* trigger `record-benchmarks` or `embedding-benchmark`
- [ ] Verify a real release still triggers both as before


carlos-alm:
Addressed Claude's review feedback: none required — Greptile's review (5/5 confidence, "Safe to merge") did not flag any findings or gaps against the workflow diff itself.

What was done since that review:
- CI had 2 failing checks (`CI Testing Pipeline`, `Pre-publish benchmark gate`), both caused by `tests/benchmarks/regression-guard.test.ts` flagging `Full build`/`1-file rebuild` timing regressions against the stale 3.15.0 baseline — this branch predated the fix for that (PR #2107, already on `main`).
- Merged `origin/main` into this branch (no conflicts) to pick up #2107's baseline-drift exemption. Verified the merge did not alter this PR's actual change (the `benchmark.yml` allow-list diff is identical before/after the merge).
- Ran `npm run lint` and the regression-guard test suite locally to confirm green before pushing.

carlos-alm:
@greptileai

Issue #2108, opened 2026-07-16
cost: full 6-platform native build + dev release runs on every merge to main


carlos-alm:
## Problem

`.github/workflows/publish.yml` triggers on `push: branches: [main]` in addition to `release: types: [published]`. That means **every single merge to main** (not just tagged releases) runs:

- `preflight`: full `npm test` + a from-source native addon build (ubuntu)
- `build-native`: the full native matrix across ~6 platform targets (2x Linux incl. cross-compiled arm64, macos-latest, macos-14, windows-latest)
- `publish-dev`: packages tarballs for all 6 platforms and creates a GitHub **pre-release** (`dev-v<version>`), pruning to the last 5

Correction to the original writeup: this does **not** publish to the npm registry — only `release`/`workflow_dispatch` events reach the `publish` job that does `npm publish`. The dev path is a GitHub release with tarballs, consumed by things like the `dogfood` skill ("dev from GitHub release tarballs"). The expensive part — the 6-platform native cross-compilation in `preflight` + `build-native` — still runs on every merge regardless.

Measured: 242 Publish runs in the last 30 days (~8/day), tracking 1:1 with merge frequency, not release frequency.

## Why it matters

This repo is public, so GitHub Actions minutes are free on standard hosted runners (confirmed via billing API — all sampled runs report `billable_ms: 0`) and this cost is 100% discounted on the actual invoice. But the org's usage report still lists it at gross/pre-discount cost, and it's reportedly ~75% of what the bill would be without that discount — plus it's a lot of real, repeated compute (6-platform Rust cross-compilation, ~8x/day) for a dev channel that doesn't need per-commit freshness.

## Proposed fix

Decouple the expensive native-build + dev-release path from every push:
- Move the dev-build trigger from `push` to a scheduled cadence (coalesces any number of merges between ticks into a single build)
- Skip the scheduled run entirely if there are no new commits since the last dev release (avoids no-op builds on quiet days)
- Leave the `release: published` (stable) and `workflow_dispatch` (manual retry) paths untouched

## Related

Part of a broader CI/CD cost-reduction effort:
- #2109 — Benchmark fires on every Publish completion
- #2110 — native build matrix duplicated across 3 workflows
- #2111 — CI's 3-OS matrix runs unconditionally
- #2112 — claude.yml automated-review has no scope gating (the one confirmed non-discounted $ cost)

PR #2115, opened 2026-07-16
fix(ci): extract reusable workflow for the native cross-compilation matrix

### LINKED ISSUE #2110: cost: native distribution build matrix duplicated between build-native.yml and publish.yml
## Correction to original writeup

The original version of this issue lumped `ci.yml`'s \`native-host-build\`/\`parity\` jobs in with this. On closer read, those serve a genuinely different purpose: a 3-OS **host-native** build (no cross-compilation, no musl/arm64 variants) that also runs \`cargo test --release\`, feeding the JS test suite and the WASM/native parity check. That's not the same thing and shouldn't be force-merged into this — tracked separately if it needs attention.

## The real duplication

\`build-native.yml\`'s \`build\` job and \`publish.yml\`'s \`build-native\` job both define the **same 6-target cross-compilation matrix** (linux-x64, linux-arm64 cross-compiled, linux-x64-musl cross-compiled, macos-latest, macos-14, windows-latest) with near-identical steps — \`publish.yml\`'s version just adds one "Sync Cargo.toml version" step before the build. Any change to the target list, toolchain setup, or macOS cargo-shim workaround has to be made in both places (drift risk — they've already diverged slightly in \`os: windows-latest\` vs the pinned \`windows-2022\` used elsewhere).

## Proposed fix

Extract a reusable `workflow_call` workflow (`build-native-matrix.yml`) parameterized by an optional `version` input (skips the Cargo.toml sync step when unset), and have both `build-native.yml` (PR validation) and `publish.yml` (release/dev builds) call it instead of maintaining two copies.

Part of the broader CI/CD cost-reduction effort (see #2108).

carlos-alm:
## Summary
- `build-native.yml`'s `build` job and `publish.yml`'s `build-native` job each defined the same 6-target cross-compilation matrix (linux-x64, linux-arm64 cross, linux-x64-musl cross, macos-latest, macos-14, windows-latest) with near-identical steps — publish.yml's copy just adds a "Sync Cargo.toml version" step
- Extracted `.github/workflows/build-native-matrix.yml` as a `workflow_call` reusable workflow with an optional `version` input (the Cargo.toml sync step is skipped when unset)
- `build-native.yml` and `publish.yml`'s `build-native` job now both call it instead of maintaining two copies

## Why
Part of the broader CI/CD cost-reduction effort (#2108). Beyond the duplicated compute, the two copies had already started drifting (`windows-latest` here vs the pinned `windows-2022` used in `ci.yml`'s separate host-build matrix) — a single source of truth removes that risk for the target list, toolchain setup, and the macOS cargo-shim workaround.

Note: `ci.yml`'s `native-host-build`/`parity` jobs are intentionally **not** folded into this — they build a different thing (3-OS host-native build + `cargo test`, no cross-compilation/musl/arm64 variants) for a different purpose (JS test + WASM/native parity validation), not distributable platform binaries.

Closes #2110

## Test plan
- [x] YAML parses for all three files
- [x] `actionlint` — no new findings (pre-existing shellcheck notes only, all outside the touched regions)
- [ ] Next PR touching `crates/**` should confirm `build-native.yml` still produces artifacts named `native-<platform>` as before
- [ ] Next Publish run (dev or release) should confirm `build-native` job still produces the same artifacts with the version-synced Cargo.toml


carlos-alm:
@greptileai

PR #2116, opened 2026-07-16
fix(ci): skip native-host-build when no native-relevant paths changed

### LINKED ISSUE #2111: cost: CI's 3-OS matrix runs unconditionally regardless of whether native code changed
## Problem

\`.github/workflows/ci.yml\` runs three jobs — \`native host build\`, \`test\`, and \`engine parity\` — each matrixed across \`ubuntu-latest\`, \`macos-latest\`, \`windows-2022\` (9 parallel legs total), unconditionally on every push to main and every PR. 880 CI runs in the last 30 days.

For PRs that only touch TypeScript/docs/config with no effect on the native engine, the macOS and Windows legs of \`native host build\` and \`engine parity\` are re-verifying something that can't have changed — the native addon build and WASM/native parity behavior are only affected by changes under \`crates/**\`, \`src/domain/parser.ts\`, extractor code, or the native loader (\`src/infrastructure/native.ts\`).

## Proposed fix

Scope the macOS/Windows legs of \`native host build\` and \`engine parity\` to only run when the diff touches native-relevant paths (mirroring the path filter already used in \`build-native.yml\` and \`perf-canary.yml\`). For pure-TS/doc-only PRs, fall back to an Ubuntu-only leg (or skip those jobs entirely, relying on \`build-native.yml\`'s own path-scoped trigger for actual native changes).

This is likely the single biggest lever for reducing total CI compute-hours, since most day-to-day PRs in this repo don't touch \`crates/**\`.

Part of the broader CI/CD cost-reduction effort (see #2108).

carlos-alm:
## Summary
- \`native-host-build\` (3-OS matrix: ubuntu/macos/windows) ran an unconditional Rust build + \`cargo test\` on every push to main and every PR, regardless of whether the diff touched anything native. \`test\`, \`parity\`, and \`pre-publish-benchmark\` all depend on its artifacts.
- Added a \`detect-changes\` job that diffs the PR base (or the previous push SHA) for \`crates/**\`, \`Cargo.toml\`, \`Cargo.lock\`, and the native addon loader (\`src/infrastructure/native.ts\`)
- \`native-host-build\` now skips entirely (all 3 OSes) when none of those changed — **fail-open**: if the detector itself fails or is inconclusive, the build still runs, so a broken detector can't silently hide a real Rust regression
- \`test\`, \`parity\`, and \`pre-publish-benchmark\` keep running on **all 3 OSes regardless** — they just fall back to the last-published native binary (via the normal \`npm install\` optional-dependency resolution) instead of a freshly PR-built one when \`native-host-build\` was skipped. This preserves full cross-platform test/parity coverage for every PR; only the redundant Rust rebuild+\`cargo test\` is cut for non-native changes.

## Why
Part of the broader CI/CD cost-reduction effort (#2108). This is the biggest single lever in that effort — most day-to-day PRs in this repo don't touch \`crates/**\`, so they were paying for a 6-leg macOS/Windows/Linux native rebuild (plus \`cargo test\` on each) for no reason. 880 CI runs in the last 30 days were all paying this cost unconditionally.

Closes #2111

## Test plan
- [x] YAML parses, \`actionlint\` shows no new findings (pre-existing shellcheck notes only, unrelated lines)
- [ ] A pure-TS/docs PR should show \`native-host-build\` skipped and \`test\`/\`parity\`/\`pre-publish-benchmark\` still green on all 3 OSes
- [ ] A PR touching \`crates/**\` should show \`native-host-build\` still running on all 3 OSes as before
- [ ] Push to main (no native changes) should show the same skip behavior


carlos-alm:
Thanks for the review — confirming the fail-open design and the tag-based diff are intentional, as noted. CI on 485dd63e now shows `native-host-build` correctly running on all 3 OSes (previously it was wrongly skipped, which was the root cause of the Test/Engine-parity failures on the prior push — the fallback binary had drifted 62 native commits past the pinned v3.15.0 release). Watching Test/Engine-parity/pre-publish-benchmark now that a fresh binary is available.

Issue #2135, opened 2026-07-20
bug: audit <file> reports "No file matching" for real files that are pure re-export barrels


carlos-alm:
## Found during dogfooding v3.16.0

**Severity:** Low
**Command:** `codegraph audit <file>`

## Reproduction

```bash
codegraph build .
codegraph where -f src/domain/graph/builder.ts    # confirms file IS in the graph, 77 importers
codegraph deps src/domain/graph/builder.ts        # confirms file IS in the graph, 0 imports, 77 importers
codegraph audit src/domain/graph/builder.ts       # => "No file matching \"src/domain/graph/builder.ts\""
```

`src/domain/graph/builder.ts` is a real, tracked file in the graph (both `where -f` and `deps` find it immediately with a full importer list). It's a pure re-export barrel with zero function/class/method-kind symbols of its own — everything it exports is defined in `src/domain/graph/builder/*.ts` and re-exported.

## Expected behavior

`audit` on a real, graph-tracked file with zero own function-kind definitions should say something like "no functions to audit in this file (0 own definitions)" — distinguishing "this file isn't in the graph at all" from "this file exists but has nothing at the function level to analyze" (e.g. a barrel/re-export-only file, a pure type-declaration file, etc).

## Actual behavior

`audit <file>` prints the exact same "No file matching" message used for files that genuinely aren't in the graph, which is misleading — it reads as "codegraph doesn't know this file," when actually the file is fully tracked.

## Root cause

- `src/features/audit.ts` `auditData()` calls `explainData(target, ..., {depth:0})`, which for a file-like target calls `explainFileImpl(db, target, ...)` — this queries only function/method/class-kind definitions in that file, with no separate check for whether the file itself exists as a tracked node.
- `src/presentation/audit.ts:99-100` then renders `data.functions.length === 0` as `No file matching "${target}"` unconditionally, conflating "file not in graph" with "file in graph, zero own function-kind symbols."

## Suggested fix

In `auditData`/`explainFileImpl`, check whether the file exists as a node in the graph independently of whether it has function-kind children, and have `presentation/audit.ts` render a distinct message for "file exists, 0 functions to audit" vs. "file not found in graph."

PR #2142, opened 2026-07-20
fix(audit): distinguish "file not found" from "file has zero functions"

carlos-alm:
## Summary
`codegraph audit <file>` printed the identical "No file matching" message whether the file genuinely isn't tracked in the graph, or the file is real and fully tracked but just has zero own function/method/class definitions (e.g. a pure re-export barrel) — misleadingly implying the graph is missing the file entirely.

## Found during
Dogfooding v3.16.0 — see #2135

## Test plan
- [x] `npx vitest run tests/integration/audit.test.ts` — 19/19 passing (added 3 new edge-case tests covering the `found` field)
- [x] Manually verified against a real build: barrel file → "No functions to audit..."; nonexistent file → "No file matching..."; normal file → unaffected

### DIFF
--- a/src/features/audit.ts
+++ b/src/features/audit.ts
@@ -162,7 +162,18 @@ export function auditData(
   }
 
   if (results.length === 0) {
-    return { target, kind: explained.kind as 'function' | 'file', functions: [], found: false };
+    // `found` reflects whether the target exists in the graph at all -- check
+    // the pre-filter `explained.results`, not the post `--file`/`--kind` filtered
+    // `results`. Otherwise a real, tracked function filtered out entirely by
+    // e.g. `--file src/other.js` would be misreported as "not found" (the same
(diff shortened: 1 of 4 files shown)


carlos-alm:
Addressed Greptile's finding: the `found: false` early-return was checking the post-`--file`/`--kind`-filtered `results.length`, so a function that genuinely exists in the graph but is entirely filtered out by those options was misreported as "not found" — the same bug this PR fixes, reachable via a different input shape.

Fixed in 5929ef08: `found` now reflects `explained.results.length === 0` (pre-filter), not the post-filter `results.length`. Added a regression test (`--file filter removing every match does not set found: false`) covering the exact scenario from the report (function exists in one file, filtered by `--file` pointing elsewhere).

carlos-alm:
@greptileai

PR #2210, opened 2026-08-02
fix(native): mirror JS migration v20 (edges.dynamic_kind) in Rust init_schema (#2001)

### LINKED ISSUE #2001: Rust native init_schema() never adds edges.dynamic_kind column (JS migration v20 has no Rust mirror)
## Summary

`src/db/migrations.ts` migration version 20 adds `edges.dynamic_kind` (`ALTER TABLE edges ADD COLUMN dynamic_kind TEXT` + index) — used to tag dynamic-call sink edges (e.g. `unresolved-dynamic`, `reflection`, `computed-key`, `value-ref`, `eval`).

The Rust mirror, `NativeDatabase::init_schema()` in `crates/codegraph-core/src/db/connection.rs` (doc comment: "Mirrors `initSchema()` from `src/db/migrations.ts`"), has no equivalent migration — its `MIGRATIONS` const stops at `version: 19` (the P6 full-rebuild sentinel) and never adds `dynamic_kind`.

Despite this, many native extractors actively write to that column: `crates/codegraph-core/src/extractors/{java,dart,groovy,scala,lua}.rs` (and others) set `dynamic_kind: Some(...)` on emitted calls, and `db/repository/edges.rs`'s bulk-insert SQL explicitly includes `dynamic_kind` in its `INSERT OR IGNORE INTO edges (...)` column list.

## Impact

On any database whose schema was initialized/migrated exclusively through the Rust path (`init_schema()`) without ever having JS's `migrations.ts` apply migration 20 first, `edges` has no `dynamic_kind` column — so every edge insert that includes it should fail (rusqlite "no such column" style error), silently dropping the dynamic-call sink edges these extractors are specifically designed to emit. In practice this may be masked by JS always opening/migrating the DB before native inserts run in today's pipeline, but that's an implicit ordering dependency, not a guarantee — and it means the Rust engine's schema self-sufficiency is broken for this column.

## Suggested fix

Add a `version: 20` migration to Rust's `MIGRATIONS` const mirroring the JS `ALTER TABLE edges ADD COLUMN dynamic_kind TEXT` + `CREATE INDEX idx_edges_dynamic_kind` exactly, per the file's own "any changes here MUST be reflected there" contract.

## How this was found

Found while investigating issue #1865 (`reconnectReverseDepEdges` sibling-count edge case) — out of scope for that fix, filing separately per repo scope-discipline convention.

### LINKED ISSUE #2066: bug: native engine's init_schema() never mirrors migration v20 (edges.dynamic_kind), but do_insert_edges always references it

carlos-alm:
## Summary

Rust's `MIGRATIONS` array (`crates/codegraph-core/src/db/connection.rs`) stopped at `version: 19`, never adding the `edges.dynamic_kind` column that `src/db/migrations.ts`'s migration v20 adds. Despite this, several native extractors (java, dart, groovy, scala, lua, ...) set `dynamic_kind` on emitted calls, and `db/repository/edges.rs`'s bulk-insert SQL unconditionally references the column — so any database whose schema was initialized purely through the native `init_schema()` path (never touched by JS's `initSchema()` first) would fail every edge insert with `no such column: dynamic_kind`.

## Fix

Added the missing `version: 20` migration, mirroring the JS `ALTER TABLE`/`CREATE INDEX` exactly, in the correct array position between v19 and v21 — the apply loop tracks `current_version` incrementally through the array in iteration order, so the new entry must sit between them, not after v21 (which would have been silently skipped otherwise since `20 > 21` is false).

## Verification

- Opened a fresh `:memory:` database via `NativeDatabase` directly, called `init_schema()` with **no prior JS migration involved at all**, and confirmed `edges.dynamic_kind` now exists (`PRAGMA table_info(edges)` includes it)
- Added a Rust unit test (`init_schema_adds_edges_dynamic_kind_column`) covering this exact scenario for durable regression coverage
- `cargo test --release` (crates/codegraph-core): 670 passed
- `npm test`: 260 files, 4207 passed, 30 skipped, 2 todo, 0 failed
- `npm run lint` / `cargo build --release`: clean, no warnings

This also fixes #2066, an independently-filed duplicate of this exact gap (same root cause, same suggested fix).

Closes #2001
Closes #2066

### DIFF
--- a/crates/codegraph-core/src/db/connection.rs
+++ b/crates/codegraph-core/src/db/connection.rs
@@ -337,8 +337,13 @@ const MIGRATIONS: &[Migration] = &[
         // dataflow_vertices and dataflow_summary on the next `codegraph build`.
         up: "SELECT 1",
     },
-    // NOTE: TS migration v20 (edges.dynamic_kind) has no Rust counterpart yet — see #2066.
-    // v21 below is independent of it (a standalone new table), so this gap does not block it.
+    Migration {
+        version: 20,
+        up: r#"
+      ALTER TABLE edges ADD COLUMN dynamic_kind TEXT;
+      CREATE INDEX IF NOT EXISTS idx_edges_dynamic_kind ON edges(dynamic_kind) WHERE dynamic_kind IS NOT NULL;
+    "#,
+    },
     Migration {
         version: 21,
         up: r#"
@@ -1618,4 +1623,36 @@ fn row_to_json(
         map.insert(col_names[i].clone(), val);
     }
     serde_json::Value::Object(map)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Regression for #2001/#2066: a database whose schema was initialized
+    /// purely through the native `init_schema()` (never touched by the TS
+    /// `initSchema()` in src/db/migrations.ts) must still end up with
+    /// `edges.dynamic_kind` — migration v20 was missing from Rust's
+    /// `MIGRATIONS` array entirely, so `do_insert_edges`'s unconditional
+    /// `dynamic_kind` column reference would fail with "no such column" on
+    /// any DB that only ever ran the native migration path.
+    #[test]
+    fn init_schema_adds_edges_dynamic_kind_column() {
+        let db = NativeDatabase::open_read_write(":memory:".to_string(), None)
+            .expect("open_read_write should succeed for :memory:");
+        db.init_schema().expect("init_schema should succeed");
+
+        let conn = db.conn().expect("connection should still be open");
+        let mut stmt = conn.prepare("PRAGMA table_info(edges)").unwrap();
+        let has_dynamic_kind = stmt
+            .query_map([], |row| row.get::<_, String>(1))
+            .unwrap()
+            .filter_map(Result::ok)
+            .any(|name| name == "dynamic_kind");
+
+        assert!(
+            has_dynamic_kind,
+            "edges.dynamic_kind column missing after native-only init_schema()"
+        );
+    }
 }
--- a/crates/codegraph-core/src/db/connection.rs
+++ b/crates/codegraph-core/src/db/connection.rs
@@ -744,6 +744,21 @@ impl NativeDatabase {
                 let _ =
                     conn.execute_batch("ALTER TABLE edges ADD COLUMN dynamic INTEGER DEFAULT 0");
             }
+            // #2001/#2066: version-gated migration v20 alone cannot repair a
+            // native-only database whose schema_version was already advanced
+            // past 20 by the pre-fix MIGRATIONS array (which jumped straight
+            // from v19 to v21, never applying v20) — its stored version being
+            // >= 21 makes the `migration.version > current_version` gate skip
+            // v20 forever on every subsequent init_schema() call. This
+            // unconditional, reality-checked backfill (mirroring the
+            // confidence/dynamic columns just above) repairs those databases
+            // too, not just fresh ones.
+            if !has_column(conn, "edges", "dynamic_kind") {
+                let _ = conn.execute_batch("ALTER TABLE edges ADD COLUMN dynamic_kind TEXT");
+            }
+            let _ = conn.execute_batch(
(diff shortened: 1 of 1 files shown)


carlos-alm:
Good catch — fixed in 0d0ba059. Added `edges.dynamic_kind` to the existing unconditional "legacy column compat" backfill block (the same one that already repairs `edges.confidence`/`edges.dynamic` for pre-migration-system databases), which runs after the version-gated migration loop regardless of `schema_version`. This repairs databases already stamped past v20 by the pre-fix `MIGRATIONS` array, not just fresh ones.

Added a regression test (`init_schema_repairs_edges_dynamic_kind_on_a_database_already_past_v20`) that simulates the exact prior-bug state — schema stamped to version 23, `dynamic_kind` dropped back out — and confirms `init_schema()` repairs it.

carlos-alm:
@greptileai

carlos-alm:
Good catch — fixed in 341894c2. Propagated the `ALTER TABLE edges ADD COLUMN dynamic_kind` failure instead of swallowing it with `let _ =`. Since `dynamic_kind` is referenced unconditionally by every edge insert (`db/repository/edges.rs`), a silently failed repair would have let `init_schema()` report success while leaving edge inserts broken with a much harder-to-diagnose "no such column" error at insert time. Left the partial-index creation (`CREATE INDEX ... idx_edges_dynamic_kind`) as best-effort since it's non-critical for correctness — by the time it runs the column is guaranteed to exist.

carlos-alm:
@greptileai

Issue #2214, opened 2026-08-02
Rust: if-let pattern-binding call-assignments don't propagate method-call return types cross-file


carlos-alm:
## Context

While fixing #2007 (Rust `crate::`/`self::`/`super::` import-path resolution), I verified the fix end-to-end against `tests/benchmarks/resolution/fixtures/rust/`. Recall went from 20/24 (83.3%) to 23/24 (95.8%) on both engines, with zero precision regression and full dual-engine parity maintained. One edge remains unresolved on both engines:

```
main.rs::main -> models.rs::User.display_name (receiver-typed)
```

## Root cause (distinct from #2007 — verified, not the import-resolution gap)

`main.rs`:
```rust
if let Some(user) = service.get_user(1) {
    println!("Found: {}", user.display_name());
}
```

Unlike the three now-fixed edges (`main -> UserService.add_user/get_user/remove_user`, all typed via the single-hop `let service = build_service();` bare-function-call return-type propagation), this case needs **two** additional capabilities neither engine has today:

1. Recognizing `if let Some(x) = expr` (or `if let Ok(x) = expr`) as a call-assignment binding `x` to `expr`'s return type, unwrapped from `Option<T>`/`Result<T, _>` — today's call-assignment extraction (`match_rust_call_assignments` in `crates/codegraph-core/src/extractors/rust_lang.rs` and its TS mirror) only recognizes plain `let x = expr();`.
2. Propagating the return type of a **method call** on an already-typed receiver (`service.get_user(1)` where `service: UserService`, requiring `UserService.get_user`'s return type `Option<User>` from `service.rs`'s return-type map), not just a **bare function call** (`build_service()`). `propagate_return_types_across_files`/`propagateReturnTypesAcrossFiles`'s current logic (`pipeline.rs:1526-1660`, `build-edges.ts:531-589`) only handles the bare-call case via `imported_map.get(&callee_name)`.

## Suggested approach

- Extend the Rust extractor's call-assignment matcher to also recognize `if let Some(x)`/`if let Ok(x)` (and their TS mirror) as call-assignments, unwrapping the outer `Option`/`Result` type when seeding the type map.
- Extend `propagate_return_types_across_files`/`propagateReturnTypesAcrossFiles` to resolve a call-assignment's callee through the type map first (method call on a typed receiver) before falling back to the bare global-function lookup, threading through the already-resolved receiver's defining file to look up the method's return type.

[...]

PR #2215, opened 2026-08-02
fix(resolve): resolve Rust crate::/self::/super:: module paths to real files

### LINKED ISSUE #2007: Rust use crate::… import paths never resolve to real files — module-function edges rely on accidental global-name fallback
## Summary

While fixing #1876 (receiver-typed calls + trait dispatch through local vars), I traced why cross-file return-type propagation (Phase 8.2 — `returnTypeMap`/`callAssignments`, the same generic mechanism the JS/TS extractor already uses) does not resolve `let service = build_service();` in `main.rs` even though `build_service` is declared in `service.rs` and imported via `use crate::service::build_service;`.

Root cause: `resolveImportPathJS` (`src/domain/graph/resolve.ts`) has no handling for Rust's `crate::`/`self::`/`super::`-prefixed module paths. For any import source that doesn't start with `.` (which covers every Rust `use` path — Rust has no relative-path import syntax), it falls through to the workspace/`exports` resolution branches (npm-package semantics) and, finding no match, **returns the literal unresolved import-source string** (e.g. `"crate::service::build_service"`) as if it were a resolved file path.

This means `buildImportedNamesMap`'s `importedNames` map for a Rust file maps each imported local name to this bogus non-file string, not to the actual `.rs` file that declares it.

## Why this was hidden until now

Rust's actual "module-function" call edges (e.g. `UserService.add_user -> create_user` via `use crate::models::create_user;`) still resolve today — but not through real import-aware resolution. They fall through `resolveCallTargets`'s import-aware branch (which fails silently, since `lookup.byNameAndFile(name, "crate::models::create_user")` never matches a real file), all the way to `resolveExactGlobalMatch`'s bare global-name fallback, which happens to succeed because the callee names in the fixture are globally unique. This is coincidental, not by design — any Rust project with two same-named functions in different modules would silently produce zero or wrong edges for this pattern, and it also means Rust import edges/confidence scoring never get the `importedFrom === targetFile` proximity bonus real import resolution provides.

It also blocks legitimate cross-file features that depend on knowing the real target file, such as the return-type propagation added in #1876 — `main -> UserService.add_user/get_user/remove_user` (typing `service` from `build_service()`'s cross-file return type) remains unresolved specifically because of this gap, not because of anything specific to receiver typing.

## Fix direction

Add a Rust-aware branch to `resolveImportPathJS` (and the native Rust mirror in `resolve_import.rs`/equivalent) that maps `crate::a::b::name` / `self::a::name` / `super::a::name` paths to real files by walking the project's module tree (`mod` declarations → file or `dir/mod.rs`/`dir.rs` conventions), analogous to how Python's dotted-module imports or Go's package-path imports are resolved elsewhere in this file. This is a non-trivial addition (Rust's module system has its own visibility/path rules) — scoping it as its own issue rather than folding it into #1876's receiver-typing fix.

## Evidence

Fixture: `tests/benchmarks/resolution/fixtures/rust/`. After #1876's fix, `resolution-benchmark.test.ts`'s rust suite sits at 100% precision / 83.3% recall (20/24), with the remaining 4 false negatives all being `main -> UserService.*`/`User.display_name` edges that depend on this import-path gap (see `expected-edges.json`, `mode: "receiver-typed"` entries citing "local var typed by return value").

carlos-alm:
## Summary

`resolveImportPathJS` (and the mirrored native `resolve_import_path_inner`) had no handling for Rust's `use`-path syntax. Any non-relative import source (which covers every Rust `use` path — Rust has no relative-import syntax) fell straight through workspace/`exports` resolution (npm-package semantics) and, finding no match, returned the literal unresolved string as if it were a resolved file path. This broke cross-file return-type propagation (Phase 8.2) for any bare function call whose target was imported via `use crate::...`, since the resolver never learned which file actually declared it.

## Changes

- Adds `crate::`/`self::`/`super::` resolution to both engines (`src/domain/graph/resolve.ts` and `crates/codegraph-core/src/domain/graph/resolve.rs`), walking the project's directory tree per Rust's module-file conventions (`mod.rs`/`lib.rs`/`main.rs` vs. sibling `foo/` directories), gated on the project's known-files list.
- Handles both extractor shapes for `use` sources: the single-item form (`crate::service::build_service`, trailing item name with no file of its own) and the braced-list form (`crate::models`, module path only).
- Also fixes the native `resolve_import_path` singular entry point (and its `resolve_import` FFI binding), which unconditionally dropped `known_files` even when the caller had a real set available (`ImportEdgeContext` already holds one) — making Rust path resolution structurally impossible on that code path regardless of the batch-path fix.
- Threads `ctx.allFiles` through the two TS call sites where it's cheaply available (`getResolved`, `findBarrelCandidates` in `resolve-imports.ts`).

## Verification

Rebuilt both `dist/` and the native `.node` addon, then verified end-to-end against `tests/benchmarks/resolution/fixtures/rust/` on both engines directly (not just the threshold-gated benchmark test, which was already passing at the old 83.3% recall):

- **Before**: 20/24 expected edges matched (83.3% recall) on both engines.
- **After**: 23/24 matched (95.8% recall) on both engines, zero precision regression (0 spurious extra edges), full dual-engine parity maintained (both engines produce identical results).

The one remaining gap (`main -> User.display_name`) needs method-call return-type propagation through `if let Some(x) = ...` pattern bindings — a distinct mechanism from import-path resolution. Filed as #2214.

## Test plan
- [x] New Rust unit tests in `resolve.rs` (12 new, incl. crate/self/super, braced vs. single-item shapes, dead-end fallback, the singular-path regression) — `cargo test --lib domain::graph::resolve`: 57/57 pass
- [x] New TS unit tests in `tests/unit/resolve.test.ts` (9 new) — 77/77 pass
- [x] Full Rust suite — `cargo test --lib`: 682/682 pass
- [x] `cargo clippy --lib` — no new warnings
- [x] `npm run lint` — clean
- [x] `npm test` — full suite, 261 files / 4272 tests pass (includes the full 34-language resolution benchmark)
- [x] Direct end-to-end verification against the Rust fixture on both engines (see above)

Closes #2007

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/stages/import_edges.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/import_edges.rs
@@ -67,6 +67,7 @@ impl ImportEdgeContext {
             import_source,
             &self.root_dir,
             &self.aliases,
+            Some(&self.known_files),
             Some(&self.workspaces),
         )
     }
--- a/crates/codegraph-core/src/domain/graph/resolve.rs
+++ b/crates/codegraph-core/src/domain/graph/resolve.rs
@@ -539,7 +539,10 @@ fn walk_rust_module_segments(
 /// the project's directory tree per Rust's module-file conventions (issue
 /// #2007). Returns `None` (falls through to the bare-specifier fallback)
 /// when `known_files` isn't available or no match is found — e.g. `#[path]`
-/// attribute overrides, which this convention-based resolver doesn't model.
+/// attribute overrides, or a Cargo.toml `[[bin]]`/`[[example]]`/`[[test]]`/
+/// `[[bench]]` section declaring a target at a custom, non-conventional
+/// path (issue #2217), neither of which this convention-based,
+/// known-files-only resolver models.
 fn resolve_rust_use_path(
     from_file: &str,
     import_source: &str,
--- a/crates/codegraph-core/src/domain/graph/resolve.rs
+++ b/crates/codegraph-core/src/domain/graph/resolve.rs
@@ -376,18 +376,52 @@ fn rust_module_dir(file: &str) -> String {
     }
 }
 
-/// Find the crate-root .rs file (main.rs or lib.rs) whose directory is an
-/// ancestor of `from_file`, walking up from `from_file`'s directory and
-/// stopping at `root_dir`. Returns the absolute path, or `None` if no crate
-/// root is found among `known_files` — scoping to the nearest ancestor
-/// crate root (rather than a project-wide search) correctly handles a
-/// Cargo workspace with several crates, each resolving `crate::` relative
-/// to its own root.
+/// Cargo directory names whose direct .rs children are each their own,
+/// independent crate root — a separate binary/example/test/bench target,
+/// never sharing a `crate::` module tree with `src/main.rs`/`src/lib.rs` or
+/// with each other. `foo/main.rs` nested one level inside one of these
+/// (a multi-file binary/example) is already handled by the ordinary
+/// main.rs/lib.rs search below and doesn't need this special case.
+const CARGO_STANDALONE_TARGET_DIRS: [&str; 4] = ["bin", "examples", "tests", "benches"];
+
+/// True if `file` is a standalone Cargo target root — a `.rs` file directly
+/// inside `src/bin/`, `examples/`, `tests/`, or `benches/` (not itself
+/// named main.rs/lib.rs, which the ordinary crate-root search already finds).
+fn is_rust_cargo_target_root(file: &str) -> bool {
+    let path = Path::new(file);
+    let stem = path.file_stem().and_then(|s| s.to_str()).unwrap_or("");
+    if stem == "main" || stem == "lib" || stem == "mod" {
+        return false;
+    }
+    let Some(parent) = path.parent() else {
+        return false;
+    };
+    let Some(dir_name) = parent.file_name().and_then(|s| s.to_str()) else {
+        return false;
+    };
+    CARGO_STANDALONE_TARGET_DIRS.contains(&dir_name)
+}
+
+/// Find the crate-root .rs file whose directory is an ancestor of
+/// `from_file`, walking up from `from_file`'s directory and stopping at
+/// `root_dir`. Returns the absolute path, or `None` if no crate root is
+/// found among `known_files` — scoping to the nearest ancestor crate root
+/// (rather than a project-wide search) correctly handles a Cargo workspace
+/// with several crates, each resolving `crate::` relative to its own root.
+///
+/// A standalone Cargo target file (`src/bin/foo.rs`, `examples/foo.rs`,
+/// `tests/foo.rs`, `benches/foo.rs`) is its own crate root regardless of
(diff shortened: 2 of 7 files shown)


carlos-alm:
Thanks for the careful review — both findings are valid.

**Issue 2 (Cargo target roots) — fixed in 278fd12a.** `findRustCrateRoot`/`find_rust_crate_root` now recognize standalone Cargo target files (`src/bin/foo.rs`, `examples/foo.rs`, `tests/foo.rs`, `benches/foo.rs`) as their own independent crate root, returning the file itself immediately instead of walking up and wrongly attributing `crate::` to an unrelated `src/main.rs`/`src/lib.rs` elsewhere in the project (each such file compiles as its own crate and never shares a module tree with the package's main binary/library or its sibling targets). Same fix applied to `rustParentModuleFile`/`rust_parent_module_file` for `super::` — these files have no parent module to walk up to. Added regression tests on both engines confirming this, and re-verified end-to-end against the fixture: still 23/24, zero regression.

**Issue 1 (incremental path) — filed as #2216, not fixed in this PR.** You're right that `incremental.ts`'s three `resolveImportPath(...)` call sites (and the Rust equivalents) never thread `knownFiles`/`known_files` through at all, so a Rust file rebuilt incrementally would silently miss these edges even though a full rebuild resolves them correctly. This is a materially larger, separately-scoped plumbing gap: unlike the full-build path (which already collects the project's file list in memory before import resolution runs), the incremental path has no in-memory project-wide file list at all — it would need a DB query (or a project-wide list threaded down from the watcher loop) added across several function signatures (`rebuildFile`, `rebuildEdgesForTargetFile`, `buildImportEdges`, `emitEdgesForImport`, `runReverseDepCascade`, `buildImportedNamesMap`, plus the Rust mirrors). Filing separately per repo scope-discipline rather than expanding this PR.

carlos-alm:
@greptileai

carlos-alm:
The custom-Cargo.toml-target-path case you raised is real and is the same class of limitation as the \`#[path]\` attribute override this resolver already documented as out of scope: a convention-based, known-files-only resolver can't know a target lives at a non-conventional path without parsing Cargo.toml metadata it never reads. That's a materially different mechanism (TOML parsing + a target-path map, not file-tree convention) from this PR's fix — filed as #2217, and added a doc-comment note in 38c7c5e2 alongside the existing \`#[path]\` note so this boundary is visible in the code, not just the tracker.

On the incremental-path gap (#2216): still tracked separately and unchanged from my last reply — it's genuinely a larger, separately-scoped plumbing task (no in-memory project-wide file list exists on that path today, for *any* language, not just Rust), not something this PR's resolver-logic fix can reasonably absorb.

To be clear about this PR's actual scope: it fixes \`crate::\`/\`self::\`/\`super::\` resolution for the full-build path (verified 23/24 on the real fixture, up from 20/24, on both engines with zero precision regression) and the conventional Cargo target layout. It was never scoped to close every convention-based-resolver edge case in one pass — Cargo.toml path overrides and incremental-path threading are both real, distinct follow-on features with their own tracked issues.

carlos-alm:
@greptileai

Issue #1744, opened 2026-07-02
Incremental rebuild leaves edges.technique NULL for native-engine call edges (full build sets 'ts-native')


carlos-alm:
## Summary

Discovered while verifying phase-11 Titan decomposition of `src/domain/graph/builder/stages/build-edges.ts` (a pure extract-method refactor, unrelated to this bug — reproduces identically on unmodified `HEAD`).

For the **native engine**, a call edge's `technique` column differs depending on whether it was inserted via a full build vs. an incremental rebuild of the same final source state:

- Full build (`codegraph build --engine native` or `--no-incremental`): `applyEdgeTechniquesAfterNativeInsert` in `build-edges.ts` backfills `technique = 'ts-native'` on every natively-inserted `calls` edge that doesn't already have an explicit technique.
- Incremental rebuild (`codegraph build --engine native` after a single-file edit): `buildCallEdges` → `emitIncrementalCallEdges` in `src/domain/graph/builder/incremental.ts` inserts edges via `stmts.insertEdge.run(...)`, which never sets `technique` at all — it's left `NULL`.

## Repro

1. Build a small fixture project fully with `--engine native`.
2. Make a trivial one-line edit to any file that has a `calls` edge into it (e.g. append an unrelated comment).
3. Rebuild incrementally (`codegraph build --engine native`).
4. Dump the `edges` table and compare `technique` for the affected edges against a fresh `--no-incremental` full rebuild of the same post-edit state.

Example divergence on `tests/fixtures/sample-project`:
```
incremental: {"src":"main","tgt":"Calculator.compute", ..., "technique":null}
full:        {"src":"main","tgt":"Calculator.compute", ..., "technique":"ts-native"}
```
All other edge fields (source, target, kind, confidence, dynamic) are identical — only `technique` diverges.

## Impact

Any query or feature that filters/groups by `edges.technique` (e.g. `codegraph roles`, technique-coverage reporting in the resolution benchmark, coverage dashboards) will undercount native-engine edges that were last touched by an incremental rebuild vs. a full rebuild, even though the edges themselves are correct.

## Suggested fix

Mirror `applyEdgeTechniquesAfterNativeInsert`'s backfill behavior in `incremental.ts`'s `emitIncrementalCallEdges` (or run an equivalent scoped backfill after incremental native inserts), so `technique` is populated consistently regardless of which rebuild path produced the edge.

PR #2221, opened 2026-08-02
fix: give reverse-dep edge reconnection a content-hash identity signal

### LINKED ISSUE #2015: reconnectReverseDepEdges: simultaneous add+remove of same-named siblings (net-zero count change) is undetectable from line data alone
## Summary

Follow-up from #1865 (fixed by PR #2014). The fix there replaces the ordinal/nearest-line heuristic with a dominant-line-shift alignment that correctly reconnects reverse-dep edges when a same-(name, kind) sibling group shifts AND its size changes by a same-named sibling being added or removed in the same edit — verified for both the add-only and remove-only cases on both engines.

There is one remaining theoretical gap, not covered by #1865's literal reproduction: if a same-named/same-kind sibling is **both** added and removed in the same edit (e.g. sibling B is renamed away while a new, unrelated sibling E is added elsewhere in the same group), the group's sibling **count** stays the same (e.g. 4 -> 4), so the alignment takes the "count unchanged -> match by rank" fast path — but the *identity* has actually changed. Since nothing about raw line numbers can distinguish "this is genuinely the same declaration, just shifted" from "a different declaration happens to now occupy this rank," this compound case can still silently mis-wire an edge.

This is a fundamentally different class of gap than #1865 — it requires an actual identity signal beyond line position, e.g. per-declaration content hashing (hash each sibling's own body text, persisted per-node so it survives a purge+reinsert cycle), as originally suggested as a *possible* direction in #1865's own issue text. It would need a new persisted column on `nodes` (and its Rust migration mirror), populated at node-insertion time in both engines, and is a substantially larger, schema-touching change — out of scope for a targeted line-alignment fix.

## Reproduction (theoretical — not yet verified with a concrete fixture)

1. File `conn.ts` has 4 same-named/same-kind siblings (e.g. `close()` in `openA`..`openD`).
2. Edit: rename `openB`'s `close` to `shutdown` (removes one from the group) AND simultaneously add a new `close()` to a 5th function `openE` inserted at a line position that keeps the total group size at 4.
3. Incremental rebuild: the ordinal-by-rank fast path (count unchanged: 4 -> 4) will match old rank i to new rank i, but if `openE`'s new close() lands at the same rank openB's old close() held, a reverse-dep caller of B could get silently reconnected to E instead of dropped.

## Suggested direction

Per-declaration content hashing: compute a hash of each declaration's own source text (line..endLine) at insertion time, persist it on the node row, and use it as the primary reconnect-matching signal (falling back to the #1865 alignment when hashes are unavailable, e.g. on DBs from before this feature). Needs a schema migration in both `src/db/migrations.ts` and `crates/codegraph-core/src/db/connection.rs`, and hash computation wired into both the WASM/JS and native node-insertion paths.

## How this was found

Identified during design of the #1865 fix — the dominant-shift alignment provably handles any count change correctly (verified via unit tests using this repo's own real fixture line numbers), but a net-zero-count compound edit is an inherent blind spot for any line-position-only heuristic, not a bug in that specific fix.

carlos-alm:
## Summary
- `reconnectReverseDepEdges`/`pick_reconnect_target` previously matched siblings by line position and rank alone during incremental rebuilds. When a same-named/same-kind sibling is both renamed away and a different one added in the same edit, the sibling group's size stays unchanged (e.g. 4 -> 4), so the existing rank-based fallback silently reattaches the edge to whichever new declaration now occupies the old rank.
- Add a per-declaration `content_hash` (SHA-256 of the declaration's own source text), computed centrally once per file right after extraction in both engines, persisted via a new `nodes.content_hash` column (migration v24).
- `pickReconnectTarget`/`pick_reconnect_target` now try an exact hash match first. If the saved target's hash matches none of the current candidates, the edge is confidently dropped instead of falling through to the rank-based guess. Line-based sibling alignment (#1865) remains the fallback when hash data is unavailable or ambiguous (2+ candidates share a hash).
- Both engines verified to produce byte-identical `content_hash` values and identical reconnect behavior for the compound rename+add scenario.

## Test plan
- [x] `npm test` — 263 test files, 4294 tests passing (both wasm/native engine variants of the new integration test)
- [x] `cargo test --manifest-path crates/codegraph-core/Cargo.toml --lib` — 698 tests passing, including new hash-matching unit tests and one full end-to-end reconnection test
- [x] `cargo clippy --manifest-path crates/codegraph-core/Cargo.toml --lib` — no new warnings vs `origin/main`
- [x] `npx biome check src/ tests/` and `npx tsc --noEmit -p .` — clean
- [x] New regression test `tests/integration/issue-2015-reverse-dep-compound-rename-and-add.test.ts` reproduces the exact compound rename+add scenario and fails without the fix (verified via a stale pre-fix native binary reproducing the bug), passes with it

Closes #2015

### DIFF
--- a/crates/codegraph-core/src/db/connection.rs
+++ b/crates/codegraph-core/src/db/connection.rs
@@ -378,6 +378,22 @@ const MIGRATIONS: &[Migration] = &[
       CREATE INDEX IF NOT EXISTS idx_dataflow_call_edge ON dataflow(call_edge_id);
     "#,
     },
+    Migration {
+        // Per-declaration content hash (issue #2015): reverse-dep-edge
+        // reconnection during incremental rebuilds previously matched
+        // siblings by line position alone, which is provably unsafe when a
+        // same-named/same-kind sibling group has one member renamed away
+        // and a different one added in the same edit — the group's size
+        // stays unchanged, so the line-alignment fast path matches by rank
+        // and can silently reconnect a caller to the wrong declaration. A
+        // content hash gives reconnection a true identity signal to try
+        // first, falling back to line alignment only when a hash is
+        // unavailable (e.g. rows from before this migration).
+        version: 24,
+        up: r#"
+      ALTER TABLE nodes ADD COLUMN content_hash TEXT;
+    "#,
+    },
 ];
 
 // ── napi types ──────────────────────────────────────────────────────────
@@ -726,6 +742,18 @@ impl NativeDatabase {
             if !has_column(conn, "nodes", "visibility") {
                 let _ = conn.execute_batch("ALTER TABLE nodes ADD COLUMN visibility TEXT");
             }
+            // #2015: content_hash is referenced unconditionally by every node
+            // insert (see insert_nodes.rs), unlike the other legacy columns in
+            // this block — propagate failure instead of swallowing it, mirroring
+            // the #2001/#2066 dynamic_kind fix on the edges table below.
+            if !has_column(conn, "nodes", "content_hash") {
+                conn.execute_batch("ALTER TABLE nodes ADD COLUMN content_hash TEXT")
+                    .map_err(|e| {
+                        napi::Error::from_reason(format!(
+                            "legacy repair: add nodes.content_hash failed: {e}"
+                        ))
+                    })?;
+            }
             let _ = conn.execute_batch(
                 "UPDATE nodes SET qualified_name = name WHERE qualified_name IS NULL",
             );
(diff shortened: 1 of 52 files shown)


carlos-alm:
Investigated the sole-candidate hash bypass. Gating that path on the hash regresses a much more common scenario than it fixes: `tests/integration/issue-1744-incremental-edge-technique.test.ts` edits a single function's own body in place (a file with exactly one function, whose body text changes) — its `content_hash` necessarily changes even though it's still the same declaration. Applying the hash gate there caused the reverse-dep edge to be dropped instead of reconnected, failing that test.

Unlike the multi-candidate case, a lone candidate has no sibling to corroborate a hash mismatch as evidence of "an unrelated declaration" — a hash difference is indistinguishable from an ordinary body edit when there's only one candidate to compare against. The 1-to-1 rename-and-replace scenario you flagged is real, but it's genuinely unresolvable with content-hash alone at group size 1; fixing it would require tracking whether the exact old hash reappears elsewhere in the file under a new name, which is a materially bigger feature (file-wide rename detection) than this PR's scope.

Reverted the sole-candidate path to trust the candidate unconditionally (matching pre-#2015 behavior) in `0777f5a8`, replaced the two tests that encoded the now-reverted expectation with one documenting the correct behavior, and verified #1744 passes again on both engines.

@greptileai please re-review

Issue #2222, opened 2026-08-02
loadConfig() still resolves from process.cwd() in startMCPServer and outputResult's table display


carlos-alm:
## Summary

Follow-up to #1881 and #2017, both of which fixed the `process.cwd()`-based `loadConfig()` pattern at a named list of call sites by switching to `resolveConfigForDbPath()`/`resolveDbConfig()` (in `src/db/connection.ts`, exported from `src/db/index.ts`) — deriving `loadConfig()`'s rootDir from the resolved `--db` path instead of the invoking process's cwd.

A further grep for `loadConfig(` across `src/` while working #2017 turned up two more instances of the same bug, both outside #2017's named scope:

- `startMCPServer()` — `src/mcp/server.ts:208`: `const config = options.config || loadConfig();`. This function already takes `customDbPath?: string` as its first parameter, so the fix is the same mechanical pattern as #2017: `resolveDbConfig(customDbPath)`.
- `outputResult()` — `src/presentation/result-formatter.ts:158`: `const displayOpts = opts.display ?? (loadConfig() as { display: DisplayOpts }).display;`. Unlike the other call sites, this function has **no** `customDbPath`/`dbPath` parameter at all — it's a shared presentation-layer formatter called from many CLI commands after they've already fetched their data. Fixing it properly requires either threading a resolved config/dbPath through from every caller (a much bigger, more invasive change than the other instances), or resolving `display` config at the call site (where the DB path is already known) and always passing it in via `opts.display` instead of falling back to `loadConfig()` inside the shared formatter.

## Fix direction

- `startMCPServer`: same as #2017 — `options.config || resolveDbConfig(customDbPath)`, importing `resolveDbConfig` from `../db/index.js`.
- `outputResult`: needs its own design decision (thread a dbPath through, or push config resolution to call sites) rather than the other instances' one-line mechanical swap — worth scoping as its own PR.

## Scope

Out of scope for #2017, which was scoped to the call sites explicitly named in that issue. Filed separately per the project's one-concern-per-PR convention.

Issue #1881, opened 2026-07-06
loadConfig() resolves from process.cwd() instead of the resolved --db path in several read-only query functions


carlos-alm:
## Summary

Issue #1605 fixed `openRepo()` / `openReadonlyWithNative()` (via the shared `resolveDbSettings()` helper in `src/db/connection.ts`) to derive `loadConfig()`'s `rootDir` from the resolved `--db` path rather than `process.cwd()`, so cross-repo `--db /other/project/.codegraph/graph.db` invocations read the *target* repo's `.codegraphrc.json` instead of the invoking directory's.

That fix was scoped to `openRepo()`/`openReadonlyWithNative()` only. While wiring `db.busyTimeoutMs` through the remaining ad-hoc read-only query call sites for #1763, I found the same `process.cwd()`-based pattern still present in several other functions that independently call `loadConfig()`:

- `manifestoData()` in `src/features/manifesto.ts` — `loadConfig(process.cwd())`
- `hybridSearchData()` in `src/domain/search/search/hybrid.ts` — bare `loadConfig()`
- `moduleBoundariesData()` in `src/features/structure-query.ts` — bare `loadConfig()`
- `diffImpactData()` in `src/domain/analysis/diff-impact.ts` — bare `loadConfig()`
- `resolveComplexityQueryOptions()` (used by `complexityData()`) in `src/features/complexity-query.ts` — `loadConfig(process.cwd())`
- `auditData()` in `src/features/audit.ts` — bare `loadConfig()` (note: `resolveThresholds()` in the same file *does* correctly derive rootDir from `customDbPath`, so this file is internally inconsistent)

By contrast, `checkData()` in `src/features/check.ts` already derives `repoRoot` from the resolved DB path and calls `loadConfig(repoRoot)` correctly.

## Impact

Running any of the affected commands with `--db /other/project/.codegraph/graph.db` from a different working directory reads config (manifesto rule thresholds, `analysis.impactDepth`/`auditDepth`, `search.topK`/`rrfK`, `structure.cohesionThreshold`, and now `db.busyTimeoutMs`) from the *invoking* directory's `.codegraphrc.json` instead of the target repo's. Confirmed via a test that had to work around this (see PR for #1763) by mocking `process.cwd()` rather than relying on DB-path-derived config resolution.

## Fix direction

[...]

PR #2223, opened 2026-08-02
fix: resolve loadConfig() from --db path, not process.cwd(), in remaining call sites

### LINKED ISSUE #2017: loadConfig() still resolves from process.cwd() in triageData, sequenceData, communitiesData, briefData, statsData, searchData/multiSearchData, and resolveAnalysisOpts
## Summary

Issue #1881 fixed the `process.cwd()`-based `loadConfig()` pattern in `manifestoData`, `hybridSearchData`, `moduleBoundariesData`, `diffImpactData`, `complexityData`, and `auditData` by introducing a shared `resolveConfigForDbPath()` helper in `src/db/connection.ts` (also exported from `src/db/index.ts`) that derives `loadConfig()`'s rootDir from the resolved `--db` path.

While fixing #1881, a full grep for `loadConfig(` across `src/` turned up more functions with the exact same bug — they take a `customDbPath` (or equivalent) parameter but still call bare `loadConfig()` / `loadConfig(process.cwd())`, so `--db /other/repo/.codegraph/graph.db` invoked from a different directory reads the *invoking* directory's `.codegraphrc.json` instead of the target repo's:

- `resolveRiskConfig()` (used by `triageData()`) — `src/features/triage.ts:127`
- `sequenceData()` — `src/features/sequence.ts:339`
- `communitiesData()` — `src/features/communities.ts:196`
- `briefData()` — `src/domain/analysis/brief.ts:113`
- `statsData()` — `src/domain/analysis/module-map.ts:565`
- `searchData()` and `multiSearchData()` — `src/domain/search/search/semantic.ts:89` and `:218`
- `resolveAnalysisOpts()` (shared by `fn-impact.ts`, `exports.ts`, `context.ts` — i.e. `codegraph fn-impact`, `codegraph exports`, `codegraph context`) — `src/domain/analysis/query-helpers.ts:54`

## Fix direction

Same as #1881: replace `opts.config || loadConfig()` (or `loadConfig(process.cwd())`) with `opts.config || resolveConfigForDbPath(customDbPath)`, importing `resolveConfigForDbPath` from `../db/index.js` (adjusting relative path per file). For `resolveAnalysisOpts()`, this requires threading a `customDbPath` parameter through from each of its three callers (`fn-impact.ts`, `exports.ts`, `context.ts`), which already have the DB path available in their own function signatures.

## Scope

Out of scope for #1881, which was scoped to the six call sites explicitly named in that issue. Filed separately per the project's one-concern-per-PR convention.

carlos-alm:
## Summary
- Issue #1881 fixed the `process.cwd()`-based `loadConfig()` pattern at six named call sites via a shared `resolveDbConfig()` helper that derives `loadConfig()`'s rootDir from the resolved `--db` path.
- A follow-up grep for `loadConfig(` across `src/` turned up more instances of the same bug:
  - `resolveRiskConfig()` (used by `triageData()`) — `src/features/triage.ts`
  - `sequenceData()` — `src/features/sequence.ts`
  - `communitiesData()` — `src/features/communities.ts`
  - `statsData()` — `src/domain/analysis/module-map.ts`
  - `searchData()` / `multiSearchData()` — `src/domain/search/search/semantic.ts`
  - `resolveAnalysisOpts()` — `src/domain/analysis/query-helpers.ts`, threaded through its three callers (`fn-impact.ts`, `exports.ts`, `context.ts`)
- Fixed the same way as #1881: `opts.config || loadConfig()` → `opts.config || resolveDbConfig(customDbPath)`, threading `customDbPath` through wherever it wasn't already a local variable.
- Two further instances (`startMCPServer`, `outputResult`) turned up during the audit but are out of scope for this PR — filed separately as #2222.

## Test plan
- [x] New regression test `tests/unit/loadconfig-dbpath-resolution-2017.test.ts` covers every fixed call site by spying on `loadConfig()` and asserting it's invoked with the `--db` path's derived rootDir (not `process.cwd()`), while cwd is spoofed to an unrelated, differently-configured directory
- [x] Verified each assertion actually catches the bug by temporarily reverting each fix and confirming the corresponding test fails (several call sites needed a stronger assertion than "contains repoDir" once I found `openRepo()`/`openReadonlyWithNative()`/`resolveBusyTimeoutMs()` already resolve config correctly as a side effect — a decoy that would pass even with the bug reverted — so those tests also assert no call was made with an `undefined` rootDir)
- [x] `npm test` — 264 test files, 4306 tests passing
- [x] `npx tsc --noEmit -p .` and `npx biome check src/ tests/` — clean

Closes #2017

### DIFF
--- a/src/domain/analysis/context.ts
+++ b/src/domain/analysis/context.ts
@@ -443,7 +443,7 @@ export function contextData(
     const noSource = opts.noSource || false;
     const includeTests = opts.includeTests || false;
 
-    const { noTests, displayOpts } = resolveAnalysisOpts({
+    const { noTests, displayOpts } = resolveAnalysisOpts(customDbPath, {
       ...opts,
       config: opts.config ?? config,
     });
@@ -517,7 +517,7 @@ export function explainData(
     const depth = opts.depth || 0;
     const kind = isFileLikeTarget(target) ? 'file' : 'function';
 
-    const { noTests, displayOpts } = resolveAnalysisOpts({
+    const { noTests, displayOpts } = resolveAnalysisOpts(customDbPath, {
       ...opts,
       config: opts.config ?? config,
     });
--- a/src/domain/analysis/exports.ts
+++ b/src/domain/analysis/exports.ts
@@ -39,7 +39,7 @@ export function exportsData(
   } = {},
 ) {
   return withReadonlyDb(customDbPath, (db, config) => {
-    const { noTests, displayOpts } = resolveAnalysisOpts({
+    const { noTests, displayOpts } = resolveAnalysisOpts(customDbPath, {
       ...opts,
       config: opts.config ?? config,
     });
--- a/src/domain/analysis/fn-impact.ts
+++ b/src/domain/analysis/fn-impact.ts
@@ -277,7 +277,10 @@ export function fnImpactData(
   } = {},
 ) {
   return withRepo(customDbPath, (repo, dbConfig) => {
-    const { noTests, config } = resolveAnalysisOpts({ ...opts, config: opts.config ?? dbConfig });
+    const { noTests, config } = resolveAnalysisOpts(customDbPath, {
+      ...opts,
+      config: opts.config ?? dbConfig,
+    });
     const maxDepth = opts.depth || config.analysis?.fnImpactDepth || 5;
     const hc = new Map();
 
--- a/src/domain/analysis/module-map.ts
+++ b/src/domain/analysis/module-map.ts
@@ -3,9 +3,9 @@ import {
   openReadonlyOrFail,
   openReadonlyWithNative,
   resolveBusyTimeoutMs,
+  resolveDbConfig,
   testFilterSQL,
 } from '../../db/index.js';
-import { loadConfig } from '../../infrastructure/config.js';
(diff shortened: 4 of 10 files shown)


carlos-alm:
Investigated — this is real, but it's a pre-existing defect in the shared \`resolveDbConfig\`/\`deriveRootDirFromDbPath\` helper (\`src/db/connection.ts\`), not something this PR introduces. That file is untouched here (\`git diff origin/main..HEAD -- src/db/connection.ts\` is empty) — this PR only adds *more callers* of the existing helper.

Confirmed empirically against an ORIGINAL #1881 call site (\`manifestoData\`, completely untouched by this PR): pointing \`--db\` at a real file outside the \`<repo>/.codegraph/graph.db\` convention (e.g. \`/repo/custom.db\`) already resolves the wrong rootDir and silently falls back to \`DEFAULTS\` instead of the project's \`.codegraphrc.json\`. So this equally affects all six of #1881's original call sites, not just the ones this PR migrates.

Filed separately as #2224 with the repro and fix direction (`deriveRootDirFromDbPath` needs to detect the actual path shape instead of assuming a fixed two-level strip) — fixing the shared helper itself deserves its own PR and test coverage across every consumer, not a scope-expanding patch buried in this one.

@greptileai please re-review

Issue #1763, opened 2026-07-03
db.busyTimeoutMs: extend config-driven wiring to read-only query paths and Rust connection.rs


carlos-alm:
Follow-up to #1749 / the grind-phase-7 wiring of `DEFAULTS.db.busyTimeoutMs` (src/infrastructure/config.ts) into `openDb()` / `openReadonlyOrFail()` (src/db/connection.ts).

That change wired the config value as the default parameter of both functions, and threads the *actual resolved* `config.db.busyTimeoutMs` value through every call site that already holds a loaded `CodegraphConfig` with no reordering risk: the build pipeline (`pipeline.ts`, `native-orchestrator.ts`, `native-db-lifecycle.ts`), watch mode (`watcher.ts`), and `openRepo()`/`openReadonlyWithNative()` (via the renamed `resolveDbSettings()` helper, which already called `loadConfig()` once for engine resolution).

Two gaps remain, deliberately deferred as out of scope for that change:

1. **~15 ad-hoc read-only query call sites** (`features/manifesto.ts`, `features/owners.ts`, `features/dataflow.ts`, `domain/search/search/hybrid.ts`, `domain/analysis/roles.ts`, `domain/analysis/query-helpers.ts`, `features/structure-query.ts`, `domain/search/search/prepare.ts`, `cli/commands/embed.ts`, `features/cfg.ts`, etc.) call `openReadonlyOrFail()` directly and get the `DEFAULTS.db.busyTimeoutMs` default rather than the user's configured override. Several of these already call `loadConfig()` themselves, but *after* opening the db, so wiring the value through would require reordering config-load-then-db-open in each file — each reorder needs verification that it doesn't change `ConfigError`-throw timing/observable behavior for callers that currently see a "no db found" error before any config validation runs.

2. **Rust native DB layer** (`crates/codegraph-core/src/db/connection.rs:504,522`) independently hardcodes `PRAGMA busy_timeout = 5000` and doesn't read `CodegraphConfig` at all. Per the mirrored-engine-layout convention in CLAUDE.md, this should eventually read the same `db.busyTimeoutMs` value, but that requires the Rust db layer to gain config-loading capability it doesn't currently have — a larger design question than this ticket's scope.

Neither gap causes incorrect behavior today (both fall back to the same hardcoded default that was already in place before the wiring), but full config-driven coverage remains incomplete until these are addressed.

carlos-alm:
[...]

PR #2226, opened 2026-08-02
fix: apply configured busy_timeout to remaining raw new Database() call sites

### LINKED ISSUE #2020: Several call sites open better-sqlite3 directly via new Database(), bypassing busy_timeout entirely
## Summary

Issue #1763 threaded `config.db.busyTimeoutMs` through every call site that goes via `openReadonlyOrFail()` / `openReadonlyWithNative()`. However, several call sites open `better-sqlite3` directly with `new Database(dbPath, { readonly: true })`, bypassing that helper entirely — so they never set `PRAGMA busy_timeout` at all (not even the hardcoded 5000ms default), let alone the user-configured value.

## Call sites found

- `src/features/snapshot.ts:103`
- `src/features/sequence.ts:183`
- `src/features/branch-compare.ts:205` and `:246` (`loadSymbolsFromDb`, `loadCallersFromDb`)
- `src/mcp/tools/find-cycles.ts:13`
- `src/mcp/tools/export-graph.ts:18`
- `src/cli/commands/info.ts:61`
- `src/db/repository/native-repository.ts:241` (`#getFallbackDb`, lazy fallback handle)

None of these call `db.pragma('busy_timeout = ...')` after opening.

## Impact

On a busy/locked DB (e.g. a concurrent `codegraph build` holding the write lock), these call sites get `SQLITE_BUSY` immediately instead of waiting up to the configured (or default) timeout — inconsistent with the rest of the codebase's read-only DB access, and worse than even the pre-#1763 baseline (which at least had the hardcoded 5000ms default via `openReadonlyOrFail`).

## Suggested fix

Route these through `openReadonlyOrFail()` / `resolveBusyTimeoutMs()` (or set the pragma directly with a resolved value) — same pattern already established by #1763.

## Discovered while

Working on #1882 (threading `config.db.busyTimeoutMs` into the Rust `NativeDatabase` factories) — noticed `branch-compare.ts`'s two `new Database()` call sites use this raw pattern while auditing that file's `NativeDatabase.openReadonly()` call for the fix.

carlos-alm:
## Summary
- Issue #1763 threaded `config.db.busyTimeoutMs` through every call site going via `openReadonlyOrFail()`/`openReadonlyWithNative()`. Several call sites bypassed that helper entirely by opening better-sqlite3 directly with `new Database(dbPath, { readonly: true })`, so they never set `PRAGMA busy_timeout` at all — not even the hardcoded 5000ms default, worse than the pre-#1763 baseline.
- On a busy/locked DB (e.g. a concurrent `codegraph build` holding the write lock), these call sites got `SQLITE_BUSY` immediately instead of waiting up to the configured timeout.
- Added `db.pragma(\`busy_timeout = ${resolveBusyTimeoutMs(dbPath)}\`)` right after each raw `new Database()` call, matching the pattern already established by #1763:
  - `src/features/snapshot.ts` (`snapshotSave`)
  - `src/features/sequence.ts` (`annotateDataflow`'s native-fallback branch)
  - `src/features/branch-compare.ts` (`loadSymbolsFromDb`, `loadCallersFromDb` — exported for direct testing, mirroring the same file's existing `openNativeDbForFanMetrics` export)
  - `src/mcp/tools/find-cycles.ts` and `export-graph.ts`
  - `src/cli/commands/info.ts` (`printBuildMetadata`)
  - `src/db/repository/native-repository.ts` (`#getFallbackDb`'s lazy fallback handle)

## Test plan
- [x] New regression test `tests/unit/busy-timeout-raw-database-sites-2020.test.ts` covers all sites except `branch-compare.ts`'s two (now-exported) functions directly
- [x] Verified each assertion actually catches the bug by temporarily reverting each fix and confirming the corresponding test fails — an early version of the `branch-compare.ts` test (going through the full `branchCompareData()` git-worktree flow) was a decoy that passed even with the bug present, since `buildGraph()` and other already-fixed call sites in that flow already call `busy_timeout` correctly; fixed by exporting `loadSymbolsFromDb`/`loadCallersFromDb` for direct testing instead
- [x] `npm test` — 266 test files, 4317 tests passing
- [x] `npx tsc --noEmit -p .` and `npx biome check src/ tests/` — clean (this is a pure TypeScript change, no Rust/native involved)

Closes #2020

### DIFF
--- a/src/cli/commands/info.ts
+++ b/src/cli/commands/info.ts
@@ -46,8 +46,11 @@ function printEngineInfo(
   console.log();
 }
 
-/** Print the "Build metadata" block read from the graph DB, if one exists. Never throws. */
-async function printBuildMetadata(
+/**
+ * Print the "Build metadata" block read from the graph DB, if one exists. Never throws.
+ * Exported for direct testing of its own busy_timeout pragma (issue #2020).
+ */
+export async function printBuildMetadata(
   ctx: CliContext,
   opts: CommandOpts,
   activeName: string,
--- a/src/cli/commands/info.ts
+++ b/src/cli/commands/info.ts
@@ -53,12 +53,13 @@ async function printBuildMetadata(
   activeName: string,
 ): Promise<void> {
   try {
-    const { findDbPath, getBuildMeta } = await import('../../db/index.js');
+    const { findDbPath, getBuildMeta, resolveBusyTimeoutMs } = await import('../../db/index.js');
     const Database = (await import('better-sqlite3')).default;
     const dbPath = findDbPath(opts.db as string | undefined);
     const fs = await import('node:fs');
     if (fs.existsSync(dbPath)) {
       const db = new Database(dbPath, { readonly: true });
+      db.pragma(`busy_timeout = ${resolveBusyTimeoutMs(dbPath)}`);
       const buildEngine = getBuildMeta(db, 'engine');
       const buildVersion = getBuildMeta(db, 'codegraph_version');
       const builtAt = getBuildMeta(db, 'built_at');
--- a/src/db/repository/native-repository.ts
+++ b/src/db/repository/native-repository.ts
@@ -45,6 +45,7 @@ import type {
   TriageNodeRow,
   TriageQueryOpts,
 } from '../../types.js';
+import { resolveBusyTimeoutMs } from '../connection.js';
 import { normalizeFileFilter } from '../query-builder.js';
 import {
   type FnDepsCallerNode,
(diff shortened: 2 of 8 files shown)


carlos-alm:
Good catch — this one was accurate, unlike some other recent Greptile findings I've had to push back on. Added direct tests for both gaps in \`4aeab391\`:

- \`sequence.ts\`'s \`annotateDataflow\` native-fallback branch — exported it and tested with a stub \`Repository\` (not \`instanceof SqliteRepository\`, \`hasDataflowTable() => true\`) to force the branch that opens its own handle.
- \`info.ts\`'s \`printBuildMetadata\` — exported it and called directly with a minimal \`opts.db\`.

Verified both new tests fail when their respective pragma call is reverted, confirming they actually catch the gap you flagged.

@greptileai please re-review

Issue #2182, opened 2026-07-30
Dart: function_signature/method_signature never contains its own body — breaks per-function complexity/dataflow analysis


carlos-alm:
## Summary

While attempting to wire Dart into `COMPLEXITY_RULES`/`HALSTEAD_RULES` (part of #1923's tier-2 language rollout), I found that tree-sitter-dart's grammar puts a function's body in a node **sibling** to its signature, not a **child** of it:

```
program
  function_signature [line 1]      <- name, params, return type
  function_body [line 1-3]         <- SIBLING, not a child of function_signature
    block
      ...actual statements...
```

Confirmed identically for methods (`method_signature` + sibling `function_body` under `class_body`), and for both block bodies (`{ ... }`) and arrow bodies (`=> expr;`).

## Why this breaks analysis

`functionNodes: new Set(['function_signature', 'method_signature'])` is already used by the existing `dataflowDart` config in `src/ast-analysis/rules/b2.ts` (added for P5 Batch B2), and is exactly what I was about to add for complexity/Halstead. But every consumer of `functionNodes` in this codebase assumes the function's *entire relevant content* lives in the matched node's own subtree:

1. **Standalone DFS reference** (`computeFunctionComplexity`/`computeHalsteadMetrics` in `src/features/complexity.ts`): walks only `functionNode`'s own children. Since `function_signature`/`method_signature`'s only children are the parameter list, walking it can never see any `if`/`for`/`while`/etc. in the body — cyclomatic/cognitive complexity for every Dart function would silently compute as the trivial base case (verified directly: a function with an if/else-if/else chain still reports `cyclomatic: 1`, not `3`).
2. **Visitor-based batch engine** (`src/ast-analysis/visitor.ts` + `src/ast-analysis/visitors/complexity-visitor.ts`, the actual path used during `codegraph build`): dispatches `enterFunction`/`exitFunction` at the boundaries of nodes matching `functionNodeTypes`. `exitFunction` fires as soon as the walker finishes `function_signature`'s own subtree — i.e. *before* the walker ever reaches the sibling `function_body` — so the accumulator for that function is finalized (and `activeFuncNode` reset to `null`) before a single statement in the real body is visited. `enterNode` then no-ops for the entire body (`if (fileLevelWalk && !activeFuncNode) return;`).
[...]

Issue #1923, opened 2026-07-06
Complexity metrics (cognitive/cyclomatic/Halstead) missing for 24 of 34 supported languages


carlos-alm:
## Summary

`codegraph complexity --file <f> --health -T --json` silently returns `functions: []` for any file in a language outside a small supported subset — not because those files lack functions (`codegraph where --file` extracts them correctly), but because `COMPLEXITY_RULES`/`HALSTEAD_RULES` (`src/ast-analysis/rules/index.ts`) and the native mirror (`lang_rules`/`halstead_rules` in `crates/codegraph-core/src/ast_analysis/complexity.rs`) simply have no entry for the language. This is the same root-cause shape as #1782 (Lua), just unfixed for the remaining languages.

Confirmed via direct repro (native engine) for a spot-check of 4 languages, all returning `functions: 0` despite the fixtures having real functions:

```
$ codegraph complexity --file tests/benchmarks/resolution/fixtures/elixir/main.ex --health -T --json --engine native
functions: 0
$ codegraph complexity --file tests/benchmarks/resolution/fixtures/zig/main.zig --health -T --json --engine native
functions: 0
$ codegraph complexity --file tests/benchmarks/resolution/fixtures/gleam/main.gleam --health -T --json --engine native
functions: 0
$ codegraph complexity --file tests/benchmarks/resolution/fixtures/clojure/main.clj --health -T --json --engine native
functions: 0
```

## Exact scope (post-#1782, i.e. after Lua is fixed)

`LANGUAGE_REGISTRY` has 36 entries (35 real languages + intentionally-excluded `hcl`, which has no function/branch concept and should stay excluded).

**TS `COMPLEXITY_RULES`/`HALSTEAD_RULES` supported (11):** javascript, typescript, tsx, python, go, rust, java, csharp, ruby, php, lua

**TS missing (24):** c, cpp, kotlin, swift, scala, bash, elixir, dart, zig, haskell, ocaml, ocaml-interface, fsharp, fsharp-signature, gleam, clojure, julia, r, erlang, solidity, objc, cuda, groovy, verilog

**Native Rust `lang_rules`/`halstead_rules` supported (16, post-#1782):** javascript, typescript, tsx, python, go, rust, java, csharp, ruby, php, c, cpp, kotlin, swift, scala, bash, lua

[...]

PR #2233, opened 2026-08-03
feat: add complexity metrics for Objective-C (#1923)

carlos-alm:
## Summary

Progress on #1923 — adds `objc` complexity/Halstead support (tier 2).

tree-sitter-objc extends tree-sitter-c: if/else/for/while/switch/case/logical-operator/conditional-expression node kinds are byte-identical to plain C (confirmed by parsing sample ObjC control flow and inspecting the S-expression), including the same `else_clause` wrapper (Pattern A). Wires `COMPLEXITY_RULES`/`HALSTEAD_RULES` (TS) and `lang_rules()`/`halstead_rules()` (native) for `objc`, extending C's rule shape with:

- `method_definition` (the `-`/`+` method body) added to `functionNodes` alongside `function_definition`
- `catch_clause` (from `@try`/`@catch`/`@finally`) treated as a branch/nesting node, the same treatment C++'s `catch_clause` already gets
- `message_expression`/`selector_expression` treated as compound Halstead operators, mirroring `call_expression`

Also fixes a native/WASM divergence this surfaced: the native ObjC extractor's `handle_method` ran complexity/CFG analysis unconditionally on both `method_declaration` (an `@interface`/`@protocol` signature, which never has a body) and `method_definition`, producing a spurious trivial complexity entry (cognitive 0, cyclomatic 1) for every interface declaration that duplicated the real `method_definition` entry under the same dotted name. Gated both on actual body presence and set `bodyless` accordingly; ported the same explicit `bodyless` signal to the TS/WASM extractor for defense-in-depth (its `hasFuncBody` skip logic otherwise relies on an `endLine > line` heuristic that a multi-line bodyless signature could slip past).

## Verification

- `npm run lint` — clean
- `npm test` — 267/267 test files, 4335 passed, 30 skipped, 2 todo
- `cargo test --lib` (crates/codegraph-core) — 707/707 passed
- Native and WASM engines produce **byte-identical** `codegraph complexity --health --json` output on a hand-built fixture covering if/else-if/else, classic and fast-enumeration for loops, switch/case, `@try`/`@catch`/`@finally`, ternary, and message sends.

## Remaining scope on #1923

Tier 1 (c, cpp, kotlin, swift, scala, bash) and CUDA were already done. Dart is blocked on #2182 (grammar structural issue) — do not attempt until that's fixed. Remaining unattempted: zig, haskell, ocaml, ocaml-interface, fsharp, fsharp-signature, gleam, clojure, julia, r, erlang, solidity, groovy, verilog.

## Test plan

- [x] `npm run lint`
- [x] `npm test`
- [x] Native vs WASM diff on a hand-built ObjC fixture (identical output)

### DIFF
--- a/crates/codegraph-core/src/ast_analysis/complexity.rs
+++ b/crates/codegraph-core/src/ast_analysis/complexity.rs
@@ -382,6 +382,36 @@ pub static CPP_RULES: LangRules = LangRules {
     switch_like_nodes: &["switch_statement"],
 };
 
+// tree-sitter-objc extends tree-sitter-c: if_statement/for_statement/
+// while_statement/do_statement/switch_statement/case_statement/
+// conditional_expression/binary_expression are byte-identical to plain C
+// (confirmed by parsing sample ObjC control flow and inspecting the
+// S-expression), including the same else_clause wrapper (Pattern A). Two
+// additions on top of C_RULES:
+//   - `method_definition` (the `-`/`+` method body) joins `function_definition`
+//     in function_nodes — its compound_statement body is a direct child
+//     (unlike tree-sitter-dart's function_signature/function_body sibling
+//     split, #2182), confirmed by parsing
+//     `@implementation Foo - (void)bar { .. } @end`.
+//   - `catch_clause` (from `@try`/`@catch`/`@finally`, which tree-sitter-objc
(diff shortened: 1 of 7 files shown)


carlos-alm:
Thanks for the review — 5/5 confidence, no actionable defects, nothing further to address here.

Issue #2235, opened 2026-08-03
typeMap collisions across same-named locals in different functions cause native/wasm divergence (residual gap after #1949)


carlos-alm:
## Found while verifying #1949's fix

After implementing #1949's native CHA typed-dispatch fallback and re-running the repo's own `src/` self-build repro, native's edge count for `calls`/`receiver` narrowed substantially against wasm but a residual 16-caller gap remains, all still targeting `NativeDbProxy.prepare`/`.transaction` — but this residual gap is a **different root cause**, not the CHA-dispatch issue #1949 fixed.

## Repro

```bash
node dist/cli.js build src --engine native --no-incremental --db /tmp/native.db
node dist/cli.js build src --engine wasm    --no-incremental --db /tmp/wasm.db
```

Compare edges from `openReadonlyOrFail` (`src/db/connection.ts`):

- **native**: `receiver` edge for `db` → `LockedDatabase` (confidence 0.9); no `calls` edge to `prepare`/`pragma` at all.
- **wasm**: `receiver` edge for `db` → `BetterSqlite3Database` (confidence 1.0); `calls` edges to `BetterSqlite3Database.prepare`/`.pragma` (0.6, proximity) AND to `NativeDbProxy.prepare`/`.pragma` (0.8, CHA-expanded).

`openReadonlyOrFail` casts its local `db` to `BetterSqlite3Database` directly (`as unknown as BetterSqlite3Database`, line 389). `LockedDatabase` (`type LockedDatabase = BetterSqlite3Database & { __lockPath?: string }`) is a **different local variable, in a different function in the same file**, cast via `as unknown as LockedDatabase` at line 171.

## Root cause (diagnosed, not yet fixed)

Both engines' typeMap is a **flat, per-file, name-only** structure (`HashMap<&str, (&str, f64)>` in `build_type_map` / `Map<string, TypeMapEntry>` via `setTypeMapEntry`) — there is no per-function scoping. When two *different* functions in the same file each have their own local/parameter named `db` typed differently, both entries key-collide under the plain name `"db"`.

[...]

Issue #2139, opened 2026-07-20
bug: native engine misses call edges when receiver's static type is an interface with multiple implementers (NativeDbProxy, TreeSitterNode, Repository)


carlos-alm:
## Found during dogfooding v3.16.0

**Severity:** Medium
**Command:** `codegraph build --engine native` vs `--engine wasm`

## Reproduction

```bash
codegraph build . --engine native --no-incremental
sqlite3 .codegraph/graph.db "SELECT n1.file,n1.name,n1.line,e.kind,n2.file,n2.name,n2.line \
  FROM edges e JOIN nodes n1 ON e.source_id=n1.id JOIN nodes n2 ON e.target_id=n2.id \
  ORDER BY 1,2,3,4,5,6,7;" | sort > /tmp/edges_native.txt

codegraph build . --engine wasm --no-incremental
sqlite3 .codegraph/graph.db "... same query ..." | sort > /tmp/edges_wasm.txt

comm -23 /tmp/edges_native.txt /tmp/edges_wasm.txt   # empty — native has no edges wasm lacks
comm -13 /tmp/edges_native.txt /tmp/edges_wasm.txt   # 84 edges wasm has that native lacks
```

Built against codegraph's own source (993 files): native = 43195 edges, wasm = 43279 edges. Every one of the 84-edge difference is wasm-only (native is a strict subset — no edges are unique to native), so this is a native under-resolution gap, not a wasm over-resolution one, for the dominant pattern below.

## Dominant pattern (65 of 84 edges): interface-typed receiver with multiple implementers

```
src/domain/graph/builder/incremental.ts|backfillIncrementalEdgeTechniques|1187|calls|src/domain/graph/builder/native-db-proxy.ts|NativeDbProxy.prepare|28
src/domain/graph/builder/helpers.ts|runChaPostPass|621|calls|src/domain/graph/builder/native-db-proxy.ts|NativeDbProxy.transaction|82
src/ast-analysis/visitors/cfg-shared.ts|getBodyStatements|136|calls|src/types.ts|TreeSitterNode.namedChild|866
src/shared/hierarchy.ts|resolveViaRepo|29|calls|src/db/repository/in-memory-repository.ts|InMemoryRepository.getClassHierarchy|507
src/shared/hierarchy.ts|resolveViaRepo|29|calls|src/db/repository/native-repository.ts|NativeRepository.getClassHierarchy|441
[...]

PR #2236, opened 2026-08-03
fix(native): add CHA typed-dispatch fallback to resolve_call_targets_core

### LINKED ISSUE #1949: native vs wasm role classification disagrees on core-vs-utility for 117 symbols on a full build of this repo's own codebase
## Summary

While verifying dual-engine parity for an unrelated fix (the incremental "1-file rebuild" benchmark regression), I ran a full (non-incremental) build of this repo's own codebase with each engine and diffed the resulting `codegraph roles -T --json` output on symbols common to both engines.

```
native: 9772 symbols, wasm: 9776 symbols, 9772 in common
117 of the 9772 common symbols have a different role between engines
```

Every mismatch I sampled (30 shown, pattern held for all I checked) is the same shape: **native classifies the symbol `core`, wasm classifies it `utility`** — e.g. `src/domain/analysis/module-map.ts::findHotspots`, `src/db/migrations.ts::applyMigrations`, `src/domain/analysis/fn-impact.ts::expandImplementors`, etc. Never core→leaf, dead→anything, or the reverse direction (utility→core) — always the same asymmetric direction.

`core` vs `utility` is decided by comparing a node's fan-in/fan-out against the **global median** fan-in/fan-out (see `classify_node` in `crates/codegraph-core/src/graph/classifiers/roles.rs` / `graph/classifiers/roles.ts`). The two engines' full builds produced slightly different totals on identical source (native: 19285 nodes / 40073 edges; wasm: 19289 nodes / 40634 edges — a ~561 edge / 4 node difference), which is consistent with (but not proven to be caused by) a difference in the computed median shifting which side of the fan-in/fan-out threshold some borderline nodes land on.

## Why I'm not fixing this now

This is unrelated to the fix I was making (which only touches the *incremental* role-classification and directory-structure-metrics code paths — `do_classify_incremental`/`classifyNodeRolesIncremental` and `refresh_affected_directory_metrics`/`refreshAffectedDirectoryMetrics`). This divergence shows up on a **full, non-incremental build**, which always goes through `do_classify_full`/`classifyNodeRolesFull` — code neither my diff nor its TS/Rust mirrors touch. I verified this is pre-existing (present on `main` before my change) by inspection: my diff has zero non-comment references to `do_classify_full`/`classifyNodeRolesFull`/`build_full_structure`/`buildDirectoryStructure`.

## Suggested next step

1. Confirm the node/edge count difference's root cause first (likely a pre-existing, narrower extraction/resolution parity gap in one or more languages/constructs — this repo's own codebase exercises 34 languages) — `codegraph diff-impact`/a targeted native-vs-wasm node-count diff per file would narrow it down.
2. Once the node/edge counts agree, re-check whether the 117 core/utility mismatches disappear (they may be a downstream symptom rather than an independent bug in the median/threshold logic itself).
3. If a genuine median-computation or threshold-comparison divergence remains after (2), that's the real bug to fix in `classify_node`/`compute_global_medians` (Rust) and its TS mirror.

Filed as a separate issue per this repo's scope-discipline convention rather than expanding the incremental-rebuild regression fix to cover it.

carlos-alm:
## Summary

Closes #1949.

Native's call resolver seeded typed-receiver dispatch (`db.prepare()` where `db: BetterSqlite3Database`) to the interface's own qualified method node **only when** `computeConfidence(callerFile, targetFile) >= 0.5`. For callers many directories away from the interface declaration (e.g. `src/types.ts`), that proximity check fails in both engines and no seed edge is ever created. WASM never has this gap because its CHA post-pass (`resolveChaTargets` in `cha.ts`) resolves typed-receiver dispatch to concrete implementors **unconditionally**, independent of file proximity — "file proximity is not meaningful for virtual dispatch confidence." Native's `runPostNativeCha` only *expands* an existing seed edge; it has no equivalent unconditional fallback.

This adds a BFS+RTA fallback tier to `resolve_call_targets_core` (`crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs`), tried only when the existing tiers found nothing and the receiver has a typeMap-resolved type:

- `build_cha_implementors_map` / `collect_cha_instantiated_types` build a CHA context once per build from `FileEdgeInput.classes` (`extends`/`implements`) and high-confidence (`>= 0.9`) typeMap entries — mirrors `cha.ts`'s `buildChaContext` (the `newExpressions`-based RTA path has no native equivalent, so only the typeMap fallback applies).
- `resolve_cha_dispatch` BFS-walks the implementors map, RTA-filtering by instantiated types — mirrors `resolveChaTargets` exactly, including "always traverse non-instantiated intermediates."
- A new `CHA_TYPED_DISPATCH_CONFIDENCE` constant (0.8) mirrors the identically-named TS constant in `helpers.ts`. Since `resolve_call_targets_core` has ~15 early-return tiers, the new confidence is threaded through as a `confidence_override: &mut Option<f64>` out-param (set only by this one new tier) rather than widening every existing return to a tuple — keeps the blast radius small in a function shared by all 34 supported languages.

## Verification

- **Unit tests**: 4 new tests in `build_edges.rs`'s `call_edge_tests` module — resolves the distant-interface pattern, respects the RTA filter, does not override an already-successful proximity hit, and BFS reaches through a non-instantiated abstract intermediate class. `cargo test --lib`: 711/711 pass.
- **Parity**: `node scripts/parity-compare.mjs --json` and `--hybrid --json` across all 34 language fixtures + pts-javascript + jelly-micro: only pre-existing divergence is `jelly-micro`'s 3-edge cross-fixture `super()` dispatch bleed — confirmed identical on unmodified `origin/main` (unrelated to this change; tracked in #2062). Zero new regressions, zero precision/recall regressions for any language.
- **Repro re-check** (this repo's own `src/`, 340 files, native vs wasm):
  - Isolated before/after (same commit, only this fix applied/reverted): native/wasm `calls`+`receiver` edge count gap unchanged in scope-irrelevant kinds; `NativeDbProxy.{prepare,transaction,exec}` edges gained exactly 51 new `ts-native`-technique, 0.8-confidence direct edges (this fallback's signature) — edge count 30892 → 30943.
  - Role mismatches (`codegraph roles -T --json`, native vs wasm): 55 → 49 (6 fixed), precisely the caller files the issue's own investigation named (`incremental.ts`, `stages/build-edges.ts`, `stages/detect-changes.ts`, `stages/native-orchestrator.ts`).
- `npm run lint`: clean. `npm test`: 267 files / 4335 tests pass. `cargo clippy --lib`: only pre-existing `too_many_arguments` lint category ticks up by one on the two functions this out-param necessarily touches (already pervasive crate-wide, not CI-gated).
- `codegraph diff-impact --staged -T`: no findings — `crates/**` is explicitly excluded from this repo's own dogfooding `.codegraphrc.json` scope, so the Rust crate isn't self-analyzed (expected, not a bug).

## Follow-up (filed, not fixed here)

The remaining 49 role mismatches are unrelated pre-existing gaps, diagnosed while verifying this fix:
- #2062 — `resolveThisDispatch` cross-file same-named-class collision (the jelly-micro divergence above).
- #2139 — dogfooding report largely superseded by this fix's dominant pattern; its secondary/tertiary observations remain open.
- #2235 (new) — native/wasm typeMap seeding disagree on which same-named local a file's several unrelated functions resolve to (e.g. `LockedDatabase` vs `BetterSqlite3Database` for two different `db` locals in `db/connection.ts`) — an extraction-layer gap, not a resolution-layer one, so out of scope for this PR.

## Test plan

- [x] `cargo test --lib` (711/711)
- [x] `cargo clippy --lib` on touched file
- [x] `node scripts/parity-compare.mjs --json` and `--hybrid --json` (34 fixtures)
- [x] `npx vitest run tests/benchmarks/resolution/resolution-benchmark.test.ts` (206/206, WASM-only benchmark, unaffected as expected)
- [x] `npm run lint`, `npm test`
- [x] Isolated before/after repro on this repo's own `src/`

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
@@ -1,4 +1,4 @@
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
 
 use napi_derive::napi;
 
@@ -21,6 +21,15 @@ const IMPLEMENTS_TARGET_KINDS: &[&str] = &["interface", "trait", "class"];
 /// Confidence penalty per alias hop — mirrors `PROPAGATION_HOP_PENALTY` in
 /// `src/extractors/javascript.ts`.
 pub(crate) const PROPAGATION_HOP_PENALTY: f64 = 0.1;
+
+/// Fixed confidence for typed-receiver (interface/CHA) dispatch edges produced
+/// by the native CHA fallback in `resolve_call_targets_core` (#1949). File
+/// proximity is not meaningful for virtual dispatch — mirrors
+/// `CHA_TYPED_DISPATCH_CONFIDENCE` in `src/domain/graph/builder/helpers.ts`,
+/// which all three WASM engine paths (inline, WASM post-pass) and the native
+/// post-pass CHA expansion (`runPostNativeCha` in
+/// `src/domain/graph/builder/stages/native-orchestrator.ts`) already agree on.
+pub(crate) const CHA_TYPED_DISPATCH_CONFIDENCE: f64 = 0.8;
 
 #[napi(object)]
 pub struct NodeInfo {
@@ -156,6 +165,22 @@ struct EdgeContext<'a> {
     /// (`x.name(...)`) across every file in this build pass — see
     /// `collect_invoked_property_names` for the #1895 liveness rationale.
     invoked_property_names: HashSet<&'a str>,
+    /// CHA + RTA typed-dispatch context (#1949): interface/class name →
+    /// concrete classes that implement or extend it, built once per build
+    /// pass from every file's `classes` (`extends`/`implements`). Used by
+    /// `resolve_call_targets_core`'s CHA fallback tier only — mirrors the
+    /// `implementors` half of `ChaContext`/`buildChaContext` in
+    /// `src/domain/graph/builder/cha.ts` (the `parents` half, used for
+    /// `this`/`self`/`super` dispatch, has no native equivalent here — that
+    /// dispatch is handled separately by `runPostNativeThisDispatch` in
+    /// `native-orchestrator.ts`).
+    cha_implementors: HashMap<&'a str, Vec<&'a str>>,
+    /// RTA: class names that appear as a high-confidence (>= 0.9) typeMap
+    /// target anywhere in this build pass — mirrors the typeMap fallback
+    /// branch of `collectInstantiatedTypes` in `cha.ts` (the native
+    /// `FileEdgeInput` has no dedicated `newExpressions` list, so only that
+    /// fallback branch applies here).
+    cha_instantiated_types: HashSet<&'a str>,
 }
 
 impl<'a> EdgeContext<'a> {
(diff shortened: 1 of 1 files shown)


carlos-alm:
Both flagged concerns are valid, but verified against the WASM source this PR mirrors (`src/domain/graph/builder/cha.ts`) — they're **pre-existing characteristics already present identically in WASM**, not new divergences introduced here:

**Issue 1 (global simple-name implementor map):** `buildChaContext` in `cha.ts` builds the exact same flat, bare-name-keyed `implementors` map across every file in the build, with no import/module scoping — the Rust `build_cha_implementors_map` is a faithful mirror of that same structure. Same class of bug as #2062 (`resolveThisDispatch` cross-file same-named-class collision).

**Issue 2 (inherited methods not resolved):** `resolveChaTargets` in `cha.ts` does the exact same direct qualified lookup (`${cls}.${methodName}`) with no walk up the `extends` chain for a method a concrete class inherits without overriding — the Rust `resolve_cha_dispatch` mirrors that same limitation.

Since this PR's job is specifically to close the proximity-gate parity gap (native producing zero edges where WASM already produces some via unconditional CHA dispatch), fixing either of these unilaterally in the new Rust code only would make native *more* correct than WASM for this pattern — a new divergence in the other direction, which conflicts with this repo's dual-engine parity requirement. Filed both as a dual-engine follow-up: #2237.

@greptileai

carlos-alm:
@greptileai

PR #2264, opened 2026-08-04
fix(lua): track complexity for anonymous function-expression assignments

### LINKED ISSUE #2036: follow-up: Lua anonymous function expressions (local f = function() end) not tracked in complexity metrics
Deferred from PR #1924 review.

**Original reviewer comment:** https://github.com/optave/ops-codegraph-tool/pull/1924#discussion_r3532056501

**Context:**

`function_nodes` for Lua in both engines only contains `"function_declaration"` (`crates/codegraph-core/src/ast_analysis/complexity.rs`, mirrored in `src/ast-analysis/rules/b3.ts`'s TS equivalent). This covers `function f() end`, `local function f() end`, and dotted forms like `function M.foo() end`.

It does **not** cover the common Lua module-table idiom `local M = {}; M.foo = function(...) end` — here the RHS is a `function_definition` node (an anonymous function expression), assigned via a plain `assignment_statement`/`variable_declaration`, not matched by `function_nodes` or by `find_first_function`'s existing `variable_declarator` special-case (which is JS/TS-shaped, not Lua-shaped). Confirmed both engines agree (both are missing this), so it's not an engine-parity bug — it's a genuine coverage gap in both.

A fair portion of real Lua codebases using the `M.foo = function() ... end` idiom will get zero complexity data for those functions.

**Why deferred:** PR #1924 (which added Lua's complexity/Halstead config in the first place) already explicitly scoped out two closely-related gaps as separate follow-ups (#1922, #1923) to keep that PR narrowly focused on "Lua has zero complexity config at all." This is a third, comparable-scope gap discovered by the same review, in the same spirit — it needs either extending `function_nodes` with attribution logic for anonymous function expressions, or Lua-specific walk-up handling mirroring JS's `variable_declarator` case, in both engines.

carlos-alm:
## Summary

- Lua's `M.foo = function(...) end` module-table idiom and `local f = function() end` assign an anonymous `function_definition` node via a plain `assignment_statement`. Neither engine's extractor turned this into a `Definition`, and neither engine's complexity function-node list recognized `function_definition` as a function scope — so these functions silently got zero complexity/Halstead data.
- Added Definition-creation for identifier/dotted anonymous-function assignments in both extractors (`crates/codegraph-core/src/extractors/lua.rs`, `src/extractors/lua.ts`), mirroring the existing named `function_declaration` handling (plain identifier → kind `function`; dotted `dot_index_expression` → kind `method`).
- Added `function_definition` alongside `function_declaration` to `LUA_RULES.function_nodes` (native) and `complexityLua.functionNodes` (WASM), so nested anonymous functions (e.g. a callback literal passed as an argument) get correct scope/nesting attribution too — mirroring how JS already lists `arrow_function`/`function_expression`.

## Verification

Built the issue's own repro (`local M = {}; M.foo = function(x) if x then return 1 else return 2 end end` + `local f = function(x) return x + 1 end`) through the full `buildGraph` pipeline with both engines. Both now produce identical, real complexity:

- `M.foo` (method): cyclomatic=2, cognitive=2, maxNesting=1
- `f` (function): cyclomatic=1, cognitive=0

Closes #2036

## Test plan

- [x] `cargo test -p codegraph-core --lib` — 748 passed, including 3 new tests covering the module-table idiom, local anonymous assignment, and nested-callback scoping
- [x] `npx vitest run tests/parsers/lua.test.ts` — 26 passed, including 3 new extractor tests
- [x] `npx vitest run tests/integration/issue-2036-lua-anonymous-function-complexity.test.ts` — 4 passed (WASM + native dual-engine parity)
- [x] `npm test` — 273 files, 4422 tests passed
- [x] `npm run lint` — clean
- [x] `codegraph diff-impact --staged -T` — reviewed, scoped to the two new/changed functions

### DIFF
--- a/crates/codegraph-core/src/ast_analysis/complexity.rs
+++ b/crates/codegraph-core/src/ast_analysis/complexity.rs
@@ -507,7 +507,15 @@ pub static LUA_RULES: LangRules = LangRules {
     logical_node_types: &["binary_expression"],
     optional_chain_type: None,
     nesting_nodes: &["if_statement", "for_statement", "while_statement", "repeat_statement"],
-    function_nodes: &["function_declaration"],
+    // "function_declaration" covers named forms (`function f() end`,
+    // `local function f() end`, `function M.foo() end`). "function_definition"
+    // is the anonymous function *expression* node — the RHS of the common
+    // module-table idiom `local M = {}; M.foo = function(...) end` (issue
+    // #2036) as well as `local f = function() end` and any function literal
+    // passed as a callback argument. Both node types have the same
+    // `_function_body` shape (parameters/body fields), so every rule below
+    // (branch/nesting/Halstead scope detection) applies identically to either.
+    function_nodes: &["function_declaration", "function_definition"],
     if_node_type: Some("if_statement"),
     else_node_type: Some("else_statement"),
     elif_node_type: Some("elseif_statement"),
--- a/crates/codegraph-core/src/extractors/lua.rs
+++ b/crates/codegraph-core/src/extractors/lua.rs
@@ -110,6 +110,9 @@ fn extract_lua_params(func_node: &Node, source: &[u8]) -> Vec<Definition> {
 /// Detect `<builtin> = <identifier>` assignments — a locally declared
 /// function bound to a Lua global/builtin identifier (e.g.
 /// `require = traced_require`), the monkey-patch pattern from issue #1776.
+/// Also detects anonymous `function(...) end` values assigned via a plain
+/// `identifier`/dotted target (issue #2036) — see
+/// `handle_lua_function_expr_assignment` below.
 /// Mirrors `handleLuaAssignmentStatement` in `src/extractors/lua.ts` — see
 /// that function's doc comment for the full rationale.
 ///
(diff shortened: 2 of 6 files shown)


carlos-alm:
Confirmed and investigated — this is a real gap, but it's pre-existing shared infrastructure, not something this PR introduces. `matchResultToDef` in `src/ast-analysis/apply-results.ts` disambiguates same-line candidates via `funcNode.childForFieldName('name')`, which is always null for *any* anonymous function node (JS arrow/function expressions included), falling back to `candidates[0]`.

I reproduced the identical failure mode with plain JS, no Lua involved:

```js
const a = (x) => {
  if (x) { return 1; }
  return 0;
}, b = (x) => {
  return 2;
};
```

`b`'s Definition gets `a`'s complexity (cyclomatic 2/cognitive 1 instead of the correct 1/0) today, on `main`, unrelated to this PR — `handleVarFnAssignment` sets `Definition.line` from the *enclosing statement*, not each arrow function's own line, so `b`'s line (1) collides with `a`'s real result.

Filed #2265 with the repro and a suggested column-aware fix, scoped as its own PR since it touches the shared matching path used by every language (not just Lua) and deserves focused review rather than scope-creeping this fix. This PR's own scope — the issue's exact repro (`M.foo` and `f` on separate lines) — is unaffected and verified with identical dual-engine numbers.

carlos-alm:
@greptileai

Issue #1920, opened 2026-07-06
WASM dynamic-import destructure extraction misses rest/default-value bindings that native engine handles


carlos-alm:
## Context

Discovered while implementing the fix for #1781 (dynamic `import()` + destructure consumer crediting).

## Description

`extractDynamicImportNames` in `src/extractors/javascript.ts` (WASM/TS engine) only recognizes `shorthand_property_identifier_pattern` and `pair_pattern` children when collecting names from an object-pattern destructure of a dynamic `import()` result. It does not handle:

- Rest elements: `const { a, ...rest } = await import('./foo.js')` — only extracts `["a"]`, silently drops `rest`.
- Default values: `const { a = 1 } = await import('./foo.js')` — extracts `[]` (nothing at all).

The native Rust engine's equivalent (`collect_object_pattern_names` in `crates/codegraph-core/src/extractors/javascript.rs`) already handles both cases correctly via dedicated `rest_pattern`/`rest_element` and `object_assignment_pattern` branches (see its existing test `finds_dynamic_import_with_aliased_default_destructuring`), so this is an engine-parity gap: native extracts the binding, WASM does not.

Practical impact: a destructured rest/default binding from a dynamic import will not get symbol-level consumer credit under the WASM engine (dead-export false positive / `codegraph exports` undercounting), while the native engine gets it right.

## Repro

\`\`\`js
// WASM path
const symbols = extractSymbols(parseTree, 'test.js');
// input: const { a, ...rest } = await import('./foo.js');
// symbols.imports[0].names === ['a']   (rest missing)

// input: const { a = 1 } = await import('./foo.js');
// symbols.imports[0].names === []      (a missing entirely)
\`\`\`

Compare to Rust's \`collect_object_pattern_names\`, which handles \`rest_pattern\`/\`rest_element\` (via \`extract_rest_identifier\`) and \`object_assignment_pattern\` (extracts the \`left\` field).

## Suggested fix

Mirror Rust's \`collect_object_pattern_names\` handling in TS's \`extractDynamicImportNames\` object_pattern branch: add cases for \`rest_pattern\` (extract the wrapped identifier) and \`object_assignment_pattern\` (extract the \`left\`/key identifier, matching how `pair_pattern` already extracts its `key`).

## Source

- **Discovered during:** implementation of #1781
- **Severity:** bug (engine parity + minor undercounting)
- **Category:** codegraph

Issue #1901, opened 2026-07-06
Engine parity: native JS extractor emits no definition for const array-pattern destructuring


carlos-alm:
## Summary

For a top-level `const [a, b] = someCall();` declaration, the WASM/TS extractor and the native Rust extractor disagree:

- **WASM** (`src/extractors/javascript.ts`, both the walk path in `handleVariableDeclarator` and the query path in `extractDestructuredDeclarators`): emits a single `Definition` with `kind: 'constant'`, whose `name` is the literal source text of the whole pattern (e.g. `"[a, b]"` as one node — not per-element `a`/`b`).
- **Native** (`crates/codegraph-core/src/extractors/javascript.rs`, `handle_var_decl`): emits **nothing** at all for `array_pattern` name nodes in a `const` declaration — there is no `array_pattern` branch in `handle_var_decl`, only in `extract_dynamic_import_names`/`collect_array_pattern_names` (a completely separate code path used only for `const { a, b } = await import(...)` name extraction, which returns bare name strings, not `Definition` structs).

## Evidence

Fixture (`sample.js`):
```js
function getArr() { return [1, 2]; }
const [a, b] = getArr();
```

WASM (`--engine wasm`) `where --file` output includes:
```json
{ "name": "[a, b]", "kind": "constant", "line": 3 }
```

Native (`--engine native`) `where --file` output for the same file: no `[a, b]` node at all — 8 nodes total vs WASM's 9.

## Where this was found

Discovered while investigating issue #1773 (destructured object-pattern bindings misclassified as `kind: "function"`). The object-pattern bug is fixed in that issue's PR; this array-pattern **parity gap** (not a kind-misclassification — a totally missing definition on the native side) is a separate, pre-existing bug and is out of scope there.

Note: the WASM code's own comment claims the single-constant-node behavior is "matching native engine behaviour" (see `src/extractors/javascript.ts`, added in commit `3b5ab14f9`) — that claim is incorrect; native does not currently replicate this at all.

## Suggested fix

Per CLAUDE.md's dual-engine parity mandate, decide the correct shared behavior (likely: extract each bound identifier in the array pattern as its own `constant`-kind `Definition`, superseding the current WASM "one node named by raw pattern text" approach, which itself is not very useful for querying/resolution since `[a, b]` is not a real identifier) and make both engines match it exactly.

carlos-alm:
[...]

PR #2277, opened 2026-08-04
fix(extractors): extract rest/default bindings from destructured const decls

### LINKED ISSUE #2051: JS/TS destructured-binding definition extraction misses rest/default bindings in both engines
## Context

Discovered while implementing the fix for #1920 (dynamic `import()` destructure extraction missing rest/default-value bindings).

## Description

`extractDestructuredBindings` in `src/extractors/javascript.ts` (WASM/TS engine) and its mirror `extract_destructured_bindings` in `crates/codegraph-core/src/extractors/javascript.rs` (native engine) create `constant`-kind `Definition` nodes for each bound name in a plain object-destructuring declaration (e.g. `const { a, b } = someValue;`). Both only recognize `shorthand_property_identifier_pattern` and `pair_pattern` children — neither handles:

- Rest elements: `const { a, ...rest } = someValue;` — `rest` never gets a `Definition`, so `rest()` (if later called) can never resolve to this binding.
- Default values: `const { a = 1 } = someValue;` — `a` never gets a `Definition` at all.

This is the same class of bug just fixed in #1920 for `extractDynamicImportNames`/`extract_dynamic_import_names` (which populates `Import.names` for dynamic `import()` destructures), but affects a different pair of functions that populate the generic `definitions` array for any object-destructuring declaration, not just dynamic imports. Verified present in both engines — this is a genuine parity-preserving gap (both wrong the same way), not just a WASM-only gap.

## Repro

\`\`\`js
const symbols = extractSymbols(parseTree, 'test.js');
// input: const { a, ...rest } = someValue;
// symbols.definitions contains `a` but not `rest`

// input: const { a = 1 } = someValue;
// symbols.definitions does not contain `a` at all
\`\`\`

## Suggested fix

Mirror the `rest_pattern`/`rest_element` and `object_assignment_pattern` handling added to `extractDynamicImportNames`/`extract_dynamic_import_names` in #1920: add matching branches to `extractDestructuredBindings` (TS) and `extract_destructured_bindings` (Rust) so a rest binding gets its own `constant` Definition and a shorthand default's left-hand identifier is recorded.

## Source

- **Discovered during:** implementation of #1920
- **Severity:** bug (symbol-extraction undercounting, dead-code false positives for rest/default destructured bindings)
- **Category:** codegraph

carlos-alm:
## Summary
- `extractDestructuredBindings` (TS/WASM) and its mirror `extract_destructured_bindings` (native) only recognized `shorthand_property_identifier_pattern` and `pair_pattern` children when creating `constant`-kind `Definition`s for a plain object-destructuring `const` declaration, so:
  - `const { a, ...rest } = someValue;` never got a `Definition` for `rest`.
  - `const { a = 1 } = someValue;` never got a `Definition` for `a` at all.
- Verified present identically in both engines — a genuine parity-preserving gap, not a WASM-only bug.
- Fixed by adding `object_assignment_pattern` (default value) and `rest_pattern`/`rest_element` branches to both engines, mirroring the equivalent fix already landed for `extractDynamicImportNames`/`extract_dynamic_import_names` in #1920 (PR #2052). The TS side reuses the existing shared `extractRestPatternIdentifier` helper; the Rust side reuses the existing `extract_rest_identifier` helper.
- Both the walk-based and query-based paths call the same shared `extractDestructuredBindings`/`extract_destructured_bindings` function, so both are fixed by this one change.
- Checked the array-pattern counterpart (`extractArrayPatternBindings`/`extract_array_pattern_bindings`) for the same class of bug per the issue's broader "destructured-binding" title — it already correctly handles both `assignment_pattern` (default) and `rest_pattern`/`rest_element` in both engines (added in #1901), so no change was needed there.

## Repro (issue's own example)
```js
const { a, ...rest } = someValue;
// before: definitions contains `a` but not `rest`
// after:  definitions contains both `a` and `rest`, kind 'constant'

const { a = 1 } = someValue;
// before: definitions does not contain `a` at all
// after:  definitions contains `a`, kind 'constant'
```
Confirmed via a real `buildGraph()` run against a small fixture on both `engine: 'wasm'` and `engine: 'native'` (native addon rebuilt + codesigned locally) — both produce identical `constant` definitions: `a`, `d`, `rest`, `someValue`.

## Test plan
- `npx vitest run tests/parsers/javascript.test.ts` — 291 passed (3 new, covering rest/default/mixed bindings in object-pattern destructured const declarations)
- `cargo test --lib javascript` (crates/codegraph-core) — 216 passed (3 new, mirroring the TS cases)
- Rebuilt and codesigned the native addon locally (`npx napi build --platform --release` + `codesign --sign - --force`) so the fix was exercised by the real native engine
- `npm test` — full suite: 273 files, 4427 passed, 30 skipped, 2 todo, 0 failed
- `npm run lint` — clean
- `node scripts/parity-compare.mjs --langs javascript,typescript,tsx,pts-javascript,dynamic-javascript,dynamic-typescript --hybrid` — PARITY OK, all engines identical across all 6 JS/TS-related fixtures
- `npx vitest run tests/benchmarks/resolution/resolution-benchmark.test.ts` — 206 passed, javascript/typescript/pts-javascript all at 100% precision/recall
- `codegraph diff-impact --staged -T` — confirms the change is scoped to exactly `extractDestructuredBindings`, 6 transitive callers, all within the JS/TS extractor

Closes #2051

### DIFF
--- a/crates/codegraph-core/src/extractors/javascript.rs
+++ b/crates/codegraph-core/src/extractors/javascript.rs
@@ -3525,6 +3525,29 @@ fn extract_destructured_bindings(
                             content_hash: None,
                             accessor_kind: None,
                         });
+                    } else if value.kind() == "assignment_pattern" {
+                        // { original: renamed = defaultValue } — the local
+                        // binding is the assignment_pattern's left identifier
+                        // (Greptile follow-up to #2051, mirrors the identical
+                        // branch already in collect_object_pattern_names
+                        // since #1824).
+                        if let Some(left) = value.child_by_field_name("left") {
+                            if left.kind() == "identifier" {
+                                definitions.push(Definition {
+                                    name: node_text(&left, source).to_string(),
+                                    kind: "constant".to_string(),
+                                    line,
+                                    end_line: Some(end_line),
+                                    decorators: None,
+                                    complexity: None,
+                                    cfg: None,
+                                    children: None,
+                                    bodyless: None,
+                                    content_hash: None,
+                                    accessor_kind: None,
+                                });
+                            }
+                        }
                     }
                 }
             }
@@ -6703,6 +6726,23 @@ mod tests {
             assert_eq!(def.kind, "constant");
         }
         assert!(!s.definitions.iter().any(|d| d.name == "b"), "should not use the original key");
+    }
+
+    #[test]
+    fn extracts_constant_definition_for_renamed_binding_with_default_value() {
+        // Greptile follow-up: { key: local = fallback } nests an
+        // assignment_pattern under pair_pattern's value field — a distinct
+        // shape from the plain shorthand default ({ a = 1 }) case above.
+        // Without this branch the pair_pattern handler rejected the nested
+        // assignment_pattern and `local` never got a Definition at all.
+        let s = parse_js("const { key: local = fallback } = someValue;");
+        let def = s
+            .definitions
+            .iter()
+            .find(|d| d.name == "local")
+            .expect("should extract local definition");
+        assert_eq!(def.kind, "constant");
+        assert!(!s.definitions.iter().any(|d| d.name == "key"), "should not use the original key");
     }
 
     /// Regression test for issue #1271: native engine missing receiver edges.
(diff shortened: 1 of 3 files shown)


carlos-alm:
Good catch — fixed in 0fa8d572. `extractDestructuredBindings`/`extract_destructured_bindings`'s `pair_pattern`/`pair` branch now also handles an `assignment_pattern` nested under `value` (`const { key: local = fallback } = x`), mirroring the identical branch already in `extractDynamicImportNames`/`collect_object_pattern_names` since #1824. Added regression tests in both engines confirming `local` gets a `constant` Definition and `key` does not.

carlos-alm:
@greptileai

Issue #2053, opened 2026-07-08
Native Rust computes trivial complexity for bodyless Rust trait methods; WASM correctly omits them (engine parity gap)


carlos-alm:
## Summary

`handle_trait_item` in `crates/codegraph-core/src/extractors/rust_lang.rs` calls `compute_all_metrics(&child, source, "rust")` unconditionally for every trait member, regardless of whether `child.kind()` is `function_signature_item` (a required trait method with **no default implementation**, no body) or `function_item` (a default-implemented trait method, has a body):

```rust
if child.kind() != "function_signature_item" && child.kind() != "function_item" {
    continue;
}
if let Some(meth_name) = child.child_by_field_name("name") {
    symbols.definitions.push(Definition {
        name: format!("{}.{}", trait_name, node_text(&meth_name, source)),
        kind: "method".to_string(),
        line: start_line(&child),
        end_line: Some(end_line(&child)),
        decorators: None,
        complexity: compute_all_metrics(&child, source, "rust"),
        cfg: build_function_cfg(&child, "rust", source),
        children: None,
        ...
    });
}
```

`compute_all_metrics` doesn't check whether the passed node's kind is actually a member of the language's own `function_nodes` rule set — it just walks whatever node it's given, counting branch/nesting node types among its descendants. For a `function_signature_item` (no body, just `fn name(&self, ...) -> T;`), this produces a **trivial but non-null** `ComplexityMetrics { cognitive: 0, cyclomatic: 1, ... }`, because a bodyless signature has no descendants that look like branches.

Meanwhile, the WASM/TS side's `COMPLEXITY_RULES` for Rust (`src/ast-analysis/rules/rust.ts`) deliberately excludes `function_signature_item` from `functionNodes`:

```ts
functionNodes: new Set(['function_item', 'closure_expression']),
```

So the WASM complexity visitor never even attempts to compute complexity for a `function_signature_item` — it produces **no complexity entry at all** for that definition.

## Reproduction

```rust
trait Repo {
    fn save(&mut self, id: &str, value: i32) -> bool;   // required, no body

    fn find(&self, id: &str) -> Option<i32> {             // default impl, has body
        if id.is_empty() { return None; }
        for i in 0..3 { if i == 1 { return Some(i); } }
        None
    }
}
```

```
$ codegraph complexity --file repo.rs -T --json --engine wasm
functions: 1   [ Repo.find ]

$ codegraph complexity --file repo.rs -T --json --engine native
[...]

PR #2284, opened 2026-08-05
fix: skip complexity/CFG for bodyless functions in the WASM engine (#2055)

### LINKED ISSUE #2055: C# WASM complexity visitor computes trivial complexity for bodyless interface method signatures; native correctly omits them
## Summary

For C#, native and WASM disagree on whether an interface method signature (no body) should get a complexity entry at all:

- `crates/codegraph-core/src/extractors/csharp.rs`'s `handle_interface_decl` explicitly passes `complexity: None, cfg: None` for interface methods, with a comment: "Interface method declarations have no body — skip CFG and complexity to mirror the WASM extractor and avoid producing meaningless metrics for body-less declarations."
- But the WASM/TS side's `COMPLEXITY_RULES` for C# (`src/ast-analysis/rules/csharp.ts`) puts `method_declaration` in `functionNodes` unconditionally (matching both bodied class methods and bodyless interface signatures, since C# uses the *same* node type for both):
  ```ts
  functionNodes: new Set(['method_declaration', 'constructor_declaration', 'lambda_expression', 'local_function_statement']),
  ```
  The complexity visitor walks any node matching `functionNodes` without checking whether it actually has a body — so once the file-level gate lets the visitor run, it computes a trivial-but-non-null `{ cyclomatic: 1, cognitive: 0 }` for the interface stub too.

So the "mirror the WASM extractor" comment in `csharp.rs` describes intent, not actual current WASM behavor — WASM does **not** skip bodyless `method_declaration` nodes; it just walks them and gets a trivial result.

## Reproduction

```csharp
interface IRepo {
    bool Save(string id, int value);
}

class Repo : IRepo {
    public bool Save(string id, int value) {
        if (value < 0) { return false; }
        for (int i = 0; i < 3; i++) { if (i == value) { return true; } }
        return true;
    }
}
```

```
$ codegraph complexity --file Repo.cs -T --json --engine native
functions: 1   [ Repo.Save ]

$ codegraph complexity --file Repo.cs -T --json --engine wasm
functions: 2   [ Repo.Save, IRepo.Save ]   ← IRepo.Save has cyclomatic: 1, cognitive: 0
```

Per CLAUDE.md: "Both engines must produce identical results. If they diverge, the less-accurate engine has a bug." Here WASM is the less-accurate engine — it fabricates a complexity entry for a method with no executable body.

## Suggested fix

The WASM complexity visitor (`src/ast-analysis/visitors/complexity-visitor.ts` / the shared walker) needs a real "does this function-node have a body" check before computing metrics — not just node-type membership in `functionNodes` — so it agrees with native's explicit exclusion for bodyless `method_declaration` nodes. This likely affects any language where the same node type is used for both bodied and bodyless declarations (C# interfaces; Java interfaces/abstract methods happen to already agree between engines only because native's `java.rs` *also* computes unconditionally for these — a related but separate inconsistency worth auditing at the same time).

## Discovery context

Found while fixing #1922 (WASM complexity engine skips entire file when every function has a dotted name) — live-testing native vs WASM parity for C# interface + implementing-class fixtures as part of verifying the mirrored `Definition.bodyless` signal added in that fix. Independent, pre-existing bug in how each engine handles signature-only declarations that happen to share a node type with real methods — not fixed there to keep that PR scoped to the file-level complexity gate. Related to #2053 (same class of "signature-only node computed as if bodied" bug, opposite direction, for Rust trait methods).

carlos-alm:
## Summary

`storeComplexityResults`/`storeCfgResults` (`src/ast-analysis/apply-results.ts`) merge the shared visitor's per-file walk results onto every `function`/`method` `Definition` without checking whether it actually has a body. The visitor walks by node-TYPE membership in `functionNodes`, which for C#/Java shares one node type (`method_declaration`) between a bodied class method and a bodyless interface/abstract signature — so it computes a trivial-but-real result for the bodyless one too, and this merge step blindly attached it. Native's `csharp.rs` already explicitly skips this case; the "mirror the WASM extractor" comment in `csharp.rs` described intent, not actual WASM behavior.

Fix: add the same `hasFuncBody` gate already used elsewhere in this codebase (`classifyDefinitionForNativeBulk`, `initWasmParsersIfNeeded`) to both merge functions — this is the shared code both `ast-analysis/engine.ts` and `domain/wasm-worker-entry.ts` import, so it fixes every caller (single-file and worker-pool builds) in one place. Also gated `computeJsFallbackMetrics`'s independent per-definition loop in `features/complexity.ts` for defense-in-depth on that separate fallback path.

**Also fixed `java.rs` in this PR.** Fixing the shared engine generically also affects Java, which the issue's own discovery context flagged as a related inconsistency: native's `java.rs` *also* computed complexity/cfg unconditionally for interface methods — the same bug shape #2053/#2054 already fixed for Rust/PHP. Leaving it unfixed here would have introduced a **new** WASM/native divergence for Java (this PR would make WASM correct while native stayed wrong). Fixed `java.rs`'s `handle_interface_decl` to match `csharp.rs`'s existing pattern.

## Verification

- `cargo test --lib` — 761/761 passed (including a new regression test for `java.rs`)
- `npm run lint` — clean
- `npm test` — 273/273 test files, 4438 passed (including 2 new regression tests in `apply-results.test.ts`)
- Native and WASM produce **identical** output on:
  - the issue's C# repro (single interface + implementing class)
  - a Java interface + default-method repro (default methods still get full metrics — only the bodyless required signature is skipped)
  - a 4-file multi-file build (exercises the worker-pool code path, not just the single-file path)

## Test plan

- [x] `npm run lint`
- [x] `npm test`
- [x] Native vs WASM diff on C# and Java repros (identical output)
- [x] Multi-file build exercises the worker-pool path
- [x] Default/concrete methods still get full metrics; only bodyless signatures are skipped

Closes #2055

### DIFF
--- a/crates/codegraph-core/src/extractors/java.rs
+++ b/crates/codegraph-core/src/extractors/java.rs
@@ -167,16 +167,30 @@ fn handle_interface_decl(node: &Node, source: &[u8], symbols: &mut FileSymbols) 
             let Some(child) = body.child(i) else { continue };
             if child.kind() != "method_declaration" { continue; }
             if let Some(meth_name) = child.child_by_field_name("name") {
+                // Interface method declarations have no body (unless it's a
+                // default/static method with one) — skip CFG and complexity
+                // for the bodyless case to mirror the WASM extractor and
+                // avoid producing meaningless metrics for body-less
+                // declarations. Mirrors csharp.rs's handle_interface_decl.
+                let is_bodyless = child.child_by_field_name("body").is_none();
                 symbols.definitions.push(Definition {
                     name: format!("{}.{}", iface_name, node_text(&meth_name, source)),
                     kind: "method".to_string(),
                     line: start_line(&child),
                     end_line: Some(end_line(&child)),
                     decorators: None,
-                    complexity: compute_all_metrics(&child, source, "java"),
-                    cfg: build_function_cfg(&child, "java", source),
+                    complexity: if is_bodyless {
+                        None
+                    } else {
+                        compute_all_metrics(&child, source, "java")
+                    },
+                    cfg: if is_bodyless {
+                        None
+                    } else {
+                        build_function_cfg(&child, "java", source)
+                    },
                     children: None,
-                    bodyless: Some(child.child_by_field_name("body").is_none()),
+                    bodyless: Some(is_bodyless),
                     content_hash: None,
                     accessor_kind: None,
                 });
@@ -578,6 +592,34 @@ mod tests {
             .find(|d| d.name == "Repo.save")
             .unwrap();
         assert_ne!(concrete_save.bodyless, Some(true));
+    }
+
+    /// Regression test for #2055: an interface method with no body must not
+    /// get a fabricated complexity/CFG entry — mirrors csharp.rs's
+    /// handle_interface_decl, which already skips these. A default method
+    /// (has a body) must still get real metrics.
+    #[test]
+    fn interface_method_has_no_complexity_default_method_does() {
+        let s = parse_java(
+            "interface Repo {\n\
+               boolean save(String id, int value);\n\
+               default boolean saveOrDefault(String id, int value) {\n\
+                 if (value < 0) { return false; }\n\
+                 return true;\n\
+               }\n\
+             }\n",
+        );
+        let save = s.definitions.iter().find(|d| d.name == "Repo.save").unwrap();
+        assert!(save.complexity.is_none());
+        assert!(save.cfg.is_none());
+
+        let default_method = s
+            .definitions
+            .iter()
+            .find(|d| d.name == "Repo.saveOrDefault")
+            .unwrap();
+        assert!(default_method.complexity.is_some());
+        assert!(default_method.cfg.is_some());
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


carlos-alm:
Addressed the "single-line bodies lose analysis" finding on `src/ast-analysis/apply-results.ts:102` (and the equivalent in `storeCfgResults`, plus the `computeJsFallbackMetrics` fallback in `src/features/complexity.ts`).

Confirmed the bug: `hasFuncBody(def)` requires `endLine > line` in addition to `!bodyless`, so a genuinely bodied single-line function/method (`endLine === line`) was wrongly treated as bodyless and lost its already-computed complexity/CFG result.

Fix: all three call sites now check `def.bodyless !== true` directly instead of `hasFuncBody(def)` — the actual, reliable per-definition signal every extractor sets, with no line-span assumption. Verified a single-line method sharing a file with a multi-line method now correctly gets its complexity/CFG attached, and the original bodyless-interface-method fix is unaffected. Added 2 regression tests (one per merge function) covering a genuinely bodied single-line definition.

carlos-alm:
@greptileai

PR #2286, opened 2026-08-05
fix: correct switch/case complexity shadowing and comment-prefix undercounting (#2058)

### LINKED ISSUE #2058: Native complexity rules misclassify switch/case for c/cpp/kotlin/scala (and drop the switch container entirely for swift); comment-line undercounting for c/cpp/kotlin/swift/scala
## Summary

While porting the native `LangRules`/`HalsteadRules` for c/cpp/kotlin/swift/scala/bash to the TS/WASM side (#1923), several latent correctness issues were found in the **existing native rules** (`crates/codegraph-core/src/ast_analysis/complexity.rs`). All are pre-existing — unrelated to the TS-side port — and were mirrored verbatim into the new TS rules for exact engine parity, since fixing rule *semantics* is a separate, larger workstream from fixing rule *presence* (which is #1923's scope). Filing here so they don't get silently perpetuated.

## Finding 1: case-node shadowing in `C_RULES`, `CPP_RULES`, `KOTLIN_RULES`, `SCALA_RULES`

In `walk()`, node classification order is: logical-op check → optional-chain check → **`is_branch(kind)` (returns unconditionally on match)** → pattern-C-else check → `is_case(kind)`. So any node type listed in *both* `branch_nodes` and `case_nodes` is always handled as a generic branch — the `case_nodes` arm for that type is dead code.

Four of the five affected languages put the actual per-case node type in `branch_nodes` alongside (or instead of) the switch container:

- `C_RULES` / `CPP_RULES`: `branch_nodes` includes `"case_statement"` (also listed in `case_nodes`); the switch container `"switch_statement"` is **not** in `branch_nodes` at all (only in `switch_like_nodes`, which is only consulted from inside the branch handler).
- `KOTLIN_RULES`: `branch_nodes` includes both the container `"when_expression"` *and* the case `"when_entry"` (also in `case_nodes`).
- `SCALA_RULES`: `branch_nodes` includes both the container `"match_expression"` *and* the case `"case_clause"` (also in `case_nodes`).

Every unaffected switch-having language in the same file (JS/TS, Java, C#, PHP, Ruby, Bash) puts only the *container* in `branch_nodes` + `switch_like_nodes` (contributing nesting once, cyclomatic 0) and puts the *case* node in `case_nodes` only (flat `cyclomatic += 1`, no per-case cognitive/nesting weight). For C/C++/Kotlin/Scala, each case arm instead gets the full "regular branch" treatment (`cognitive += 1 + nesting_level`, `cyclomatic += 1`), which very likely inflates cognitive-complexity scores for every switch/when/match statement in these four languages relative to the metric's intended semantics (and relative to every other supported language).

## Finding 2: `SWIFT_RULES` drops the switch container from complexity entirely

`SWIFT_RULES.branch_nodes` and `.nesting_nodes` never include `"switch_statement"` — only the per-case `"switch_entry"` appears (in both `branch_nodes` and `case_nodes`, hitting the same shadowing bug as Finding 1). Net effect: a Swift `switch` statement contributes zero nesting for its cases and each `case` arm is scored as an unnested "regular branch" rather than as a switch case.

## Finding 3: `comment_prefixes()` omits `*`/`*/` continuation-line prefixes for c/cpp/kotlin/swift/scala

```rust
"javascript" | "typescript" | "tsx" | "go" | "rust" | "java" | "csharp" => {
    &["//", "/*", "*", "*/"]
}
...
"c" | "cpp" => &["//", "/*"],
"kotlin" => &["//", "/*"],
"swift" => &["//", "/*"],
"scala" => &["//", "/*"],
```

All of these are C-family/JVM languages with the same `/** ... */` block-comment style as JS/Java/C#, but only get the 2-entry `["//", "/*"]` list instead of the 4-entry list that also matches bare `*` continuation lines and closing `*/` lines. For a function with a multi-line Javadoc-style comment, only the first line (`/**`) is counted as a comment line for these 5 languages — undercounting `commentLines` and skewing the Maintainability Index input.

## Scope note

Not fixed as part of #1923, which is scoped to *adding* missing complexity/Halstead configs for languages that have none — not auditing the *correctness* of configs that already exist and are already used in production by the native engine today (real native users are seeing these numbers right now; changing the methodology needs its own careful before/after validation). Both TS-side mirrors intentionally match current native behavior byte-for-byte in #1923's PR to keep native/WASM parity. Fixing the underlying `branch_nodes`/`case_nodes`/`switch_like_nodes` design for c/cpp/kotlin/swift/scala correctly — likely moving the container into `branch_nodes`+`switch_like_nodes`+`nesting_nodes` and dropping the per-case node from `branch_nodes` on both engines, matching the pattern already used by JS/Java/C#/PHP/Ruby/Bash — is tracked here as follow-up work.

carlos-alm:
## Summary

Three pre-existing correctness issues in the native `LangRules`/`HalsteadRules` for c/cpp/objc/kotlin/swift/scala, mirrored verbatim into the TS rules by #1923 to keep engine parity, are fixed in both engines:

**1. Case-node shadowing** (`C_RULES`, `CPP_RULES`, `OBJC_RULES`, `KOTLIN_RULES`, `SCALA_RULES`): the walk/DFS classifier always returns after a `branch_nodes` match, so a node type listed in *both* `branch_nodes` and `case_nodes` is always treated as a generic branch — the `case_nodes` arm never fires. Fixed by removing the case-arm node type from `branch_nodes`, keeping it in `case_nodes` only — matching the pattern JS/Java/C#/PHP/Ruby/Bash already use.

**2. `SWIFT_RULES` dropped the switch container entirely**: `switch_statement` was absent from both `branch_nodes` and `nesting_nodes`, so a Swift `switch` contributed zero nesting for its cases, and `switch_entry` was double-booked hitting the same shadowing bug. Fixed by adding `switch_statement` to `branch_nodes` + `nesting_nodes` and removing `switch_entry` from `branch_nodes`.

**3. `comment_prefixes()`/`COMMENT_PREFIXES`**: c/cpp/cuda/objc/kotlin/swift/scala used a 2-entry list instead of the 4-entry list JS/Java/C# use, missing bare `*`/`*/` continuation lines and undercounting `commentLines` for any multi-line Javadoc-style comment.

**Also fixed ObjC**, added after #1923/#2058 was filed — it inherited the same case-shadowing and missing-switch-container bugs via copy-paste from C's rules. CUDA reuses `CPP_RULES`/`complexityCpp` as-is, so it's auto-fixed.

Cyclomatic values are numerically unaffected for existing tests — each case arm always contributes +1 whether via the buggy generic-branch path or the correct case-only path, and the container nets to zero either way. The bug only showed up in **cognitive** complexity, which existing tests never asserted (that's why it survived until now).

## Verification

- `cargo test --lib` — 768/768 passed (including 8 new regression tests)
- `npm run lint` — clean
- `npm test` — 273/273 test files, 4447 passed (including 7 new regression tests)
- Native and WASM produce **identical** `codegraph complexity --health --json` output on hand-built switch/when/match fixtures (including multi-value cases: `case 2: case 3:`, `2, 3 ->`, `case 2 | 3 =>`) and a multi-line-comment fixture, for all six affected languages

## Test plan

- [x] `npm run lint`
- [x] `npm test`
- [x] Native vs WASM diff on switch/when/match fixtures for c/cpp/objc/kotlin/swift/scala (identical output)
- [x] Native vs WASM diff on a multi-line-comment fixture for c and kotlin (identical output)

Closes #2058

### DIFF
--- a/crates/codegraph-core/src/ast_analysis/complexity.rs
+++ b/crates/codegraph-core/src/ast_analysis/complexity.rs
@@ -344,13 +344,21 @@ pub static PHP_RULES: LangRules = LangRules {
 // style: an else_clause node wraps either a nested if_statement for
 // `else if` or the plain else body), NOT Pattern C (Go/Java style, where the
 // `alternative` field holds the substatement directly with no wrapper node).
+// `walk()`'s node classification always returns after a `is_branch(kind)`
+// match, so a type listed in BOTH branch_nodes and case_nodes is always
+// treated as a generic branch — the case_nodes arm never fires (issue
+// #2058). `switch_statement` (the container) belongs in branch_nodes +
+// nesting_nodes (net-zero cyclomatic via switch_like_nodes, contributing
+// nesting once, matching JS/Java/C#/PHP/Ruby/Bash); `case_statement` (each
+// arm) belongs in case_nodes ONLY (flat `cyclomatic += 1`, no per-case
+// cognitive/nesting weight) — not in branch_nodes.
 pub static C_RULES: LangRules = LangRules {
-    branch_nodes: &["if_statement", "else_clause", "for_statement", "while_statement", "do_statement", "case_statement", "conditional_expression"],
+    branch_nodes: &["if_statement", "else_clause", "for_statement", "while_statement", "do_statement", "switch_statement", "conditional_expression"],
     case_nodes: &["case_statement"],
     logical_operators: &["&&", "||"],
     logical_node_types: &["binary_expression"],
     optional_chain_type: None,
-    nesting_nodes: &["if_statement", "for_statement", "while_statement", "do_statement", "conditional_expression"],
+    nesting_nodes: &["if_statement", "for_statement", "while_statement", "do_statement", "switch_statement", "conditional_expression"],
     function_nodes: &["function_definition"],
     if_node_type: Some("if_statement"),
     else_node_type: Some("else_clause"),
@@ -367,13 +375,14 @@ pub static C_RULES: LangRules = LangRules {
 // syntax), and parsing sample CUDA control flow confirms identical
 // if_statement/else_clause/for_statement/while_statement/switch_statement/
 // binary_expression node kinds to plain C++.
+// Same branch_nodes/case_nodes fix as C_RULES (issue #2058) — see comment there.
 pub static CPP_RULES: LangRules = LangRules {
-    branch_nodes: &["if_statement", "else_clause", "for_statement", "for_range_loop", "while_statement", "do_statement", "case_statement", "conditional_expression", "catch_clause"],
+    branch_nodes: &["if_statement", "else_clause", "for_statement", "for_range_loop", "while_statement", "do_statement", "switch_statement", "conditional_expression", "catch_clause"],
     case_nodes: &["case_statement"],
     logical_operators: &["&&", "||"],
     logical_node_types: &["binary_expression"],
     optional_chain_type: None,
-    nesting_nodes: &["if_statement", "for_statement", "for_range_loop", "while_statement", "do_statement", "catch_clause", "conditional_expression"],
+    nesting_nodes: &["if_statement", "for_statement", "for_range_loop", "while_statement", "do_statement", "switch_statement", "catch_clause", "conditional_expression"],
     function_nodes: &["function_definition"],
     if_node_type: Some("if_statement"),
     else_node_type: Some("else_clause"),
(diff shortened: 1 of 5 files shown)


carlos-alm:
Re: "Bare star misclassifies source lines" (`src/ast-analysis/metrics.ts:81-87`, `crates/codegraph-core/src/ast_analysis/complexity.rs`).

Verified this is a real limitation, but it's pre-existing and not introduced by this PR: it already reproduces on `main` for Rust today, on **both engines**, completely unrelated to any of this PR's changes:

```rust
fn deref_heavy(ptr: *mut i32) -> i32 {
    unsafe {
        *ptr = 5;
        *ptr = *ptr + 1;
        return *ptr;
    }
}
```

```
$ codegraph complexity --health -T --json --engine wasm    → loc:7, sloc:5
$ codegraph complexity --health -T --json --engine native  → loc:7, sloc:5  (identical)
```

`*ptr = 5;` and `*ptr = *ptr + 1;` both get wrongly counted as comment lines, since Rust already uses the 4-entry `["//", "/*", "*", "*/"]` prefix list (needed for `/** */`-style doc comments) and pointer-dereference assignment also starts with `*`. C# and Go have the same exposure today for the same reason.

This PR's scope (per #2058) was making the prefix list *consistent* across languages/engines — c/cpp/objc/kotlin/swift/scala previously used an inconsistent 2-entry list despite having the same `/** */` comment style as JS/Java/C#/Go/Rust. Extending the already-shipped 4-entry heuristic to those languages doesn't introduce a new class of bug; it applies the existing (imperfect) design consistently.

Fixing the underlying naive per-line prefix check properly means making comment detection block-comment-state-aware (track whether a line is inside an unterminated `/* ... */`, not just check if it starts with `*`) — a real algorithmic change affecting every C-style/JSDoc-style language on both engines, not specific to the languages this PR touches. Filed as #2287 to track that properly rather than scope-creep it into this consistency fix.

carlos-alm:
@greptileai

Issue #2288, opened 2026-08-05
Workspace root-import resolution via exports field returns a node_modules-symlink path that never matches the tracked file node


carlos-alm:
## Summary

`resolveViaWorkspace()`'s root-import branch (`src/domain/graph/resolve.ts`) tries the package's `exports` field first via `resolveViaExports()`, which locates the package directory through `findPackageDir()` — a `node_modules` walk. For a real monorepo where workspace tools (npm/yarn/pnpm workspaces) symlink workspace packages into `node_modules`, this returns a path like `node_modules/@myorg/core/lib/index.js`.

But `detectWorkspaces()`/`expandWorkspacePatterns()` (the same file's workspace *detection*, driven by the `workspaces`/`packages` glob in the root `package.json`) registers each package's `info.dir` from the **glob-matched real path** (`packages/core`), which is also where the project's own file collector finds and tracks the source file (`node_modules` is excluded from file collection via `IGNORE_DIRS`).

So for a workspace package whose `package.json` has only an `exports` field (no `main`), the root-import resolution returns `node_modules/@myorg/core/lib/index.js` — a path string that never matches the graph's actual tracked node for that file (`packages/core/lib/index.js`). The result: no `imports` edge, no `calls` edge, and no cross-file call attribution for that import — even though the file itself is correctly parsed and present in the graph under its real path.

## Reproduction

```
root/
  package.json            { "workspaces": ["packages/*"] }
  node_modules/@myorg/core -> ../../packages/core   (symlink, as workspace tools create)
  packages/core/
    package.json           { "name": "@myorg/core", "exports": "./lib/index.js" }
    lib/index.js            function greet() { ... }
  packages/app/src/main.js  import { greet } from "@myorg/core"; console.log(greet());
```

```
$ codegraph build --engine wasm   # or --engine native — identical result
$ sqlite3 .codegraph/graph.db "SELECT DISTINCT kind FROM edges;"
contains
```

No `imports` or `calls` edge at all. Changing `packages/core/package.json` to use `"main": "./lib/index.js"` instead of `exports` makes it work correctly (`resolveViaWorkspace()`'s entry fallback uses `info.entry`, which IS the real `packages/`-based path) — confirming the `exports`-first branch specifically is what returns the wrong path shape.

## Confirmed pre-existing, not engine-specific

[...]

PR #2289, opened 2026-08-05
feat: port package.json exports-field resolution to native engine (#2060)

### LINKED ISSUE #2060: Native engine: no package.json "exports" field resolution (resolveViaExports has no Rust counterpart)
## Context

Discovered while fixing #1927 (native engine had no workspace-awareness in `resolve_import_path_inner`).

## Description

`resolveViaExports()` in `src/domain/graph/resolve.ts` resolves a bare specifier (e.g. `import "some-pkg/sub"`) through a package's `package.json` `exports` field — conditional exports, subpath patterns (`"./lib/*"`), array fallbacks, etc. It's used for:

1. Plain `node_modules` bare specifiers in `resolveImportPathJS()`.
2. As the first attempt inside `resolveViaWorkspace()` for monorepo workspace packages (before falling back to `main`/`source`/index-file resolution).

There is **no equivalent function anywhere in `crates/codegraph-core`**. The native engine's `resolve_import_path_inner` (`crates/codegraph-core/src/domain/graph/resolve.rs`) only implements `resolve_via_alias` (tsconfig/jsconfig path aliases) for non-relative specifiers — a bare specifier pointing at a package that only exposes its entry via the `exports` field (no matching `main`/index-file convention) resolves incorrectly or not at all under native, while WASM/JS resolves it correctly.

The #1927 fix (native workspace-awareness) explicitly does **not** attempt to close this gap — `resolve_via_workspace()` in Rust skips the `exports`-field lookup and goes straight to `main`/`source`/index-file resolution, so workspace packages that rely solely on a conditional `exports` map (no `main` field, or a `main` that differs from the real entry) may still resolve differently between engines.

## Suggested next step

Port `resolveViaExports()`'s logic (`parseBareSpecifier` — already mirrored in Rust as of #1927 — `findPackageDir`, `getPackageExports`, `resolveCondition`, `resolveSubpathMap`, `matchSubpathPattern`) into `crates/codegraph-core/src/domain/graph/resolve.rs`, wire it into `resolve_non_relative_import` (and `resolve_via_workspace`'s root/subpath branches) the same way the JS engine does, and add a `crates/codegraph-core` ↔ `src/domain/graph/resolve.ts` parity test fixture exercising a package with only an `exports` field (no `main`).

## Source

Found during: fix for #1927 (native monorepo workspace package import resolution)

carlos-alm:
## Summary

`resolveViaExports()` in `src/domain/graph/resolve.ts` resolves a bare specifier (e.g. `import "some-pkg/sub"`) through a package's `package.json` `exports` field — conditional exports, subpath patterns (`"./lib/*"`), array fallbacks. There was no equivalent anywhere in `crates/codegraph-core`: `resolve_import_path_inner` only implemented `resolve_via_alias` for non-relative specifiers, so a bare specifier pointing at a package that only exposes its entry via `exports` (no matching `main`/index-file convention) resolved incorrectly or not at all under native, while WASM/JS resolved it correctly.

Ports `findPackageDir`, `getPackageExports` (with the same packageDir-keyed cache), `resolveCondition` (import/require/default priority), `matchSubpathPattern`, `resolveSubpathMap`, and `resolveViaExports` into `resolve.rs` using `serde_json::Value` (already a crate dependency) — mirroring the TS dynamic-typing dispatch (string / array / conditions-object / subpath-map) exactly. Deliberately reads `package.json` directly via `std::fs` rather than the `known_files`-aware `file_exists()` helper: `exports` resolution reaches into `node_modules`, which is never part of the project's tracked source file list.

Wired into `resolve_via_workspace` (both the root-import and subpath branches, matching `resolveViaWorkspace()`'s "try exports first" order) and into `resolve_non_relative_import` (the plain `node_modules` bare-specifier case, matching `resolveImportPathJS()`'s fallback order: alias → rust-path → workspace → exports → raw fallback).

## Verification

- `cargo test --lib` — 774/774 passed (including 6 new tests using real filesystem fixtures)
- `npm run lint` — clean
- `npm test` — 273/273 test files, 4447 passed
- Verified against a real monorepo fixture via the CLI: native and WASM produce **byte-identical** graphs

**Note on scope**: end-to-end verification surfaced a separate, pre-existing bug already present on the WASM/JS side (confirmed by calling `resolveViaWorkspace()`/`resolveViaExports()` directly, independent of this PR): a workspace package's root-import resolution via `exports` returns a `node_modules`-symlink path that doesn't match the graph's tracked file node (found via the workspace glob's real path), so no `imports`/`calls` edge gets created for that specific case. Filed as #2288 — this PR's native port correctly *mirrors* that existing WASM behavior (parity is maintained, confirmed identical on both engines), so fixing it is out of scope here.

## Test plan

- [x] `npm run lint`
- [x] `npm test`
- [x] Real monorepo fixture: native vs WASM produce identical graphs
- [x] Unit tests: simple string exports, wildcard subpath, conditional exports, no-exports-field, end-to-end via `resolve_non_relative_import`, workspace-exports-wins-over-entry

Closes #2060

### DIFF
--- a/crates/codegraph-core/Cargo.toml
+++ b/crates/codegraph-core/Cargo.toml
@@ -11,7 +11,16 @@ crate-type = ["cdylib"]
 napi = { version = "3", features = ["serde-json"] }
 napi-derive = "3"
 serde = { version = "1", features = ["derive"] }
-serde_json = "1"
+# `preserve_order` keeps object keys in JSON-source declaration order (an
+# IndexMap internally) instead of the default sorted-by-key BTreeMap. Needed
+# for package.json `exports` subpath-map wildcard resolution (resolve.rs) to
+# match resolveViaExports()'s `Object.entries()` iteration order in
+# resolve.ts — when overlapping wildcard patterns both match, whichever
+# pattern is declared first in the manifest wins on both engines; without
+# this feature, native would silently pick the lexicographically-first
+# pattern instead, diverging from WASM for any manifest where declaration
+# order differs from sort order.
+serde_json = { version = "1", features = ["preserve_order"] }
 tree-sitter = "0.25"
 tree-sitter-javascript = "0.25"
 tree-sitter-typescript = "0.23"
--- a/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
@@ -150,9 +150,13 @@ fn pipeline_setup(
     let force_full_rebuild = check_version_mismatch(conn);
     let workspaces = resolve::workspaces_from_packages(&workspace_packages);
     // Reset once per build, mirroring `setWorkspaces()`'s
-    // `_workspaceResolvedPaths.clear()` in resolve.ts — must happen before
-    // Stage 6/6b resolve any imports below.
+    // `_workspaceResolvedPaths.clear()`/`clearExportsCache()` in resolve.ts —
+    // must happen before Stage 6/6b resolve any imports below. Clearing the
+    // exports cache here too matters for a long-lived native process (MCP
+    // server, watch mode) running multiple builds: a dependency's
+    // `package.json` can change between builds (issue #2060).
     resolve::reset_workspace_resolved_paths();
+    resolve::clear_exports_cache();
 
     Ok(PipelineSetup {
         config,
--- a/crates/codegraph-core/src/domain/graph/resolve.rs
+++ b/crates/codegraph-core/src/domain/graph/resolve.rs
@@ -1915,6 +1915,37 @@ mod tests {
     }
 
     #[test]
+    fn resolve_via_exports_wildcard_selection_follows_manifest_declaration_order() {
+        // Regression guard (issue #2060, caught by Greptile review):
+        // resolveSubpathMap()/resolve_subpath_map() must check overlapping
+        // wildcard keys in the manifest's DECLARATION order (matching
+        // JS's Object.entries()), not sorted-by-key order. "./lib/zeta-*" is
+        // declared FIRST here but sorts AFTER "./lib/*" lexicographically
+        // ('*' < 'z' in ASCII) — so a subpath matching both must resolve via
+        // the first-declared pattern regardless of key sort order. Requires
+        // serde_json's `preserve_order` feature (Cargo.toml); without it,
+        // serde_json::Map iterates as a sorted BTreeMap and this test fails.
+        let tmp = make_exports_fixture(
+            "codegraph_exports_wildcard_order_test",
+            "some-pkg",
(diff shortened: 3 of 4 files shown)


carlos-alm:
Both findings confirmed as real bugs and fixed:

**Issue 1 (stale exports cache):** `clear_exports_cache()` existed but was never actually wired into production — only test code called it. Added it alongside `reset_workspace_resolved_paths()` at both once-per-build entry points (`resolve_imports` in `lib.rs`, `pipeline_setup` in `pipeline.rs`), matching `setWorkspaces()`'s cache-clearing pairing in `resolve.ts` exactly.

**Issue 2 (wildcard selection order):** confirmed `serde_json::Map` defaults to a sorted `BTreeMap` without the `preserve_order` feature, so overlapping wildcard keys were checked in lexicographic order instead of manifest declaration order. Enabled `preserve_order` in `Cargo.toml` (an already-locked transitive dependency, so no new `Cargo.lock` entries). Added a regression test with two overlapping patterns where declaration order and sort order differ — verified it fails without the fix and passes with it.

Also fixed the CI failure (Native host build windows-2022): the new tests built expected path strings via `PathBuf::join()`, which doesn't consistently forward-slash-normalize on Windows the way the actual resolver output does — fixed the test assertions, not the resolver.

`cargo test --lib` (775/775), `npm run lint`, `npm test` (4447/4447) all pass.

carlos-alm:
@greptileai

carlos-alm:
Re: "Watch rebuilds retain stale exports" (`crates/codegraph-core/src/lib.rs:158`).

Verified this is a real characteristic, but it's pre-existing and symmetric across both engines, not a native-specific gap or a regression from this PR. `resolve_import` (the single-import FFI entry point used by the incremental rebuild path) intentionally does NOT call `reset_workspace_resolved_paths()` either — only the batch entry point (`resolve_imports`) and `pipeline_setup` do, both full-build-only call sites.

Checked the WASM/JS side directly: `watcher.ts` → `rebuildFile()` (`incremental.ts`) has zero calls to `detectWorkspaces()`/`setWorkspaces()`/`clearExportsCache()` either — `_exportsCache`/`_workspaceCache` are populated once by the full-build pipeline and never refreshed during watch mode on that engine either.

Clearing the exports cache on every single incremental resolve call would also be actively wrong: it would force a `package.json` re-read/re-parse on every file save in a watch session, defeating the purpose of the cache, for a scenario (a dependency's `package.json` changing mid-watch-session) that's rare compared to the common case (only source files changing).

Filed as #2290 to track properly — fixing it means changing watch mode's file-change-detection scope (to also watch `package.json` files and invalidate/rebuild accordingly) for **both** engines, which is a watcher-architecture change, not something in scope for this resolver-parity port.

carlos-alm:
@greptileai

PR #2291, opened 2026-08-06
fix: prevent resolveThisDispatch from matching an unrelated same-named class in another file

### LINKED ISSUE #2062: resolveThisDispatch resolves this/super dispatch to the wrong file when a same-named base class exists elsewhere with no same-file match

carlos-alm:
## Summary

`resolveThisDispatch`'s `this`/`super` hierarchy walk resolved to the wrong file whenever two independent files each declared a class with the same name and the caller's own same-named ancestor had no explicit definition of the dispatched method (e.g. an implicit default constructor). `chaCtx.parents` is keyed by bare class name, so the walk had no way to tell "genuine cross-file heritage" apart from "unrelated class that merely shares a name."

Fix: before accepting a cross-file match with no same-file candidate, check whether a class of the same name is *also* declared in the caller's own file. If so, that same-file class is the caller's real ancestor at this step — it simply doesn't define the method — so keep walking instead of accepting the unrelated cross-file match. When the class genuinely isn't declared anywhere in the caller's file, the cross-file match is legitimate heritage and is returned as before.

This is TS-only logic (`src/domain/graph/builder/cha.ts`) shared by both engines — the native path runs this same JS CHA post-pass on top of native edges, so no Rust mirror change is needed.

## Test plan
- [x] New `tests/unit/cha.test.ts` — 6 cases: the exact two-`Shape`-class repro, legitimate cross-file heritage, same-file preference (regression), multi-hop walk past a collided intermediate class, no-`callerFile` fallback, non-method callerName
- [x] `npm run lint` clean
- [x] `npx tsc --noEmit` clean
- [x] Full `npm test` — 274 files / 4453 passed, no regressions
- [x] Manually verified against the issue's exact repro on both native and WASM engines: 5 false edges eliminated, correct edges preserved

Closes #2062

### DIFF
--- a/src/domain/graph/builder/cha.ts
+++ b/src/domain/graph/builder/cha.ts
@@ -26,8 +26,17 @@ import { RECEIVER_KINDS } from './call-resolver.js';
 export interface ChaContext {
   /** interface/class name → concrete classes that implement or extend it */
   readonly implementors: ReadonlyMap<string, readonly string[]>;
-  /** class name → direct parent class name (from `extends`) */
+  /**
+   * class name → direct parent class name (from `extends`), first-write-wins
+   * across the whole project. Ambiguous when the same bare class name is
+   * declared in multiple files with different parents — prefer
+   * `parentsByFile` whenever a declaration's home file is known.
+   */
   readonly parents: ReadonlyMap<string, string>;
+  /** `${className}|${file}` → direct parent class name (from `extends`), scoped
+   * to the file that declared `className` — disambiguates same-named classes
+   * declared in different files (issue #2062). */
+  readonly parentsByFile: ReadonlyMap<string, string>;
   /** RTA: class names that appear in `new X()` anywhere in the project */
   readonly instantiatedTypes: ReadonlySet<string>;
 }
@@ -35,6 +44,7 @@ export interface ChaContext {
 export const EMPTY_CHA_CONTEXT: ChaContext = {
   implementors: new Map(),
   parents: new Map(),
+  parentsByFile: new Map(),
   instantiatedTypes: new Set(),
 };
 
@@ -53,18 +63,22 @@ function recordImplements(cls: ClassRelation, implementors: Map<string, string[]
 }
 
 /**
- * Record a class's `extends` relationship into both the parents map (child →
- * direct parent, for this/super hierarchy walking) and the implementors map
- * (parent → children, for CHA dispatch expansion via extends).
+ * Record a class's `extends` relationship into the parents map (child →
+ * direct parent, for this/super hierarchy walking), the file-scoped parents
+ * map (same, but disambiguated by the declaring file), and the implementors
+ * map (parent → children, for CHA dispatch expansion via extends).
  */
 function recordExtends(
   cls: ClassRelation,
   implementors: Map<string, string[]>,
   parents: Map<string, string>,
+  parentsByFile: Map<string, string>,
+  file: string,
 ): void {
   if (!cls.extends) return;
   // child → parent (for this/super hierarchy walking)
   if (!parents.has(cls.name)) parents.set(cls.name, cls.extends);
+  parentsByFile.set(`${cls.name}|${file}`, cls.extends);
   // parent → children (for CHA dispatch expansion via extends)
   let list = implementors.get(cls.extends);
   if (!list) {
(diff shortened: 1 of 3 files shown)


carlos-alm:
Addressed both findings:

**Issue 1 (ambiguous parent chain traversal) — fixed.** This was real: `chaCtx.parents` is first-write-wins bare-name-keyed, so once the collision guard rejects a cross-file match and the walk takes another hop, that hop could land on an unrelated file's parent mapping for the same class name. Fixed by adding a file-scoped `parentsByFile` map (`${className}|${file}`) and using it for the initial `super()` hop and for every hop taken after rejecting a collision, wired through all three `ChaContext` construction sites (in-memory full build, DB-driven incremental rebuild, native this/super post-pass). Added a dedicated regression test (`does not misdirect the walk through a colliding class whose bare-name parent belongs to a different file`) that constructs the exact scenario — two files with same-named `Middle` classes extending different parents — and asserts the caller-file-scoped parent wins.

**Issue 2 (re-exports masquerade as declarations) — not reproducible, traced through the pipeline.** `export { X } from './other'` is captured by the extractor as an `imports` entry with `reexport: true` (`src/extractors/javascript.ts`, `handleExportCapture`/`handleExportStmt`), which becomes a `reexports` **edge**, not an `Export` pushed to `exps`. It never produces a `nodes` table row for `X` at the reexporting file. All three `byNameAndFile` implementations (`build-edges.ts`'s in-memory `nodesByNameAndFile`, `incremental.ts`'s `findNodeInFile` DB query, `native-orchestrator.ts`'s DB query) select only from the `nodes` table, filtered to declaration kinds — so a re-exported class can't appear as a phantom "declaration" in the caller's own file. If you have a concrete syntax (not `export { X } from '...'`) that does insert such a row, please point to it and I'll fix it; based on the extractor code I couldn't construct a reproduction.

Full test suite (274 files / 4454 tests) + lint + typecheck all pass.

carlos-alm:
@greptileai

PR #2292, opened 2026-08-06
fix(cycles): sort each Tarjan SCC's node array for deterministic order (#2064)

### LINKED ISSUE #2064: Flaky test: checkNoNewCycles 'excludeSpeculative has no effect on cycles formed entirely of static import edges'
## Summary

`tests/integration/check.test.ts > checkNoNewCycles > excludeSpeculative has no effect on cycles formed entirely of static import edges` is non-deterministically flaky — it fails intermittently with no code changes, comparing the same cycle's node order:

```
AssertionError: expected [ { …(2) } ] to deeply equal [ { nodes: [ …(2) ], …(1) } ]

- Expected
+ Received

  [
    {
      "nodes": [
-       "src/utils.js",
        "src/math.js",
+       "src/utils.js",
      ],
      "speculative": false,
    },
  ]
```

Reproduced locally by running the single test in a loop with zero code changes between runs — passed 2/4, failed 2/4:

```
$ for i in 1 2 3 4; do npx vitest run tests/integration/check.test.ts -t "excludeSpeculative has no effect on cycles formed entirely of static import edges"; done
# pass, pass, fail, fail
```

## Likely cause

The test asserts `withoutSpeculative.cycles` deep-equals `withSpeculative.cycles` for the exact same underlying cycle (`src/math.js` <-> `src/utils.js`). The cycle's `nodes` array order appears to depend on `Map`/`Set` iteration order inside the cycle-detection algorithm (Tarjan SCC or similar), which can differ between the two `checkNoNewCycles` invocations depending on incidental insertion-order differences introduced by the `speculative` parameter — the *set* of nodes in the cycle is correct, but the *order* isn't guaranteed to be stable, and the test uses `toEqual` (order-sensitive) rather than an order-independent comparison.

## Suggested fix

Either:
1. Make the test assertion order-independent (e.g. sort each cycle's `nodes` array before comparing), or
2. Make the underlying cycle-detection code deterministic in node ordering (e.g. sort nodes within a detected cycle before returning it), if a stable order is otherwise relied upon elsewhere.

Found while running the full test suite to validate an unrelated JS/TS extractor fix (#1929) — filed separately per this repo's scope-discipline convention since it's unrelated to that change.

### LINKED ISSUE #2067: test: checkNoNewCycles reports the same cycle with non-deterministic node order across calls

### LINKED ISSUE #2076: Tarjan SCC node ordering within a cycle isn't guaranteed stable across repeated calls
While resolving PR #1988's conflicts, `tests/integration/check.test.ts`'s \`excludeSpeculative has no effect on cycles formed entirely of static import edges\` test intermittently failed with an assertion diff showing the *same two nodes* in a *different order* between two back-to-back calls to \`checkNoNewCycles\` on identical input:

\`\`\`
- "src/utils.js",
+ "src/math.js",
  "src/math.js",
+ "src/utils.js",
\`\`\`

Reproduced deterministically-flaky (13/13 clean after a fix, but failed on 2 separate full-suite/full-file runs before that) by running \`tests/integration/check.test.ts\` standalone multiple times — same fixture DB, same edges, two consecutive \`findCycles\`-derived calls, different node order within the reported SCC.

Root cause is almost certainly that \`runTarjan\` (\`src/domain/graph/cycles.ts\`) delegates to the native \`detectCycles\` when available, and the underlying Rust implementation likely iterates a \`HashMap\`-backed adjacency/node-set whose default hasher (\`RandomState\`) reseeds per construction — so the *same logical graph* can enumerate its SCC members in a different order across separate calls within the same process, even though SCC *membership* is always correct.

This is harmless for `nodes` set *membership* (nothing currently depends on order), but any test or downstream consumer that asserts exact array equality on a \`Cycle.nodes\` array (rather than sorting first) will be flaky. Worked around locally in the one test that hit it by sorting both the cycle list and each cycle's \`nodes\` before comparing (see PR #1988, \`fix(cycles): classify cycles whose only closing edges are speculative\`).

**Suggested fix:** either (a) make the native SCC implementation iterate its internal maps in a deterministic order (e.g. a `BTreeMap`/sorted-by-insertion structure, or an explicit sort before returning), or (b) sort each returned SCC's node array by convention at the `runTarjan`/`detectCycles` boundary in both engines so callers never need to think about it. Given this repo's "both engines must produce identical results" policy, (b) at the TS/native boundary (or a corresponding native-side sort) is probably the simplest guaranteed fix.

Not blocking PR #1988 itself (worked around in the one test that surfaced it), but worth fixing at the source so future tests/consumers don't hit the same flakiness.

carlos-alm:
## Summary

`checkNoNewCycles`'s `excludeSpeculative has no effect on cycles formed
entirely of static import edges` test was flaky: two back-to-back Tarjan
runs on the exact same fixture reported the same cycle (`src/math.js` <->
`src/utils.js`) with a different node order.

Root cause: the native engine's Tarjan implementation
(`crates/codegraph-core/src/graph/algorithms/tarjan.rs`) builds its
adjacency map as a `HashMap<&str, Vec<&str>>`. Rust's default hasher
(`RandomState`) reseeds per `HashMap` construction, so the DFS entry point
selected from `graph.keys()` — and therefore the stack-pop order collected
into each returned SCC — is not stable across separate `detect_cycles`
calls on the exact same logical graph, even though SCC *membership* is
always correct.

This was previously worked around only in the one test that hit it
(sorting both sides before comparing), which papered over the symptom
without fixing the underlying non-determinism — the exact "don't document
a bug as expected behavior" anti-pattern this repo's CLAUDE.md calls out.
Three duplicate issues (#2064, #2067, #2076) independently reported the
same root cause, and #2076 explicitly names the fix that lands here: sort
each returned SCC's node array by convention at the runTarjan/detectCycles
boundary, in both engines, so callers never need to think about it.

## What changed

Sort each SCC's node array (canonical, `Vec::sort`/`Array.sort`) at the
point where it's collected — right where Tarjan closes out a component —
in all three Tarjan implementations this repo has:

- `crates/codegraph-core/src/graph/algorithms/tarjan.rs` — native
  `detect_cycles`, exposed via napi as `detectCycles`, the one consumers
  reach through `domain/graph/cycles.ts`'s `runTarjan()`.
- `src/domain/graph/cycles.ts`'s `tarjanFromEdges` — the WASM/JS fallback
  used when native isn't loaded, and mirrors of the above.
- `src/graph/algorithms/tarjan.ts`'s `tarjan()` — the `CodeGraph`-based
  implementation that the native module's own doc comment names as its
  mirror partner.

This makes each cycle's node order a deterministic function of its member
set alone, independent of adjacency-map/traversal iteration order, so
repeated calls agree byte-for-byte and both engines produce identical
output for the same input — the repo's standing "both engines must produce
identical results" invariant, applied to this primitive's own output.

Node order within a returned `Cycle.nodes` array was never a rigorously
guaranteed edge-by-edge traversal path for non-trivial (3+ node, branchy)
SCCs to begin with — Tarjan identifies SCC *membership*, not a Hamiltonian
cycle through it — so canonicalizing to alphabetical order doesn't regress
any tested or documented guarantee; it only replaces one arbitrary order
with a stable one.

## Verification

- lint: pass
- tests: pass (full suite, 274 files / 4461 tests, native engine active)
- Added regression tests asserting byte-identical node order across 25
  repeated calls, for both the native and JS/WASM paths, and cross-engine
  agreement (`tests/graph/cycles.test.ts`, `tests/graph/algorithms/tarjan.test.ts`).
- Added a Rust unit test asserting the sorted invariant directly plus a
  25-call determinism loop (`crates/codegraph-core/src/graph/algorithms/tarjan.rs`).
- Confirmed the fix empirically: called the rebuilt native `detectCycles`
  20x in a loop on the reported fixture — pre-fix this flips order
  intermittently (as reported), post-fix every call returns
  `["src/math.js","src/utils.js"]`.

Closes #2064
Closes #2067
Closes #2076

### DIFF
--- a/crates/codegraph-core/src/graph/algorithms/tarjan.rs
+++ b/crates/codegraph-core/src/graph/algorithms/tarjan.rs
@@ -81,6 +81,17 @@ fn strongconnect<'a>(
             }
         }
         if scc.len() > 1 {
+            // Canonicalize node order within the SCC before returning it. The
+            // adjacency map above is a HashMap, whose default hasher reseeds
+            // per construction — so the DFS entry point (and therefore the
+            // stack-pop order reconstructed here) is not stable across
+            // separate calls on the exact same logical graph, even though SCC
+            // *membership* always is. Sorting makes the returned node array a
+            // deterministic function of the SCC's member set alone, matching
+            // the JS/WASM fallback (see cycles.ts's tarjanFromEdges and
+            // graph/algorithms/tarjan.ts) so both engines agree byte-for-byte.
+            // See issues #2064, #2067, #2076.
+            scc.sort();
             state.sccs.push(scc);
         }
     }
@@ -166,5 +177,59 @@ mod tests {
         ];
         let cycles = detect_cycles(&edges);
         assert_eq!(cycles.len(), 2);
+    }
+
+    /// Regression test for issues #2064, #2067, #2076: the adjacency map is a
+    /// `HashMap`, whose default hasher reseeds per construction, so the DFS
+    /// entry point (and therefore the pre-sort stack-pop order) is not stable
+    /// across separate calls on the exact same logical graph. Each SCC's node
+    /// array must come back sorted regardless, so repeated calls agree
+    /// byte-for-byte and match the JS/WASM fallback's output.
+    #[test]
+    fn test_node_order_is_deterministic_across_repeated_calls() {
+        let edges = vec![
+            GraphEdge {
+                source: "src/math.js".to_string(),
+                target: "src/utils.js".to_string(),
+            },
+            GraphEdge {
+                source: "src/utils.js".to_string(),
+                target: "src/math.js".to_string(),
+            },
+        ];
+        let first = detect_cycles(&edges);
+        for _ in 0..25 {
+            assert_eq!(detect_cycles(&edges), first);
+        }
+        assert_eq!(
+            first,
+            vec![vec!["src/math.js".to_string(), "src/utils.js".to_string()]]
+        );
+    }
+
+    #[test]
+    fn test_scc_nodes_are_returned_sorted() {
+        let edges = vec![
+            GraphEdge {
+                source: "c".to_string(),
+                target: "a".to_string(),
+            },
+            GraphEdge {
+                source: "a".to_string(),
+                target: "b".to_string(),
+            },
+            GraphEdge {
+                source: "b".to_string(),
+                target: "c".to_string(),
+            },
+        ];
+        let cycles = detect_cycles(&edges);
+        assert_eq!(cycles.len(), 1);
+        let mut sorted = cycles[0].clone();
+        sorted.sort();
+        assert_eq!(
+            cycles[0], sorted,
+            "SCC node array must already be in sorted order"
+        );
     }
 }
--- a/src/domain/graph/cycles.ts
+++ b/src/domain/graph/cycles.ts
@@ -1,6 +1,7 @@
 import { getCallableNodes, getCallEdges, getFileNodesAll, getImportEdges } from '../../db/index.js';
 import { loadNative } from '../../infrastructure/native.js';
 import { isTestFile } from '../../infrastructure/test-filter.js';
+import { compareByCodePoint } from '../../shared/compare.js';
 import type { BetterSqlite3Database } from '../../types.js';
 
 type Edge = { source: string; target: string; speculative?: boolean };
@@ -227,7 +228,9 @@ function tarjanFromEdges(edges: Edge[]): string[][] {
         // but sorting keeps it byte-identical to the native engine's output
         // for the same logical graph regardless of either side's internal
         // traversal order. See issues #2064, #2067, #2076.
-        scc.sort();
+        // compareByCodePoint (not the default comparator) so supplementary-
+        // plane Unicode labels sort the same way Rust's UTF-8 byte order does.
+        scc.sort(compareByCodePoint);
         sccs.push(scc);
       }
     }
--- a/src/domain/graph/cycles.ts
+++ b/src/domain/graph/cycles.ts
@@ -219,7 +219,17 @@ function tarjanFromEdges(edges: Edge[]): string[][] {
         onStack.delete(w);
         scc.push(w);
       } while (w !== v);
-      if (scc.length > 1) sccs.push(scc);
+      if (scc.length > 1) {
+        // Canonicalize node order within the SCC before returning it — see
+        // the matching comment in crates/codegraph-core/src/graph/algorithms/
+        // tarjan.rs. `allNodes` here is a Set built from edge insertion order,
+        // so this JS path is already deterministic call-to-call in isolation,
+        // but sorting keeps it byte-identical to the native engine's output
+        // for the same logical graph regardless of either side's internal
+        // traversal order. See issues #2064, #2067, #2076.
+        scc.sort();
+        sccs.push(scc);
+      }
     }
   }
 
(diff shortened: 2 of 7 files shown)


carlos-alm:
Fixed the Unicode ordering divergence: added `compareByCodePoint` (`src/shared/compare.ts`) and wired it into both JS-side SCC-closing sorts (`src/graph/algorithms/tarjan.ts`, `src/domain/graph/cycles.ts`'s `tarjanFromEdges`). Rust's default `String`/`str` sort compares UTF-8 bytes, which already agrees with code-point order for the full Unicode range — no Rust change was needed there.

Added regression coverage:
- `tests/unit/compare.test.ts` — unit tests for the comparator itself, including the exact U+FFFF vs U+1F600 case where the default comparator and code-point order disagree.
- `tests/graph/algorithms/tarjan.test.ts` and `tests/graph/cycles.test.ts` — Unicode-labeled cycle tests verifying the actual SCC output now sorts correctly, including a native-vs-JS parity test that compares raw (non-re-sorted) output so the existing `sortCycles` test helper's own re-sort can't mask the divergence.

Full suite (275 files / 4467 tests) + lint + typecheck all pass.

carlos-alm:
@greptileai

Issue #2310, opened 2026-08-07
gc-worktrees.sh's --prune-branches uses git branch -d, which refuses genuinely squash-merged branches


carlos-alm:
Follow-up from #2309 (same root cause, third occurrence, found while fixing the other two).

\`.claude/scripts/gc-worktrees.sh\` correctly classifies a worktree as \"stale\" via GitHub PR state (\`gh pr list --state all --json headRefName,state\`, not git ancestry) — see the script's own \`classify()\` function and its documented rationale for why ancestry is unreliable under squash merges.

But the actual branch-deletion step (\`--prune-branches\`) still uses plain \`git branch -d\`:

\`\`\`bash
# line ~216
if git -C "$main_root" branch -d "$b" >/dev/null 2>&1; then echo "  branch deleted   $b"; fi
\`\`\`

\`-d\`'s own internal check is ancestry-based, independent of the \`gh\`-based classification that selected the branch as safe to delete. Verified directly (see #2309): once stale remote-tracking refs are pruned, \`git branch -d\` refuses a branch that GitHub confirms is genuinely merged via squash, because the local branch tip is never an ancestor of the squash commit on the default branch. Since \`classify()\` already ran the authoritative check, this refusal is silent — the script just doesn't print \"branch deleted\" for that branch and moves on, so the branch quietly survives every future \`--prune-branches\` run indefinitely.

**Suggested fix:** use \`git branch -D\` for branches \`classify()\` already confirmed merged via \`gh\` PR state — exactly the same fix applied to \`housekeep\`'s Phase 4d and Phase 1e in #2309's PR, with the same justification: this isn't a blanket force-delete, it's using the authoritative GitHub signal instead of git's unreliable-under-squash-merge ancestry heuristic.

carlos-alm:
Related finding from Greptile's review of PR #2311 (the housekeep/fixer fix this issue is a follow-up to) — worth fixing together with the `-d`-vs-`-D` issue above, since both live in the same `classify()` function / call site:

\`classify()\` matches a worktree's branch to \"settled\" purely by **name** against \`gh pr list\`'s \`headRefName\` history — it never checks whether the branch's *current* commit is the one that actually got merged. If a branch name is reused after its earlier PR merged (a later run, or a human, creating a new branch of the same name for unrelated work), \`classify()\` would still return "settled" for the new, never-merged commits, and `--prune-branches`'s `git branch -d`/`-D` would delete them.

[...]

Issue #2077, opened 2026-07-10
fix(incremental): thread configured pointsToMaxIterations through watch-mode points-to map


carlos-alm:
Deferred from PR #1997 review.

**Original reviewer comment:** https://github.com/optave/ops-codegraph-tool/pull/1997#discussion_r3540831424

**Context:** `buildCallEdges` in `src/domain/graph/builder/incremental.ts` calls `buildPointsToMapForFile(symbols, importedNames)` with only 2 arguments, so the points-to solver's fixed-point iteration cap falls back to `buildPointsToMapForFile`'s default parameter (`DEFAULTS.analysis.pointsToMaxIterations`) instead of the resolved `ctx.config.analysis.pointsToMaxIterations` that the full-build path (`build-edges.ts`) passes explicitly.

For most users (no `.codegraphrc.json` override) this is a no-op — the default parameter equals the default config value. It only diverges for projects that explicitly override `pointsToMaxIterations`, where `codegraph watch`'s incremental rebuild would silently use a different iteration cap than a full build of the same files, potentially producing different alias-resolved call edges for long points-to chains.

Fixing this properly requires threading the resolved config value through the whole watch-mode call chain: `EngineOpts` (src/types.ts) currently carries no config reference at all, so this needs either a new field on `EngineOpts` or a separate parameter threaded through `rebuildFile` → `buildCallEdges`, plus finding and updating whichever caller (watcher.ts or the `codegraph watch` CLI wiring) currently constructs `EngineOpts` without it. That's a public-interface change spanning files outside PR #1997's diff, so it's tracked here rather than fixed inline.

PR #2311, opened 2026-08-07
fix(skills): housekeep branch pruning ignores squash merges; fixer never re-runs it after draining

### LINKED ISSUE #2309: housekeep/fixer branch cleanup is broken: squash-merge false negative + no post-drain sweep
Discovered while running \`/fixer\` end-to-end (issues #1923-#2077, this run merged 18 PRs via squash).

## Bug 1: housekeep Phase 4a's merged-branch detection is wrong for squash merges

\`housekeep\`'s Phase 1c (\"Identify stale worktrees\") already documents and fixes this exact problem:

> **Do not hand-roll this check with \`git branch --merged origin/main\`.** In a repo that **squash-merges**, a landed branch's tip is never an ancestor of the default branch, so \`--merged\` lists almost nothing — the check looks right, passes review, and silently reports every finished worktree as still in flight. The script resolves merged-ness from GitHub PR state for exactly this reason.

But Phase 4a (\"Prune Merged Branches\") still uses the naive, documented-as-wrong command:

\`\`\`bash
git branch --merged origin/main
\`\`\`

Verified directly: after this \`/fixer\` run squash-merged 12 PRs, \`git branch --merged origin/main\` returned **empty** in the run's own worktree, even though 10 of those PRs' branches (\`fix/issue-2053-rust-trait-bodyless-complexity\`, \`fix/issue-2070\` through \`fix/issue-2077\`, etc.) still existed locally and were genuinely merged (confirmed via \`state.json\` + \`gh pr view\`). The check silently reported \"nothing to clean\" both times \`/fixer\` invoked \`/housekeep\` this run.

**Fix:** Phase 4a should resolve merged-ness the same way \`gc-worktrees.sh\` (used by Phase 1c) already does — via \`gh pr list --state all --json headRefName,state\`, matching MERGED/CLOSED PRs to local branch names — not \`git branch --merged\`.

## Bug 2: /fixer never calls /housekeep after Phase 4 (Drain Parked PRs)

\`/fixer\`'s Phase 3 (\"Continue the Batch Loop, or Proceed to Drain\") step 3a calls \`/housekeep\` once per completed batch specifically because branches accumulate during the Solve-and-Merge loop. But Phase 4 (Drain Parked PRs) also cuts/merges branches (the catch-up-merge procedure checks out and merges each parked PR), and nothing calls \`/housekeep\` again after the drain loop finishes.

In this run, issue #2306's parked PR (#2307) was merged during Phase 4, after the last Phase-3 \`/housekeep\` call — so its branch was never swept by the run's own cleanup step.

**Fix:** Phase 4's exit condition (or Phase 5 entry) should include one more \`/housekeep\` invocation after the drain loop completes, mirroring Phase 3's step 3a.

## Impact

Combined, these two bugs mean \`/fixer\`'s "run housekeep once per batch" design has been silently doing nothing useful for branch pruning specifically (worktree/dirt-file cleanup in the same runs was unaffected) for as long as bug 1 has existed, and even a correct detector would still miss drain-phase branches until bug 2 is fixed too.

carlos-alm:
## Summary

Two related bugs found while running \`/fixer\` end-to-end this session (issues #1923-#2077, 12 squash-merged PRs):

**1. `housekeep`'s Phase 4a ("Find merged branches") used `git branch --merged origin/main`** — ancestry-based, so a squash-merged branch's tip is never an ancestor of the new squash commit on \`main\`. Phase 1c already documents this exact failure mode and fixes it for worktree staleness (resolves merged-ness from GitHub PR state instead); Phase 4a never got the same fix.

Verified directly: after this run squash-merged 12 PRs, the ancestry check reported zero mergeable branches both times \`/fixer\` invoked \`/housekeep\`, while 10 of those branches sat locally unpruned.

Fixing detection alone isn't sufficient: \`git branch -d\`'s own internal check is the *same* ancestry test, so even a correctly \`gh pr list\`-confirmed-merged branch still gets refused by \`-d\` once stale remote-tracking refs are pruned (verified directly — reproduced this exact refusal on a real squash-merged branch from this run). Phase 4d and Phase 1e's worktree-branch cleanup now use \`-D\` for branches a GitHub PR-state check already confirmed merged — not a blanket force-delete, just deferring to a more authoritative signal than git's own unreliable-under-squash-merge heuristic.

**2. `fixer`'s Phase 3 (step 3a) runs `/housekeep` once per completed batch** specifically because branches accumulate during the Solve and Merge Loop — but Phase 4 (Drain Parked PRs) also cuts/merges branches via its catch-up-merge procedure, and nothing called `/housekeep` again after the drain loop's own exit conditions (all merged, 3-stall, or 15-pass cap). Added the same housekeep call, once, after the drain loop stops.

## Follow-up filed

#2310 — \`gc-worktrees.sh\` (a separate executable script, not a skill doc) has the identical \`-d\`-vs-ancestry bug at its own branch-deletion call site. Left for a separate PR since it's a script + its own test suite, not a doc change.

## Verification

- \`npm run lint\` clean
- \`tests/unit/docs-skills-mirror-sync.test.ts\` + \`tests/unit/docs-check-flags.test.ts\` pass (22/22) — confirms these skill docs aren't part of the public docs mirror set, so no mirror update needed
- Manually reproduced both halves of the bug on real branches from this run: \`git branch --merged origin/main\` returned empty against 10 genuinely-squash-merged branches; \`git branch -d\` on one of them refused with "not fully merged" once its stale remote-tracking ref was pruned

Closes #2309

### DIFF
--- a/.claude/skills/fixer/SKILL.md
+++ b/.claude/skills/fixer/SKILL.md
@@ -1338,7 +1338,19 @@ A reported line is not automatically a bug — `main` may have legitimately supe
 
 After the integrity check passes, re-run local verification and push, then re-evaluate the five gate conditions from Phase: Solve and Merge Loop before merging.
 
-**Exit condition:** Every PR listed in `parked.txt` is merged, or is reported as needing human review with a specific reason. No catch-up merge was pushed without a passing I4 integrity check. Every drain-phase merge's post-merge CI on `main` was confirmed green (I8) before the next merge in the pass. Drain stopped only once a pass merged nothing 3 times in a row, or the 15-pass safety cap was hit — never on a fixed pass count while merges were still happening.
+### Housekeep after the drain loop exits
+
+Every catch-up merge above cuts/checks-out a branch the same way Phase 2's Solve and Merge Loop does, so it leaves exactly the same kind of local-branch residue that Phase 3's step 3a runs `/housekeep` to clean up — but the drain loop has its own exit points (all merged, 3-stall, or 15-pass cap) separate from the batch loop, so nothing calls `/housekeep` again after it. Run it once, exactly like 3a, after the drain loop above has actually stopped (do not call it once per pass — it is idempotent, but there is nothing new to clean between passes that haven't yet merged anything):
+
+```bash
+mkdir -p .codegraph/fixer
+DRY_RUN=$(cat .codegraph/fixer/dry-run)
+echo "fixer: drain loop finished — running housekeeping before the final report"
+```
+
+Invoke the `housekeep` skill: `/housekeep --dry-run` if `DRY_RUN` is `true`, otherwise plain `/housekeep`. The same confirmation posture from 3a applies unchanged: confirm deletion only for local branches this run itself recorded as `merged` in `state.json` (drain-phase merges included — they are recorded exactly the same way as Phase 2 merges), decline every worktree-removal prompt including this run's own worktree, and decline everything else.
+
+**Exit condition:** Every PR listed in `parked.txt` is merged, or is reported as needing human review with a specific reason. No catch-up merge was pushed without a passing I4 integrity check. Every drain-phase merge's post-merge CI on `main` was confirmed green (I8) before the next merge in the pass. Drain stopped only once a pass merged nothing 3 times in a row, or the 15-pass safety cap was hit — never on a fixed pass count while merges were still happening. `/housekeep` has run once after the drain loop stopped, mirroring 3a.
 
 ---
 
--- a/.claude/skills/housekeep/SKILL.md
+++ b/.claude/skills/housekeep/SKILL.md
@@ -165,8 +165,9 @@ git worktree prune
 - If confirmed:
   ```bash
   git worktree remove "<path>"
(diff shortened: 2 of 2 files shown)


carlos-alm:
Fixed — the branch-name-reuse gap you flagged was real. Phase 4a now pulls `headRefOid` alongside `headRefName` from `gh pr list` and only classifies a local branch as confirmed-merged when its current tip SHA exactly matches the SHA GitHub recorded as that PR's head at merge time (previously it matched on `headRefName` alone). Any commit added since — a reused name, a new unpushed commit — changes the tip SHA, so the pair no longer matches and the branch is correctly left alone instead of guessed at.

Verified the new `comm -12` name+SHA intersection manually against this repo's real branch/PR history — it correctly returns exactly the genuinely-merged branches.

Also noted the same gap exists in Phase 1c's `gc-worktrees.sh`-backed classifier (which matches by name only too) and added it to #2310 (already tracking that script's separate `-d`-vs-`-D` bug) rather than fixing it here, since the actual query logic for that path lives in the external script, not this skill doc.

carlos-alm:
@greptileai

PR #2314, opened 2026-08-07
fix(incremental): revisit sibling-implementor callers in CHA post-pass (#2078)

### LINKED ISSUE #2078: fix(incremental): CHA dispatch post-pass doesn't revisit unrelated caller files after implementor add/remove
Deferred from PR #1997 review.

**Original reviewer comment:** https://github.com/optave/ops-codegraph-tool/pull/1997#discussion_r3540831470

**Context:** `applyChaDispatchPostPass` in `src/domain/graph/builder/incremental.ts` rebuilds `ChaContext` fresh from the DB (so the class hierarchy and RTA instantiation evidence are always globally current), but only re-evaluates call sites in `filesWithSymbols` — the rebuilt target file plus its direct reverse-dependency cascade.

If a watch-mode rebuild adds or removes a class that implements/extends an interface with existing dispatch call sites elsewhere, callers in files that don't directly import the changed file (i.e. they call through the interface type, not a direct reference) are not revisited. Those files can be left with missing or stale `caller -> implementor.method` edges until they themselves are rebuilt or a full rebuild runs.

A correct fix would need to identify every file with an existing CHA/RTA-dispatch call site whose target interface's implementor set the current rebuild changed — likely a DB query over `calls` edges with `technique IN ('cha', 'super-dispatch')` scoped to the affected interface(s) — and re-process those files' calls too. That's a materially different (and more expensive) post-pass design than the bounded-blast-radius one this PR ships, and deserves its own design discussion rather than a rushed inline patch, so it's tracked here as a follow-up.

carlos-alm:
## Summary

Fixes the gap deferred from PR #1997's review: `applyChaDispatchPostPass` (the incremental single-file rebuild's CHA/RTA dispatch post-pass, `src/domain/graph/builder/incremental.ts`) only re-evaluated call sites in the rebuilt file plus its reverse-dep cascade (`findReverseDeps` — files that already hold an edge to the rebuilt file's OLD nodes).

That misses a caller that dispatches through an **interface type**, not a direct reference: such a caller has no edge to a brand-new implementor added elsewhere in the same rebuild, so it was never revisited and the caller → new-implementor dispatch edge stayed silently missing until a full rebuild ran.

**Fix:** `findChaSiblingCallerFiles` — for every interface/base class a rebuilt file's classes just gained or changed a heritage (`extends`/`implements`) relationship to, find existing callers with a `technique IN ('cha', 'super-dispatch')` edge to any *other* known implementor of that same interface. Such an edge could only exist because `resolveChaTargets`'s BFS started from that interface name, so its source file is exactly the caller set this rebuild's implementor-set change needs to reach. Those files are re-parsed (mirroring the existing `parseReverseDep` helper) and folded into the CHA post-pass's file set — no other part of their graph state is touched.

**The "implementor removed" half of this issue's title was already handled**, traced through `purgeFileData`: it deletes every edge pointing at a rebuilt file's OLD nodes (both directions, via a `target_id IN (SELECT id FROM nodes WHERE file = @f)` clause) *before* the file's nodes are re-inserted — regardless of which caller holds the edge. So a stale dispatch edge to a since-removed implementor is always cleaned up when that implementor's own file is rebuilt; only the "gained" half needed new code.

## Verification
- lint: pass
- typecheck: pass
- tests: pass (279 files / 4503 tests, including 10 new tests in `tests/integration/issue-2078-cha-sibling-caller.test.ts`, covering both WASM and native engines)
- Confirmed the new test actually catches the bug: reverted the fix locally, re-ran the test suite — 4 of the 10 new tests failed with the missing edge, confirming they exercise the real gap; restored the fix and all 10 pass again.
- No native Rust changes needed: `rebuildFile`/`applyChaDispatchPostPass` is the single shared incremental watch-mode rebuild path for both engines (parser engine choice is a parameter, not a separate pipeline) — the new regression test explicitly covers both `engine: 'wasm'` and `engine: 'native'` and both pass.

Closes #2078

### DIFF
--- a/src/domain/graph/builder/incremental.ts
+++ b/src/domain/graph/builder/incremental.ts
@@ -10,6 +10,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import { bulkNodeIdsByFile, purgeFileData } from '../../../db/index.js';
+import { escapeLike } from '../../../db/query-builder.js';
 import { PROPAGATION_HOP_PENALTY } from '../../../extractors/javascript.js';
 import { debug, warn } from '../../../infrastructure/logger.js';
 import { getOrCreatePerDbChunkStmt } from '../../../shared/chunked-stmt-cache.js';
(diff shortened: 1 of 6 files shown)


carlos-alm:
Addressed both issues from the review:

**Issue 1 (transitive interface ancestry)** — fixed in ace44486. Added \`buildReverseHeritageMap\` (a child -> parents reverse lookup covering both \`extends\` and \`implements\`, since \`ChaContext.parents\` only tracks \`extends\`) so \`findChaSiblingCallerFiles\` walks upward from a rebuilt class's direct heritage to every ancestor interface, then downward from every touched name via \`chaCtx.implementors\` to every transitively reachable implementor — mirroring \`resolveChaTargets\`'s own BFS instead of stopping at direct children/parents. Added a new regression test (\`Square extends AbstractShape\`, \`AbstractShape implements IShape\`, existing caller dispatches via \`IShape\` to an unrelated sibling \`Circle\`) that fails without this fix and passes with it.

**Issue 2 (empty target sets / first implementor)** — this is a real gap, but it's a fundamentally different problem from Issue 1: there's no existing edge to seed discovery from at all, by construction. Closing it needs either a schema change (persist unresolved-but-typed call sites) or an O(all files) re-scan — exactly the "materially different (and more expensive) post-pass design" the original #2078 report explicitly deferred as needing its own design discussion, not a rushed inline patch. Filed as #2315 with the tradeoffs spelled out rather than expanding this PR's scope.

@greptileai please re-review.

PR #2317, opened 2026-08-07
fix(extractors): seed object-rest-param type annotations for CHA dispatch (#2080)

### LINKED ISSUE #2080: investigate: object-rest-param bindings never get a type-annotation-based typeMap seed
Deferred from PR #1997 review (follow-up to a Greptile finding fixed in that PR).

**Context:** Greptile flagged that `applyChaDispatchPostPass` (src/domain/graph/builder/incremental.ts) used `coerceTypeMap` instead of `buildIncrementalTypeMap` (which also runs `seedRestParamTypeMap`), an inconsistency now fixed in PR #1997.

While verifying that fix, traced through `seedRestParamTypeMap`'s actual mechanism and `collectParamBindings`'s extraction (src/extractors/javascript.ts:2915) and found the seeded value is always the call-site argument's *variable name* (`paramBindings[].argName`, populated only from a plain `identifier` argument node), not a resolved class/interface type name. `emitChaDispatchForCall` (incremental.ts) does a single `typeMap.get(call.receiver)` lookup and passes `.type` directly to `resolveChaTargets`, with no second hop to chase a variable name to its actual type.

Net effect: for a rest-destructured receiver typed against an interface (e.g. `function f({ ...rest }: IWorker) { rest.doWork() }`), CHA-style interface dispatch through `rest` likely still won't resolve even after the #1997 fix, because the rest-param path was never designed to feed a type name into that lookup — it exists for the Phase 8.3f value-chase (#1336: `eerest.e4()` resolving through a `var obj = { e4 }`-style chain), a different mechanism than typed-receiver CHA dispatch.

Worth investigating whether the extractor should seed a genuine type-annotation-based typeMap entry for object-rest-param bindings (mirroring how a plain typed parameter like `function dispatch(worker: IWorker)` gets `typeMap['worker'] = { type: 'IWorker' }` directly), so CHA/interface dispatch through a typed rest receiver actually works — or whether this is intentionally out of scope for the object-rest-param feature.

carlos-alm:
## Summary

Investigating this issue (which asks whether object-rest-param bindings should get a type-annotation-based typeMap seed for CHA/interface dispatch) revealed two related bugs in `collectObjectRestParams` (TS, `src/extractors/javascript.ts`) / `collect_object_rest_params` (native, `crates/codegraph-core/src/extractors/javascript.rs`). Both are fixed together, since the second makes the first meaningful:

1. **Neither ever unwrapped TypeScript's `required_parameter`/`optional_parameter` node.** tree-sitter-typescript wraps *every* parameter — typed or not — in that node (confirmed by parsing `function f({ ...rest }) {}` directly with the TS grammar), unlike plain tree-sitter-javascript, where `object_pattern` is a direct `formal_parameters` child. Object-rest-param bindings were therefore silently never recorded for **any** `.ts`/`.tsx` file at all, independent of type annotations. The existing #1336 test only ever exercised the `.js` grammar, so this went undetected. Fixed by unwrapping via the `pattern` field before checking for `object_pattern`.
2. **A type-annotated rest parameter (`{ ...rest }: IWorker`) never got a type-annotation-based typeMap entry**, unlike a plain typed parameter (`worker: IWorker`) — `handleParamTypeMap`'s early return on `nameNode.type !== 'identifier'` skipped `object_pattern` entirely. Extended it to seed the rest binding's own name directly from the annotation when present, so CHA/interface dispatch through a typed rest receiver resolves exactly like a plain typed parameter already does. This new direct seed (confidence 0.9) coexists correctly with the pre-existing value-chase seed (`seedRestParamTypeMap`/`buildObjectRestParamPostPass`, confidence 0.65) via `setTypeMapEntry`'s higher-confidence-wins merge — an explicit type annotation is strictly more reliable than an inferred call-site argument name.

Both fixes are necessary together: this issue specifically asks for type-annotation-based seeding, which only exists in TypeScript, but the TS extraction path for rest-param bindings was completely non-functional without fix #1 regardless of annotations — the requested feature couldn't be delivered without it.

## Verification
- lint: pass
- typecheck: pass
- Rust: `cargo test` — 799 passed (up from 795, 4 new tests), including new coverage for both the wrapping-unwrap fix and the type-annotation typeMap seed
- TS: 281 files / 4522 tests, including 6 new integration tests (`tests/integration/issue-2080-object-rest-param-typed-dispatch.test.ts`) and 5 new unit tests (`tests/parsers/javascript.test.ts`)
- Resolution benchmark suite: 206/206 pass — no precision/recall regression on any existing fixture
- Engine parity verified end-to-end: rebuilt the native addon locally and confirmed native and WASM produce byte-identical edges for both the untyped-parity scenario and the typed-CHA-dispatch scenario

Closes #2080

### DIFF
--- a/crates/codegraph-core/src/extractors/javascript.rs
+++ b/crates/codegraph-core/src/extractors/javascript.rs
@@ -258,6 +258,20 @@ fn handle_param_type_map(node: &Node, source: &[u8], symbols: &mut FileSymbols) 
         return;
     }
     if name_node.kind() != "object_pattern" { return };
+    // Only seed when the rest element is the pattern's ONLY member (`{
+    // ...rest }: IWorker`) — if a named property sits alongside it (`{
+    // doWork, ...rest }: IWorker`), TypeScript's own structural typing
+    // excludes that property from `rest`'s real type (effectively
+    // `Omit<IWorker, 'doWork'>`), so assigning the full `IWorker` type to
+    // `rest` would let a call like `rest.doWork()` — invalid, since
+    // `doWork` was destructured away — resolve a false edge via CHA
+    // dispatch (#2080 review).
+    for i in 0..name_node.child_count() {
+        let Some(sibling) = name_node.child(i) else { continue };
+        let st = sibling.kind();
+        if st == "{" || st == "}" || st == "," { continue }
+        if st != "rest_pattern" && st != "rest_element" { return }
+    }
     let Some(type_anno) = find_child(node, "type_annotation") else { return };
     let Some(type_name) = extract_simple_type_name(&type_anno, source) else { return };
     for i in 0..name_node.child_count() {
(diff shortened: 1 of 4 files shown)


carlos-alm:
Fixed in ca5c1f11. The seed now only fires when the rest element is the pattern's sole member (`{ ...rest }: IWorker`, no sibling properties) — when a named property sits alongside it (`{ doWork, ...rest }: IWorker`), the seed is skipped entirely, since TypeScript's own structural typing would exclude that property from `rest`'s real type and seeding the full annotation there would produce exactly the false CHA edge you flagged. `object_rest_param_bindings` extraction itself (the #1336 value-chase mechanism) is unaffected — it already recorded the rest binding regardless of sibling properties, and still does.

Added regression tests confirming: (1) the false edge no longer resolves in either engine for `{ doWork, ...rest }: IWorker; rest.doWork()`, and (2) the primary `{ ...rest }: IWorker` case from this PR's original scope still works.

@greptileai please re-review.

Issue #2319, opened 2026-08-07
Dart: no typeMap population for field/parameter types — receiver-typed calls unresolved (#2082 follow-up)


carlos-alm:
## Context

Follow-up to #2082, discovered while verifying the resolution-benchmark improvement from that issue's fixes.

After fixing #2082's three bugs (bare-call extraction, endLine truncation, bodyless-constructor extraction — plus an incidentally-discovered `.id`-vs-`===` node-identity bug in the selector-resolution fallback), Dart's resolution benchmark jumped from a genuine 0% floor to **100% precision, 61.9% recall (13/21 expected edges)**. All remaining false negatives are in the **receiver-typed** mode:

```
[dart] static: 4/4 (100.0% recall)
[dart] receiver-typed: 0/8 (0.0% recall)
[dart] same-file: 2/2 (100.0% recall)
```

Example from `tests/benchmarks/resolution/fixtures/dart/service.dart`:

```dart
class UserService {
  final UserRepository _repo;

  UserService(this._repo);

  User? getUser(String id) {
    return _repo.findById(id);   // expected edge: UserService.getUser -> UserRepository.findById
  }
}
```

`_repo.findById(id)` should resolve to `UserRepository.findById` because `_repo`'s type is `UserRepository` — established via the constructor's `this._repo` shorthand parameter (a field-initializing parameter, Dart's equivalent of a typed constructor param). But `src/extractors/dart.ts` never populates `ctx.typeMap` for field or parameter types at all (confirmed: `extractDartSymbols` initializes `typeMap: new Map()` and nothing in the file ever calls `.set()` on it). Without a typeMap entry for `_repo`, the call resolver has no way to know its receiver type and can't dispatch `.findById()` to `UserRepository`.

## Suggested fix direction

Mirror how other languages seed typeMap entries for constructor-initialized fields (e.g. TypeScript's `this.field = new Ctor()` → class-scoped `ClassName.field` key, see `handlePropWriteTypeMap`/`handleFieldDefTypeMap` in `src/extractors/javascript.ts`) and for `this.`-shorthand constructor parameters specifically:

1. **Field declarations with an explicit type** (`final UserRepository _repo;`): seed `typeMap['ClassName._repo']` (or an equivalent class-scoped key) from the declared type.
[...]

PR #2320, opened 2026-08-07
fix(dart): extract bare calls, fix endLine truncation, bodyless members (#2082)

### LINKED ISSUE #2082: Dart extractor: bare keyword-less calls not extracted; multi-line function endLine truncated; bodyless constructors not extracted
Discovered while adding Dart integration-test coverage to PR #2028 (constructor-call attribution, #1892).

**Context:** PR #2028 adds `new ClassName()` → constructor-method attribution. Java, PHP, and Groovy fixtures using `new ClassName()` verify this correctly. Attempting the analogous Dart fixture surfaced two separate, pre-existing gaps in `src/extractors/dart.ts`, unrelated to constructor attribution itself:

**1. Bare (keyword-less) calls are not extracted at all.**

Modern Dart permits omitting `new` for both plain function calls and constructor invocations (`var x = Foo();`, `helper();`). Neither is extracted as a `Call` by `extractDartSymbols` — confirmed empirically:

```ts
// helper(); inside main() → symbols.calls === []
// var w = Foo(); (no `new`) → symbols.calls === []
// var w = new Foo(); (explicit `new`) → symbols.calls === [{ name: 'Foo', line: N }]  ✓
```

Only the explicit `new_expression` node type is dispatched to `handleDartConstructorCall` in `walkDartNode`'s switch (`src/extractors/dart.ts`). `constructor_invocation` is also mapped to the same handler, but apparently doesn't fire for a bare capitalized call the way the code comments assume — and there is no handler at all for a bare *plain* (non-capitalized) function call.

This means the existing hand-annotated fixtures in `tests/benchmarks/resolution/fixtures/dart/*.dart` (which use the idiomatic keyword-less style throughout, e.g. `var repo = UserRepository();` in `main.dart`) do not actually resolve any of their "constructor" or "static" mode expected edges today. This is consistent with (and likely explains) the `dart: { precision: 0.0, recall: 0.0 }` floor in `tests/benchmarks/resolution/resolution-benchmark.test.ts` — the threshold is a placeholder because current recall is genuinely ~0%, not because the floor merely hasn't been ratcheted up yet.

**2. Multi-line block-bodied function/method definitions get `endLine` truncated to the signature line, not the body's closing brace.**

```ts
// Waldo makeWaldo() {\n  return new Waldo();\n}
// → definition: { name: 'makeWaldo', kind: 'function', line: 5, endLine: 5 }
// → call: { name: 'Waldo', line: 6 }   (line 6 > endLine 5!)
```

Because the call's line falls outside the recorded `[line, endLine]` span, downstream caller-attribution (`findEnclosingBinding`-style enclosing-function lookup during graph build) cannot attribute the call to `makeWaldo` — it falls back to file-level attribution (`src: 'repro.dart', srcKind: 'file'`) instead. Single-line arrow-bodied functions (`Waldo makeWaldo() => new Waldo();`) do not hit this, since the whole function (signature + body) sits on one line.

This affects essentially any real-world multi-line Dart function/method (block bodies are the common case — e.g. `UserService.createUser`/`getUser`/`removeUser` in the resolution-benchmark fixture all use block bodies), so it is likely a second, independent contributor to Dart's current ~0% resolution recall, on top of gap #1.

**Suggested fix directions (not investigated in depth — flagging for triage):**
- For #1: determine what `constructor_invocation` actually covers in the installed tree-sitter-dart grammar version, and add a real dispatch path for bare identifier-call syntax (likely needs a new node-type case in `walkDartNode`, mirroring how other languages disambiguate a capitalized bare call as a constructor vs. a plain function call).
- For #2: `handleDartClass`'s `extractDartClassMembers` (for methods) and `handleDartFunction`/`handleDartMethodSig` (for top-level functions) currently call `nodeEndLine(member)`/`nodeEndLine(node)` on the `method_signature`/`function_signature` node itself; needs to instead measure through to the enclosing declaration/body node so `endLine` covers the full definition, not just the signature.

**Impact:** Both are scoped entirely to `src/extractors/dart.ts` (WASM) — have not checked whether `crates/codegraph-core`'s Rust Dart extractor (if one exists) has analogous or different behavior; that should be checked as part of any fix for engine parity.

carlos-alm:
## Summary

Three pre-existing Dart extractor bugs, discovered together while adding Dart constructor-call attribution test coverage. Fixed in both engines (WASM `src/extractors/dart.ts` and native `crates/codegraph-core/src/extractors/dart.rs`):

1. **Bare (keyword-less) calls never extracted.** Modern Dart permits omitting `new`/parens-only calls (`helper()`, `var w = Foo()`). WASM's `resolveDartSelectorMethodName` only handled a preceding `.method` selector (chained access); native had **no dispatch case for `call_expression` at all** — tree-sitter-dart 0.2 represents every call (bare or chained) as `call_expression`, a completely different shape from the older `selector`-based grammar the existing code targeted. Every native Dart call was silently unextracted, not just bare ones.
2. **Multi-line function/method `endLine` truncated to the signature line.** `function_signature`/`method_signature` and `function_body` are SIBLING nodes in tree-sitter-dart (confirmed in both grammar versions), not parent-child. Any call inside a multi-line body fell outside `[line, endLine]`, so enclosing-function caller-attribution during graph build silently misattributed it to file scope.
3. **Bodyless (semicolon-only) constructors and abstract methods** (`Foo();`, `double area();`) — the idiomatic short form Dart codebases use throughout — were never extracted as definitions. They wrap their signature in a `declaration` node instead of the `method_signature`/`method_declaration` a block-bodied member uses, which the field-detection fallback silently swallowed.

**Also fixed, found while implementing #1:** a `===` node-reference comparison bug in the WASM sibling-walk. web-tree-sitter returns a fresh wrapper object from every `.child()` call, so two accessors for the same underlying node are never reference-equal — only `.id` is stable. This silently broke the **pre-existing** chained-method-call case too: `obj.method1().method2()` resolved **both** calls to `"method2"` (whichever selector the broken loop scanned last), not their own names — not just the new bare-call path this PR adds.

Together these took Dart's resolution-benchmark recall from a genuine **0% floor to 100% precision / 61.9% recall (13/21)**. The remaining false negatives are all receiver-typed calls needing typeMap population for Dart fields/parameters — filed as #2319 rather than expanding this PR's scope. Complexity/CFG metrics still use the signature-only node for AST walking (a separate, already-tracked issue, #2182) and are untouched here.

## Verification
- lint: pass
- typecheck: pass
- Rust: `cargo test` — 811 passed (up from 801, 10 new tests)
- TS: 281 files / 4535 tests, including 11 new unit tests (`tests/parsers/dart.test.ts`)
- Engine parity verified end-to-end: rebuilt the native addon locally and confirmed native and WASM produce byte-identical `calls`/`definitions` for every fix scenario (bare calls, chained calls, multi-line endLine, bodyless constructors/methods)
- Resolution-benchmark threshold updated to reflect the new achieved baseline (ratcheted below the exact value to avoid flakiness)

Closes #2082

### DIFF
--- a/crates/codegraph-core/src/extractors/dart.rs
+++ b/crates/codegraph-core/src/extractors/dart.rs
@@ -34,6 +34,7 @@ fn match_dart_node(node: &Node, source: &[u8], symbols: &mut FileSymbols, _depth
         "constructor_invocation" | "new_expression" => handle_dart_constructor_call(node, source, symbols),
         "type_alias" => handle_dart_type_alias(node, source, symbols),
         "selector" => handle_dart_selector(node, source, symbols),
+        "call_expression" => handle_dart_call_expression(node, source, symbols),
         _ => {}
     }
 }
@@ -133,20 +134,42 @@ fn extract_dart_class_methods(body: &Node, class_name: &str, source: &[u8], symb
                                 None => continue,
                             }
                         }
-                        None => continue,
+                        None => {
+                            // A bodyless (semicolon-only) member — `Foo();` (a
+                            // constructor, the short form idiomatic Dart uses
+                            // throughout) or `double area();` (an abstract
+                            // method with no implementation) — wraps its
+                            // constructor_signature/function_signature in a
+                            // `declaration` node instead of the
+                            // `method_declaration` a block-bodied member uses
+                            // (confirmed by parsing both forms with
+                            // tree-sitter-dart 0.2; #2082). Check for that
+                            // shape before skipping — otherwise every
+                            // fixture/codebase using this idiomatic form
+                            // silently loses its constructors/abstract methods.
+                            let bodyless_sig = find_child(&member, "declaration").and_then(|d| {
+                                find_child(&d, "constructor_signature")
+                                    .or_else(|| find_child(&d, "method_signature"))
+                                    .or_else(|| find_child(&d, "function_signature"))
+                            });
+                            match bodyless_sig {
+                                Some(s) => s,
+                                None => continue,
+                            }
+                        }
                     }
                 }
                 // Direct signatures at the top of the class body (some grammar versions)
                 _ => member,
             };
             match sig.kind() {
-                "method_signature" | "function_signature" => {
+                "method_signature" | "function_signature" | "constructor_signature" => {
                     if let Some(fn_name) = extract_dart_fn_name(&sig, source) {
                         symbols.definitions.push(Definition {
                             name: format!("{}.{}", class_name, fn_name),
                             kind: "method".to_string(),
                             line: start_line(&sig),
-                            end_line: Some(end_line(&sig)),
+                            end_line: Some(dart_function_end_line(&sig)),
                             decorators: None,
                             complexity: compute_all_metrics(&sig, source, "dart"),
                             cfg: build_function_cfg(&sig, "dart", source),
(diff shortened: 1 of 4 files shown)


carlos-alm:
Confirmed and fixed in 9846969c. A comment between a signature and its body is its own intervening sibling in tree-sitter-dart (the grammar's comment rule is an `extra` production, not folded into an adjacent node) — verified by parsing directly. \`dartFunctionEndLine\`/\`dart_function_end_line\` now skip forward past any number of consecutive comment siblings before checking for the real body (or a bare \`;\` for a bodyless signature).

Added regression tests for both a single comment and multiple consecutive comments between signature and body, in both engines — all pass, and the fix doesn't regress any of the other endLine/bodyless-member cases.

@greptileai please re-review.

PR #2329, opened 2026-08-07
fix(hooks): guard-git.sh clean guard blocks dry-run and matches raw substrings

### LINKED ISSUE #2099: guard-git.sh's clean-command guard blocks --dry-run too (and matches on raw command substrings)
## Bug

`.claude/hooks/guard-git.sh` (line ~78) blocks every `git clean` invocation unconditionally:

```bash
if echo "$NCOMMAND" | grep -qE '(^|[[:space:]]|&&[[:space:]]*)git[[:space:]]+clean'; then
  deny "BLOCKED: 'git clean' deletes untracked files that may belong to other sessions."
fi
```

This also blocks the dry-run form, which is a pure listing operation that never deletes anything. The `/housekeep` skill's Phase 2 explicitly recommends the dry-run form of `git clean -fdX` as the discovery mechanism for gitignored dirt files, so the hook currently makes that step of the skill impossible to run as documented.

Worse: the grep matches on the raw command **text**, not just actual invocations. Any Bash command whose argument text merely contains the substring "git clean" gets blocked too — e.g. `gh issue create --body "...git clean -fdX --dry-run..."` was blocked while filing this very issue, because the deny message and quoted body text contained that substring.

## Fix

The check should:
1. Special-case `--dry-run`/`-n` flags and allow those through, only blocking invocations that would actually delete (lack `-n`/`--dry-run` and include `-f`/`--force`).
2. Scope the grep to the actual command being executed (e.g. first word of a pipeline segment / structured parse), not a blind substring match over the entire raw command string — so quoted text in unrelated commands (like `gh issue create --body "..."`) doesn't trigger it.

## Found via

/housekeep run on 2026-07-12, discovered while running Phase 2 (dirt file discovery), then again while trying to file this issue.

carlos-alm:
## Summary

Closes #2099

## Bug 1: dry-run always blocked

The clean-command guard in `.claude/hooks/guard-git.sh` blocked every invocation unconditionally, including `-n`/`--dry-run` — a pure listing operation that never deletes anything. `/housekeep`'s Phase 2 explicitly recommends the dry-run form of `clean -fdX` as the discovery mechanism for gitignored dirt files, so the hook made that documented step impossible to run.

Fixed: only blocks when `-f`/`--force` is present and `-n`/`--dry-run` is absent. Git itself always treats `--dry-run` as overriding `-f` (never deletes either way), so `-fdX --dry-run` together is also allowed. A bare invocation with neither flag is left unblocked too, since git itself refuses to run it without `-f` (or a config override this hook has no visibility into anyway).

## Bug 2: raw substring matching, not command-boundary aware

Worse: every "does this command invoke a dangerous verb" check in the file (not just clean) scans the raw command **text**, with no awareness of shell quoting. Any Bash command whose argument text merely contains a git-clean-shaped substring gets blocked too — e.g. `gh issue create --body "...please dont run git-clean..."` was blocked while filing the original issue, because the quoted body text contained that substring. The same applies to a heredoc-authored commit message body (`git commit -m "$(cat <<'EOF' ... EOF)"`, the exact form this repo's own CLAUDE.md mandates) that happens to discuss the bug it's fixing.

Fixed at the root rather than per-check: added `mask-quoted-text.mjs`, which replaces the contents of every quoted string (single or double) and every heredoc body with placeholder characters before the verb-detection checks run, preserving quote/line structure. This fixes the false-positive class for every check in the file (add, reset, checkout, restore, clean, stash, push, commit, `gh pr create`), not just clean. Extraction logic that needs real values (`detect_work_dir`, `MSG_FILE`, the AI-attribution scan) keeps reading the raw, unmasked command — only the verb-detection checks (`$NCOMMAND`) get the masked version.

## Test plan

- [x] New `tests/unit/hook-guard-git-clean.test.ts` — 17 tests covering: dry-run/force distinction (5 cases), quote/heredoc false-positive avoidance while still blocking real invocations (6 cases), and the masking script directly (6 cases)
- [x] `npm test` — full suite, 4568 passed
- [x] `bash -n .claude/hooks/guard-git.sh` clean
- [x] Manually verified via direct hook invocation (piping constructed `tool_input` JSON) against the exact repro from the issue and the heredoc-commit-message scenario

### DIFF
--- a/.claude/hooks/check-git-clean-force.mjs
+++ b/.claude/hooks/check-git-clean-force.mjs
@@ -5,17 +5,21 @@
 // otherwise prints nothing. Used by guard-git.sh (#2099).
 //
 // Splits into logical command segments on every real shell command
-// separator (&&, ||, ;, |, newline) — not just `&&` — so a flag on one
+// separator (&&, &, ||, ;, |, newline) — not just `&&` — so a flag on one
 // command can never be misread as belonging to an unrelated `git clean` on
-// a different segment (Greptile review: `git clean -fd; ls -n` must still
-// block, since the `-n` belongs to `ls`, not the clean invocation).
+// a different segment (Greptile review: `git clean -fd; ls -n` and
+// `git clean -fd & ls -n` must still block, since the `-n` belongs to the
+// other command, not the clean invocation).
 //
 // Within a segment, flags are matched against whole whitespace-separated
 // tokens (not a substring/boundary regex against the raw segment text), so
 // bundled short options are recognized correctly: `-ndf` and `-fnd` both
 // carry both `-n` and `-f`, matching git's own bundled short-flag parsing.
+// Tokens at or after a bare `--` are pathspecs, not options (git's own
+// convention) — `git clean -f -- -n` must still block, since the trailing
+// `-n` names a literal path, not a dry-run flag (Greptile review).
 
-const SEGMENT_SEPARATOR = /&&|\|\||[;|\n]/;
+const SEGMENT_SEPARATOR = /&&|&|\|\||[;|\n]/;
 const CLEAN_INVOCATION = /(^|\s)git\s+clean(\s|$)/;
 const FORCE_TOKEN = /^(--force|-[a-zA-Z]*f[a-zA-Z]*)$/;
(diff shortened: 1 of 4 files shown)


carlos-alm:
@greptileai

carlos-alm:
@greptileai

Issue #2330, opened 2026-08-07
docs/examples/claude-code-hooks: pre-commit.sh, post-git-ops.sh, enrich-context.sh, update-graph.sh have also drifted from the live hooks


carlos-alm:
## Found while working on

#2105 (docs/examples/claude-code-hooks/guard-git.sh had drifted from .claude/hooks/guard-git.sh). While fixing that one and adding a byte-identity test (\`tests/unit/hook-guard-git-clean.test.ts\`) to prevent recurrence, checked the rest of the example hooks for the same problem:

\`\`\`
diff -q .claude/hooks/pre-commit.sh docs/examples/claude-code-hooks/pre-commit.sh       # differ
diff -q .claude/hooks/post-git-ops.sh docs/examples/claude-code-hooks/post-git-ops.sh   # differ
diff -q .claude/hooks/enrich-context.sh docs/examples/claude-code-hooks/enrich-context.sh  # differ
diff -q .claude/hooks/update-graph.sh docs/examples/claude-code-hooks/update-graph.sh   # differ
\`\`\`

\`track-edits.sh\` and \`lint-staged.sh\` are still byte-identical (not affected).

## Why out of scope for #2105

#2105 was specifically scoped to \`guard-git.sh\`; fixing 4 more files (each needing its own careful diff review to confirm the docs copy is either a deliberate, documented simplification or genuine drift) is a larger, separate body of work.

## Suggested fix

Same treatment as #2105 applied to each file:
1. Diff each pair and decide whether the difference is intentional (documented in the README, like guard-git.sh's now-removed branch-validation omission was) or unintentional drift.
2. For unintentional drift, resync the docs copy with the live hook.
3. Extend the docs-sync byte-identity test (\`tests/unit/hook-guard-git-clean.test.ts\`'s "docs example stays in sync" describe block, or a new dedicated test file) to cover these files too, so CI catches future drift instead of relying on someone noticing during an unrelated PR — the same mechanism that caught \`guard-git.sh\`'s original divergence.

PR #2331, opened 2026-08-07
docs(hooks): resync guard-git.sh example with the live hook

### LINKED ISSUE #2105: follow-up: docs/examples/claude-code-hooks/guard-git.sh has drifted far from .claude/hooks/guard-git.sh
Noticed while working on PR #2100 (fix/guard-git-mergehead-skip).

**Context:** `docs/examples/claude-code-hooks/guard-git.sh` is meant to be an example copy of `.claude/hooks/guard-git.sh`, but it has drifted significantly out of sync — missing the `-C` worktree detection (`detect_work_dir`), branch-name validation, AI-attribution blocking in commit messages, and now the MERGE_HEAD merge-commit bypass added in PR #2100. A `diff` between the two files shows the example is missing most of the current hook's logic.

**Why out of scope for #2100:** the drift predates this PR by many prior PRs/features; fixing it is a separate, standalone doc-sync concern, not something introduced or required by the MERGE_HEAD fix.

**Suggested fix:** either resync `docs/examples/claude-code-hooks/guard-git.sh` with the current `.claude/hooks/guard-git.sh`, or replace the static copy with a build/CI step (or a note) that keeps the example in sync automatically so this doesn't keep drifting.

carlos-alm:
## Summary

Closes #2105

`docs/examples/claude-code-hooks/guard-git.sh` had drifted far from `.claude/hooks/guard-git.sh` — missing worktree detection (`detect_work_dir`), branch-name validation, AI-attribution blocking in commit messages, the `MERGE_HEAD` skip from PR #2100, and now the entire quote-masking/dry-run-aware clean-guard rework from #2099. It no longer reflected the real hook's behavior at all.

## Changes

- Copies the current `.claude/hooks/guard-git.sh` and its two companion scripts (`mask-quoted-text.mjs`, `check-git-clean-force.mjs`) verbatim into the docs example.
- Adds a byte-identity test (`tests/unit/hook-guard-git-clean.test.ts`, "docs example stays in sync" block) so CI fails if these ever diverge again — a prose README note already failed to prevent the original drift once, so this needed to be enforced rather than just documented.
- Updates the README's hook description and removes the now-stale "the example version omits branch validation" note (the example now includes it, since it's an exact copy).

Filed #2330 for the same class of drift found on `pre-commit.sh`, `post-git-ops.sh`, `enrich-context.sh`, and `update-graph.sh` while checking the rest of the example hooks — out of scope for this fix, which is specifically about `guard-git.sh` per the issue.

## Test plan

- [x] `npx vitest run tests/unit/hook-guard-git-clean.test.ts` — 32 passed (including 3 new sync-check tests)
- [x] `npm test` — full suite, 4583 passed
- [x] `diff .claude/hooks/guard-git.sh docs/examples/claude-code-hooks/guard-git.sh` — no output (identical)

### DIFF
--- a/.claude/hooks/mask-quoted-text.mjs
+++ b/.claude/hooks/mask-quoted-text.mjs
@@ -154,6 +154,18 @@ function maskQuotedStrings(input) {
     const ch = input[i];
     if (quote) {
       if (quoteExecTriggered) {
+        // Mirror the escaped-quote handling below: an escaped `\"` inside an
+        // exec-triggered double-quoted payload is NOT the closing quote —
+        // treating it as one (Greptile review) would end the exemption
+        // early and let the real remainder of the payload (which can
+        // contain a real dangerous command, e.g.
+        // `bash -c "echo \"x\" && git clean -fd"`) fall through to normal
+        // masking, hiding it.
+        if (ch === '\\' && quote === '"' && i + 1 < input.length) {
+          out += ch + input[i + 1];
+          i++;
+          continue;
+        }
         out += ch;
         if (ch === quote) quote = null;
         continue;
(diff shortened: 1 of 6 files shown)


carlos-alm:
@greptileai

PR #2332, opened 2026-08-07
fix(deps): pin typescript below 7.x and warn loudly when an incompatible version is detected

### LINKED ISSUE #2106: TypeScript 7.x removes the classic Program/TypeChecker API that ts-resolver.ts depends on
Found while reviewing PR #2093 (`dependabot/npm_and_yarn/typescript-7.0.2`, bumping `typescript` devDependency from 6.0.3 to 7.0.2).

## Root cause

TypeScript 7.0 is the native/Go-ported compiler rewrite. Its npm package restructures `package.json` `exports`:

```
"exports": {
  ".": "./lib/version.cjs",
  "./unstable/sync": "./dist/api/sync/api.js",
  "./unstable/async": "./dist/api/async/api.js",
  ...
}
```

The `"."` entry point — what `import ts from 'typescript'` resolves to — now only exports `{ version, versionMajorMinor }`. The classic in-process, synchronous Compiler API (`ts.createProgram`, `ts.Program`, `ts.TypeChecker`, `ts.sys`, `ts.readConfigFile`, `ts.parseJsonConfigFileContent`, `ts.forEachChild`, `ts.isIdentifier`, etc.) has been removed from the package entirely — verified there is no `readConfigFile`/`createProgram` anywhere in the shipped `lib/` or `dist/` at all, not even as an unexported internal.

In its place, TS 7 ships an out-of-process, RPC-based API under `typescript/unstable/sync` and `typescript/unstable/async` (`API`, `Client`, `Snapshot`, `Project`, `Program`, `Checker`, `NodeHandle`, ...) that spawns/talks to a separate compiler process and returns proxy objects (`NodeHandle`, `Symbol`, `TypeObject`) fetched over a wire protocol, rather than raw AST nodes you can walk directly.

## Impact

`src/domain/graph/resolver/ts-resolver.ts` ("Phase 8.1: TypeScript-native type resolver") is built entirely around the old synchronous, in-process model: it calls `ts.readConfigFile`/`ts.sys`/`ts.parseJsonConfigFileContent` to build compiler options, `ts.createProgram(...)` to get a `Program`, `program.getTypeChecker()` to get a `TypeChecker`, and then walks the tree-sitter-adjacent AST directly with `ts.forEachChild`, `ts.isIdentifier`, `ts.isCallExpression`, etc., querying the checker synchronously per-node (`checker.getTypeAtLocation(node)`).

None of this maps onto the new client/server model without a substantial redesign:
- No more direct `SourceFile`/`Node` walking — the new API returns `NodeHandle` proxies you `resolve()` against a `Project`, fetched from a spawned server.
- No more synchronous single-call `checker.getTypeAtLocation(node)` — every checker operation is a request to an external process.
- Config resolution (`readConfigFile`/`parseJsonConfigFileContent`) has no equivalent surface in the new API; the closest analog is `API.parseConfigFile`, which returns a different shape.

**Compile-time**: bumping to `typescript@7.x` makes `tsc` fail outright — every CI job that runs `npm run build` (Lint, Validate commits, Engine parity ×3, Test Node 22 ×3, Pre-publish benchmark gate, impact, CI Testing Pipeline, Embedding regression tests) fails downstream of the same ~70 TS2339/TS2694/TS7006 errors, all confined to `ts-resolver.ts`.

**Runtime** (verified locally, independent of the compile-time break): even if the type errors were silenced, `(await import('typescript')).default` at runtime under TS 7 only has `{version, versionMajorMinor}` — `ts.readConfigFile` is `undefined`, so `createProgram()`'s try/catch swallows a `TypeError` and returns `null`. The feature would not crash, but would silently and permanently no-op for any project (including codegraph analyzing itself) where the installed `typescript` is v7+, discarding the compiler-verified (confidence 1.0) type enrichment this pass exists to provide, with only a `debug()`-level trace to indicate it happened.

## Recommendation

This is not a small type-signature fix — it requires redesigning `ts-resolver.ts`'s TSC integration around the new `typescript/unstable/sync` (or `/async`) client-server API: spawning/managing the compiler process, converting from tree-sitter node positions to `DocumentIdentifier`/`NodeHandle` lookups, and reworking every direct `Program`/`Checker`/`Node` call site (~70 call sites) to the new proxy-object model. This should be scoped as its own project, not bundled into a dependency-bump PR.

Until this is done, `typescript` should stay pinned below `7.0.0` (dependabot PR #2093 should not be merged as-is — recommend `@dependabot ignore this major version` on that PR until this issue is resolved).

carlos-alm:
## Summary

Closes #2106

TypeScript 7.x replaced the classic in-process Program/TypeChecker API `src/domain/graph/resolver/ts-resolver.ts` is built entirely around (`readConfigFile`, `createProgram`, `sys`, `forEachChild`, ...) with an out-of-process client-server model under `typescript/unstable/sync|async` — the package's `"."` entry point now only exports `{version, versionMajorMinor}`. Redesigning `ts-resolver.ts`'s ~70 call sites around the new proxy-object model is a substantial project of its own (as the issue itself concludes), not something to bundle into a routine dependency bump.

## Changes

- **Dependabot ignore rule** for major-version bumps of `typescript`, so it stops proposing 7.x on its weekly schedule. The dependency-bump PR that originally surfaced this (#2093) was already closed unmerged, and `package.json`'s own `^6.0.2` range already excludes 7.x for regular `npm install`/`npm update` — but dependabot doesn't respect that range when suggesting major bumps, so without this it would keep reopening the same PR.
- **Explicit, loud detection of the incompatibility for real-world usage**: `ts-resolver.ts` lazy-loads `typescript` as a peer/optional dependency of whatever project codegraph is analyzing, not codegraph's own pin — so the dependabot rule alone doesn't protect a user running codegraph against their own TS7 project. Previously this silently and permanently no-op'd the compiler-verified type enrichment pass with only a `debug()`-level trace. Now detects the missing classic API (`readConfigFile` undefined) explicitly and logs a clear, actionable `warn()` pointing at this issue, instead of degrading invisibly.
- Extracted the version-compatibility check as a small pure function (`hasClassicCompilerApi`) so it's unit-testable without mocking the real `typescript` module import.

## Test plan

- [x] `npx vitest run tests/unit/ts-resolver.test.ts` — 25 passed (including 4 new tests for `hasClassicCompilerApi`)
- [x] `npm test` — full suite, 4588 passed
- [x] `npx tsc --noEmit` clean
- [x] `codegraph diff-impact --staged -T` reviewed — 0 unexpected callers affected
- [x] Manually verified the new `.github/dependabot.yml` `ignore` rule parses correctly (`python3 -c "import yaml; yaml.safe_load(...)"`)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,18 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+      # TypeScript 7.x replaced the classic in-process Program/TypeChecker
+      # API (createProgram, TypeChecker, ts.sys, forEachChild, ...) with an
+      # out-of-process client-server model under typescript/unstable/sync|async.
+      # src/domain/graph/resolver/ts-resolver.ts is built entirely around the
+      # old synchronous API (~70 call sites) and breaks at compile time and
+      # silently no-ops at runtime under v7 (issue #2106) — needs its own
+      # redesign project before this repo can move off the 6.x line, not a
+      # routine dependency-bump PR. Remove this ignore once that redesign
+      # lands.
+      - dependency-name: "typescript"
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: "github-actions"
     directory: "/"
--- a/src/domain/graph/resolver/ts-resolver.ts
+++ b/src/domain/graph/resolver/ts-resolver.ts
@@ -17,7 +17,7 @@
  */
 import fs from 'node:fs';
 import path from 'node:path';
-import { debug } from '../../../infrastructure/logger.js';
+import { debug, warn } from '../../../infrastructure/logger.js';
 import type { CallAssignment, ExtractorOutput, TypeMapEntry } from '../../../types.js';
 
 // typescript is not a hard dependency — lazy-load it so JS-only projects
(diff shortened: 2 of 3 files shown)


carlos-alm:
@greptileai

Issue #2339, opened 2026-08-08
follow-up: barrel-only classification misclassifies a hybrid file at exact reexports==ownDefs parity


carlos-alm:
Discovered while writing a regression test for #2138 (incremental rebuild losing cross-file dispatch edges).

`detect_barrel_only_files` (Rust) / its JS mirror classify a file as "barrel-only" using the heuristic `reexports >= ownDefs`. A file with exactly ONE re-export and exactly ONE own definition (`reexports == ownDefs == 1`) hits this `>=` and gets classified as barrel-only — even though it's a genuine hybrid file (real logic + a re-export), not a pure barrel.

Once a file is classified barrel-only, `build_and_insert_call_edges`/its JS equivalent skip emitting ANY of its own outgoing `calls`/`receiver` edges entirely (`if import_ctx.barrel_only_files.contains(rel_path) { continue; }`, `crates/codegraph-core/src/domain/graph/builder/pipeline.rs`) — so on any incremental build where this file gets pulled into Stage 6b's barrel-candidate reparse (e.g. because something imports from it), its own call/receiver edges are silently dropped and never re-emitted, even on an otherwise-correct build.

Repro sketch: a file with exactly one `export { x } from './other.js'` and exactly one function definition that calls into another file — its own outgoing edges vanish the moment it's reparsed as a barrel candidate.

Suggested fix: change the heuristic's tie-break — either `reexports > ownDefs` (strict) instead of `>=`, or require some minimum reexport-to-def ratio, so a file with real logic isn't misclassified purely because it happens to have exactly as many reexports as definitions. Needs verification against whatever the ORIGINAL "barrel-only" heuristic was trying to protect (issue #1848's comment in `pipeline.rs` around `barrel_candidates_added`) before changing the comparison operator, since it may have been chosen deliberately for some other edge case.

Issue #2087, opened 2026-07-11
follow-up: persist cross-file invoked-property-name evidence for incremental dead-code classification


carlos-alm:
Deferred from PR #2034 review.

**Original reviewer comment:** inline comment — https://github.com/optave/ops-codegraph-tool/pull/2034#discussion_r3543999171

**Context:**

PR #2034 added a liveness check for object-literal-property value-refs (`{ resolve: someFn }`): the referenced function only gets a `calls` edge when the property key is independently confirmed to be invoked via member-call syntax (`x.resolve(...)`) somewhere in the files being processed (`collectInvokedPropertyNames` in `src/domain/graph/builder/call-resolver.ts`, mirrored in `crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs`).

On a **full build**, "the files being processed" is the whole codebase, so this is exact. On an **incremental rebuild** (`src/domain/graph/builder/incremental.ts`'s `buildCallEdges`), it is scoped to only the single file currently being rebuilt. If `factory.js` defines `{ resolve: neverCalled }` and the only `table.resolve(1)` call site lives in `consumer.js`, an incremental rebuild triggered solely on `factory.js` will not see `resolve` in `invokedPropertyNames` and will drop the edge — classifying `neverCalled` as dead even though it is genuinely reachable. A subsequent full rebuild corrects this, but a dead-code CI check running against an incremental watch build could see a transient false positive.

This is the same category of trade-off already made elsewhere in this codebase's incremental role classification (`hasActiveFileSiblings` and exported-via-reexport reachability both scope to an affected-file subset rather than the whole graph in `crates/codegraph-core/src/graph/classifiers/roles.rs`'s incremental path — verified by direct code reading, see `do_classify_incremental`, `find_neighbour_files`, `is_barrel_prod_reachable`), so it was accepted as-is for #2034 rather than blocking that PR.

**What the fix entails:** to close this gap without falling back to a full-codebase scan on every incremental rebuild, the codebase would need a small persistent index — e.g. a `invoked_property_names` table (or a column/marker on existing `calls`-shaped edges) that incremental rebuilds insert into/prune from as files are added, changed, or removed, so `collectInvokedPropertyNames` can query evidence across the *whole graph* in O(lookup) time rather than only the currently-parsed file(s). This is a real schema/incremental-maintenance change (new persisted state + insert/prune logic on both engines), not a one-line tweak, so it's being tracked separately rather than folded into #2034.

PR #2340, opened 2026-08-08
fix(incremental): persist per-file return-type evidence across builds

### LINKED ISSUE #2138: bug: incremental rebuild triggered by an unrelated file loses 10 receiver-dispatch edges to WasmWorkerPool
## Found during dogfooding v3.16.0

**Severity:** Medium
**Command:** `codegraph build` (incremental)

## Reproduction

```bash
codegraph build .                                    # full baseline build
sqlite3 .codegraph/graph.db "SELECT n1.file,n1.name,n1.line,e.kind,n2.file,n2.name,n2.line \
  FROM edges e JOIN nodes n1 ON e.source_id=n1.id JOIN nodes n2 ON e.target_id=n2.id \
  ORDER BY 1,2,3,4,5,6,7;" | sort > /tmp/edges_baseline.txt

# Edit a COMPLETELY UNRELATED file — src/domain/graph/builder/pipeline.ts,
# which has no direct relationship to src/domain/parser.ts or wasm-worker-pool.ts:
echo "" >> src/domain/graph/builder/pipeline.ts
codegraph build .                                     # incremental: "1 changed, 0 removed"

# Revert the edit back to the exact original bytes:
git show HEAD:src/domain/graph/builder/pipeline.ts > src/domain/graph/builder/pipeline.ts
codegraph build .                                     # incremental again: "1 changed, 0 removed"

sqlite3 .codegraph/graph.db "SELECT ... same query ..." | sort > /tmp/edges_after.txt
comm -23 /tmp/edges_baseline.txt /tmp/edges_after.txt
```

## Expected behavior

Editing (and then reverting to byte-identical content) a file with no relationship to `parser.ts`/`wasm-worker-pool.ts` should leave every edge involving those two files completely unchanged after incremental rebuild. Edge count and edge set should match a full `--no-incremental` rebuild of the same tree exactly.

## Actual behavior

The graph permanently loses exactly 10 edges (43195 → 43185, confirmed via full edge-set diff, not just count) and does not recover them even after reverting the triggering file back to its original content:

```
src/domain/parser.ts|backfillTypeMap|1049|calls|src/domain/wasm-worker-pool.ts|WasmWorkerPool.parse|183
src/domain/parser.ts|backfillTypeMap|1049|receiver|src/domain/wasm-worker-pool.ts|WasmWorkerPool|158
src/domain/parser.ts|backfillTypeMapBatch|1142|calls|src/domain/wasm-worker-pool.ts|WasmWorkerPool.parse|183
src/domain/parser.ts|backfillTypeMapBatch|1142|receiver|src/domain/wasm-worker-pool.ts|WasmWorkerPool|158
src/domain/parser.ts|ensureWasmTrees|369|calls|src/domain/wasm-worker-pool.ts|WasmWorkerPool.parse|183
src/domain/parser.ts|ensureWasmTrees|369|receiver|src/domain/wasm-worker-pool.ts|WasmWorkerPool|158
src/domain/parser.ts|parseFileAuto|1112|calls|src/domain/wasm-worker-pool.ts|WasmWorkerPool.parse|183
src/domain/parser.ts|parseFileAuto|1112|receiver|src/domain/wasm-worker-pool.ts|WasmWorkerPool|158
src/domain/parser.ts|parseFilesWasm|1181|calls|src/domain/wasm-worker-pool.ts|WasmWorkerPool.parse|183
src/domain/parser.ts|parseFilesWasm|1181|receiver|src/domain/wasm-worker-pool.ts|WasmWorkerPool|158
```

All 10 are `calls -> WasmWorkerPool.parse` (method call) or `receiver -> WasmWorkerPool` (constructor/receiver-type binding) edges from 5 caller functions in `parser.ts`, all targeting the `WasmWorkerPool` class in `wasm-worker-pool.ts`. Notably, the *sibling* `calls -> getWasmWorkerPool` edge from each of those same 5 callers survives untouched — only the receiver-typed dispatch to the `WasmWorkerPool` class itself is lost.

The `--verbose` log for the triggering incremental build shows:
```
[codegraph DEBUG] runPostNativeCha: Gate A (hierarchy) fired — running full scan
```
suggesting the incremental CHA/receiver-dispatch re-resolution ("Gate A" full scan) doesn't reproduce the same `WasmWorkerPool` receiver-dispatch edges that a genuine `--no-incremental` build computes, even though it's not scoped to only the touched file.

Once lost, a further no-op incremental build does not recover them — only `--no-incremental` restores the correct 43195-edge graph.

## Root cause

Likely in the native incremental orchestrator's CHA/receiver-dispatch re-resolution path (`runPostNativeCha` / "Gate A" hierarchy full-scan, per the debug log), which appears to run a resolution pass over class-hierarchy/receiver-typed dispatch edges but produces a different (smaller) result set than the equivalent full-build pass for the `WasmWorkerPool` class specifically.

## Suggested fix

Compare the "Gate A full scan" receiver-dispatch resolution logic against the full-build equivalent for `WasmWorkerPool`-style singleton/class receiver patterns (`getWasmWorkerPool()` returning a `WasmWorkerPool` instance, `.parse()` called on it) and find where the incremental path fails to emit or fails to preserve the `receiver`/`calls` edges targeting the class's own methods.

carlos-alm:
## Summary

Closes #2138

An incremental `codegraph build` triggered by editing a file with NO relationship to a getter/class dispatch pair permanently lost the `calls`/`receiver` edges for that dispatch (10 edges in the original report, e.g. `parser.ts`'s calls to `WasmWorkerPool.parse` via `getWasmWorkerPool()`).

## Root cause

Stage 6b's barrel-candidate re-parse re-parses any file the changed file imports from, if that file has its own re-export (making it "barrel-like" in the DB's eyes) — regardless of whether the changed file's edit has anything to do with that file's own logic. Re-parsing wipes the barrel-like file's outgoing `calls`/`receiver` edges and re-derives them from an in-memory cross-file return-type index scoped only to *this build's* file set. A factory/getter defined in a file that wasn't re-parsed this build (like `wasm-worker-pool.ts`'s `getWasmWorkerPool()`) silently drops out of that index, so the dispatch edges it fed are lost — and stay lost, since reverting the triggering edit re-triggers the same re-parse/re-derive cycle without ever seeing the untouched file's return type again.

Reproducible on both engines (verified with a dedicated fixture): the WASM/JS pipeline's `resolve-imports.ts`/`build-edges.ts` and the native orchestrator's `pipeline.rs` both re-parse barrel candidates and re-derive return types from a build-scoped index only.

## Fix

Persists per-file return-type evidence into a durable `return_types` table (schema v30, both engines) whenever a file is parsed — mirroring the `invoked_property_names` pattern already established for #2087:

- `src/db/migrations.ts` / `crates/codegraph-core/src/db/connection.rs`: new `return_types(file, fn_name, type_name, confidence)` table, migration v30.
- `src/db/repository/build-stmts.ts` / `crates/codegraph-core/.../detect_changes.rs`: per-file purge and full-reset entries, mirroring `invoked_property_names`.
- `persistReturnTypes` (`build-edges.ts`) / `import_edges::persist_return_types` (Rust): writes each build's own `returnTypeMap`/`return_type_map` entries.
- `propagateReturnTypesAcrossFiles` (`build-edges.ts`) / `build_return_type_index` (`pipeline.rs`): now union in DB-persisted entries for files not present in the current build's file set — in-memory data always wins for files this pass actually re-parsed, so a stale persisted row can never override fresh data.

## Test plan

- [x] New `tests/integration/issue-2138-incremental-return-type-persistence.test.ts` (parametrized on both engines): full build establishes `getPool() -> Pool.doWork()` dispatch edges, then touching an unrelated file that merely imports from the barrel-like caller confirms the edges survive the incremental rebuild.
- [x] Manually verified against the exact bug shape end-to-end (fixture with a getter/class pair in an untouched file + a barrel-adjacent caller + an unrelated trigger file) before and after the fix, both engines.
- [x] Full `cargo test --lib` passes (820 tests).
- [x] Full `npm test` passes (4602 tests, 286 files).

Filed #2339 separately for an unrelated edge case discovered while building the regression fixture: the barrel-only classification heuristic (`reexports >= ownDefs`) misclassifies a hybrid file as barrel-only at exact parity, stripping its own edges independent of this fix.

### DIFF
--- a/crates/codegraph-core/src/db/connection.rs
+++ b/crates/codegraph-core/src/db/connection.rs
@@ -502,6 +502,32 @@ const MIGRATIONS: &[Migration] = &[
       CREATE INDEX IF NOT EXISTS idx_invoked_property_names_name ON invoked_property_names(name);
     "#,
     },
+    Migration {
+        // #2138: durable, per-file record of every function/method's
+        // inferred return type — needed so cross-file return-type
+        // propagation (propagate_return_types_across_files /
+        // propagateReturnTypesAcrossFiles) can resolve
+        // `const x = importedFactory(); x.method()`-shaped dispatch to a
+        // file this build never re-parsed. Without this, an incremental
+        // build that re-parses a barrel-adjacent file wipes that file's
+        // outgoing calls/receiver edges and re-derives them from an
+        // in-memory return-type index scoped only to this build's file set,
+        // so a factory/getter defined in an untouched file silently drops
+        // out of dispatch resolution until a full --no-incremental rebuild.
+        // Populated once per full or incremental build (both engines).
+        // Mirrors src/db/migrations.ts v30.
+        version: 30,
+        up: r#"
+      CREATE TABLE IF NOT EXISTS return_types (
+        file TEXT NOT NULL,
+        fn_name TEXT NOT NULL,
+        type_name TEXT NOT NULL,
+        confidence REAL NOT NULL,
+        PRIMARY KEY (file, fn_name)
+      );
+      CREATE INDEX IF NOT EXISTS idx_return_types_file ON return_types(file);
+    "#,
+    },
 ];
 
 // ── napi types ──────────────────────────────────────────────────────────
--- a/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
@@ -725,6 +725,12 @@ pub fn run_pipeline(
     import_edges::persist_reexport_renames(conn, &import_ctx.reexport_map)
         .map_err(|e| format!("reexport rename persistence failed: {e}"))?;
 
+    // #2138: persist this pass's own return-type evidence before it's read
+    // back below — gives a later incremental build a durable, whole-graph
+    // view of files it doesn't itself re-parse.
+    import_edges::persist_return_types(conn, &file_symbols)
+        .map_err(|e| format!("return type persistence failed: {e}"))?;
+
     // Build import edges. A write failure here (transaction-start, a
     // malformed chunk, or commit) propagates via `?` instead of being
     // discarded — the old `run_pipeline` had no way to know edges were
@@ -737,7 +743,7 @@ pub fn run_pipeline(
     // Phase 8.2: cross-file return-type propagation — seed each file's
     // type_map with the return types of imported functions before call-edge
     // building, mirroring propagateReturnTypesAcrossFiles in build-edges.ts.
-    propagate_return_types_across_files(&mut file_symbols, &import_ctx);
+    propagate_return_types_across_files(conn, &mut file_symbols, &import_ctx);
 
     // Build call edges using existing Rust edge_builder (internal path)
     // For now, call edges are built via the existing napi-exported function's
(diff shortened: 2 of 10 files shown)


carlos-alm:
@greptileai

PR #2341, opened 2026-08-08
fix(bench): only set CODEGRAPH_ENGINE for actual engine-comparison forks

### LINKED ISSUE #2140: bug(bench): embedding-benchmark.ts forked workers print a spurious CODEGRAPH_ENGINE warning
## Found during dogfooding v3.16.0

**Severity:** Low
**Command:** `node scripts/embedding-benchmark.ts` (any invocation, all models)

## Reproduction

```bash
node --experimental-strip-types --import ./scripts/ts-resolve-loader.js scripts/embedding-benchmark.ts
```

Output for every model worker includes:
```
[codegraph WARN] CODEGRAPH_ENGINE="minilm" is not a valid engine value (expected auto|native|wasm). Falling back to "auto".
```
(and again for `jina-small`, `jina-base`, etc. — once per model tested).

## Root cause

`scripts/lib/fork-engine.ts:79` sets `CODEGRAPH_ENGINE: workerName` unconditionally when forking a worker:

```ts
env: { ...process.env, [envKey]: workerName, CODEGRAPH_ENGINE: workerName },
```

This is correct for the build/query benchmarks, where `workerName` is `"wasm"` or `"native"` (valid `CODEGRAPH_ENGINE` values). `embedding-benchmark.ts` reuses the same fork helper with `workerName` set to the *model* name (`"minilm"`, `"jina-small"`, ...), which is not a valid engine value, triggering the warning on every embed call inside the worker. It's caught and falls back to `"auto"` so it doesn't affect correctness — purely cosmetic noise.

## Suggested fix

Only set `CODEGRAPH_ENGINE` when the fork is for an engine-comparison benchmark (build/query), not for `embedding-benchmark.ts`'s model workers — e.g. gate it on a boolean param passed by the caller instead of always deriving it from `workerName`.

carlos-alm:
## Summary

Closes #2140

`forkWorker()` (`scripts/lib/fork-engine.ts`) unconditionally set `CODEGRAPH_ENGINE` to whatever `workerName` it was given. That's correct for `forkEngines()`'s own two calls (`workerName` is `'wasm'`/`'native'`, a valid engine value), but `embedding-benchmark.ts` reuses `forkWorker()` directly with `workerName` set to a *model* name (`'minilm'`, `'jina-small'`, ...) — an invalid `CODEGRAPH_ENGINE` value that tripped a `"not a valid engine value"` warning on every embed call inside that worker. Purely cosmetic noise (it already fell back to `"auto"` correctly), but a real fix rather than living with it.

## Fix

`forkWorker()` now only sets `CODEGRAPH_ENGINE` when the caller's `envKey` is the same `WORKER_ENV_KEY` `forkEngines()` itself uses internally — true for an actual engine-comparison fork, never true for a caller using its own distinct env key (like `embedding-benchmark.ts`'s `MODEL_WORKER_KEY`). No call-site changes needed anywhere, including `embedding-benchmark.ts` itself.

## Test plan

- [x] New `tests/unit/fork-engine-codegraph-engine-env.test.ts`: forks a tiny throwaway child script and confirms `CODEGRAPH_ENGINE` is set when `envKey` is the real `WORKER_ENV_KEY`, and unset for a different envKey.
- [x] Full `npm test` suite passes (4606 tests, 287 files).

### DIFF
--- a/scripts/lib/fork-engine.ts
+++ b/scripts/lib/fork-engine.ts
@@ -24,7 +24,10 @@
 import { fork } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 
-const WORKER_ENV_KEY = '__BENCH_ENGINE__';
+// Exported so tests can verify forkWorker() only sets CODEGRAPH_ENGINE for
+// this exact envKey (the engine-comparison fork), not for other callers'
+// own envKey (e.g. embedding-benchmark.ts's model-name workers, issue #2140).
+export const WORKER_ENV_KEY = '__BENCH_ENGINE__';
 const TARGETS_ENV_KEY = '__BENCH_TARGETS__';
 
 /**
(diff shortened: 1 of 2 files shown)


carlos-alm:
@greptileai

PR #2342, opened 2026-08-08
fix(export): add id-based join fields to file-level exportJSON edges

### LINKED ISSUE #2144: follow-up: exportJSON edge source/target schema inconsistent between file-level and function-level modes
Deferred from PR #2141 review.

**Original reviewer comment:** Greptile summary body — https://github.com/optave/ops-codegraph-tool/issues/2141#issuecomment-5020314056

**Context:** `exportJSON` in `src/features/export.ts` produces edges with a different `source`/`target` schema depending on `opts.fileLevel`:
- File-level (`fileLevel: true`, default): `edges[].source`/`target` are file path strings (`n1.file AS source, n2.file AS target`), which don't even match the file-level `nodes[].id` numeric field in the same payload.
- Function-level (`fileLevel: false`, `--functions`): `edges[].source`/`target` are numeric node IDs (`e.source_id`/`e.target_id`), which do match `nodes[].id`.

Greptile flagged this as "a minor API inconsistency in new code with no existing consumers" (Confidence Score 5/5, not blocking). The function-level behavior is actually the more internally-consistent one (edges reference `nodes[].id`), but changing the file-level behavior to match would be a breaking change to the existing `codegraph export -f json` output shape, which has existing consumers today — out of scope for PR #2141, whose mandate is narrowly to fix the `--functions` flag being ignored for `json`/`graphson` formats.

**Suggested follow-up:** Decide on a consistent identifier scheme for `exportJSON` edges across both levels (e.g. always key edges by `nodes[].id`, keeping file paths only as a display property), and version/document the change since it affects the public JSON export shape.

carlos-alm:
## Summary

Closes #2144

`exportJSON`'s two modes produce edges with a different `source`/`target` schema:
- File-level (default): `edges[].source`/`target` are file path strings, which don't join against this same payload's `nodes[].id`.
- Function-level (`--functions`): `edges[].source`/`target` are numeric node IDs, matching `nodes[].id`.

Changing file-level's `source`/`target` to numeric IDs directly (matching function-level) would be a breaking change to the existing `codegraph export -f json` output shape, which has real consumers relying on file paths there today — flagged in the issue as needing "a consistent identifier scheme... version/document the change since it affects the public JSON export shape."

## Fix

Purely additive: file-level edges now also carry `sourceId`/`targetId` (numeric, matching `nodes[].id`), alongside the existing `source`/`target` file-path strings. Consumers who want the same id-based join function-level mode already supports can use the new fields; anyone reading `source`/`target` as file paths today is unaffected.

## Test plan

- [x] New test in `tests/graph/export.test.ts`: file-level edges' `sourceId`/`targetId` resolve against `nodes[].id`, while `source`/`target` remain file path strings.
- [x] Verified end-to-end against this repo's own graph via `codegraph export -f json`.
- [x] Full `npm test` suite passes (4607 tests, 287 files).

### DIFF
--- a/src/features/export.ts
+++ b/src/features/export.ts
@@ -343,15 +343,29 @@ export function exportJSON(
       .all() as Array<{ id: number; name: string; kind: string; file: string; line: number }>;
     if (noTests) nodes = nodes.filter((n) => !isTestFile(n.file));
 
+    // sourceId/targetId (#2144): file-level edges' source/target are file
+    // path strings for backward compatibility with existing `codegraph
+    // export -f json` consumers, but that means they don't join against
+    // this same payload's nodes[].id the way function-level mode's edges
+    // do. sourceId/targetId are purely additive so a consumer that wants
+    // the same id-based join function-level mode already supports can use
+    // them without breaking anyone reading source/target as file paths.
     let edges = db
       .prepare(`
-      SELECT DISTINCT n1.file AS source, n2.file AS target, e.kind, e.confidence
+      SELECT DISTINCT n1.file AS source, n2.file AS target, n1.id AS sourceId, n2.id AS targetId, e.kind, e.confidence
       FROM edges e
       JOIN nodes n1 ON e.source_id = n1.id
       JOIN nodes n2 ON e.target_id = n2.id
       WHERE n1.file != n2.file AND e.confidence >= ?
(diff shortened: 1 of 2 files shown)


carlos-alm:
@greptileai

Issue #1888, opened 2026-07-06
Same-file bare-name lookup (resolveCallTargets) is unfiltered by kind and runs before type-aware dispatch, causing false calls-edges to unrelated classes/variables


carlos-alm:
## Summary

`resolveCallTargets` (`src/domain/graph/builder/call-resolver.ts`) resolves a call's targets as follows:

```ts
if (!targets || targets.length === 0) {
  targets = lookup.byNameAndFile(call.name, relPath);   // <-- unqualified, same-file, NO kind filter
  if (targets.length === 0) {
    targets = resolveByMethodOrGlobal(lookup, call, relPath, typeMap, callerName);
  }
}
```

This bare `lookup.byNameAndFile(call.name, relPath)` step runs **unconditionally for every call** (regardless of receiver — `this.x()`, `obj.x()`, bare `x()` are all treated the same at this step) and is **not filtered by symbol kind**. If any same-file node happens to share the call's bare name — a class, an interface, a top-level `const`/`variable` binding, anything — it wins immediately, before any receiver-type-aware resolution (`resolveByReceiver`, `resolveByGlobal`, the same-class qualified-method fallback, etc.) ever gets a chance to run.

This means a call can resolve to a same-named class/variable even when a **correctly-typed, legitimately-callable target also exists** via qualified lookup — the bare unqualified match is checked first and wins by construction.

## Repro

```js
class Registry {
  bar() {
    return 1;
  }
}

const obj = new Registry();

function getter() {
  this.bar();
}

Object.defineProperty(obj, 'x', { get: getter });

class bar {}
```

Building this (either engine — confirmed identical on both `wasm` and `native`, so this is a correctness bug, not an engine-parity gap) produces:

```
getter -> bar   (kind: class)   ❌ wrong — should be Registry.bar (kind: method)
```

Removing the trailing `class bar {}` makes it resolve correctly to `Registry.bar` (kind: method) on the WASM engine (confirms `Registry.bar` — a real, correctly-typed target — was reachable and correct; it's specifically being pre-empted by the coincidentally-named class).

## Why this matters

[...]

Issue #1867, opened 2026-07-06
Unused parameter warning: get_first_call_arg's source in extractors/javascript.rs


carlos-alm:
While rebuilding the native crate for an unrelated change (issue #1753), \`cargo build --lib\` emits:

\`\`\`
warning: unused variable: \`source\`
    --> crates/codegraph-core/src/extractors/javascript.rs:2499:48
     |
2499 | fn get_first_call_arg<'a>(call_node: &'a Node, source: &[u8]) -> Option<Node<'a>> {
     |                                                ^^^^^^ help: if this is intentional, prefix it with an underscore: \`_source\`
\`\`\`

This is pre-existing (unrelated to #1753's points-to solver config threading) — introduced by commit 3e8035de4b149e8fcd55012f4501b7ce6a31a4a9 ("fix: gate Array.from's callback arg by position, not just callee name"). Low priority: either the \`source\` parameter should be prefixed with \`_\` if intentionally unused, or removed from the signature (and all call sites) if truly dead.

carlos-alm:
Duplicate of #1817 — same `get_first_call_arg` unused `source` parameter warning in `crates/codegraph-core/src/extractors/javascript.rs`, just re-discovered independently at a shifted line number. The fix already exists as commit `c0e7eae8` ("fix(native): remove unused source param from get_first_call_arg") on branch `fix/issue-1817-unused-source-param-on-get-first-call-arg-in`, open as PR #1960. Verified directly: the function signature in the current stacked-branch tree already has no `source` parameter, and a dedicated test block covers all three call sites. No further work needed here — closing as duplicate once #1960 merges.

carlos-alm:
PR #1960 (fix(native): remove unused source param from get_first_call_arg) merged on 2026-07-09, closing the duplicate #1817 this issue was tracking. Verified against current `main`: `get_first_call_arg` in `crates/codegraph-core/src/extractors/javascript.rs` no longer takes a `source` parameter. Closing as duplicate/already resolved — no further work needed.

PR #2351, opened 2026-08-08
fix(ci): add non-blocking reminder for non-closing issue references in PRs

### LINKED ISSUE #2181: Recurring pattern: merged PRs using "Refs #N" instead of "Closes #N" leave already-fixed issues open
## Summary

While running `/fixer` against the low-numbered open-issue backlog, 3 of the first 4 issues processed (#1763, #1867, #1888) turned out to already be fully fixed and merged to `main` — but were never auto-closed because the merging commit/PR referenced them with `Refs #N` (or, in one case, simply didn't reference the duplicate issue number at all) instead of a closing keyword (`Closes #N`/`Fixes #N`).

Specific cases found in this run:
- **#1763** (`db.busyTimeoutMs` config wiring): fixed across PRs #1943 and #2021, neither of which referenced #1763 by number at all.
- **#1867** (unused `source` param, duplicate of #1817): #1817 was closed by PR #1960, but #1867 (the duplicate re-discovery) was left open — the fixing PR had no reason to know about the duplicate.
- **#1888** (same-file bare-name kind filter): fixed in the squashed PR #2026 (commit `fa678d41`), whose body says `Refs #1888` rather than `Closes #1888`.

## Why this matters

This repo's own convention (CLAUDE.md / memory: "PR bodies must use `Closes #N` / `Fixes #N`, not bare `(#N)`, to auto-close issues on merge") already exists specifically to prevent this — but nothing currently enforces it, and `Refs #N` slips through review as easily as `Closes #N` since both look like a valid issue reference at a glance. The practical cost: any tooling that works the open-issue backlog by number (`/fixer`, manual triage, `gh issue list`) re-investigates issues that have zero remaining work, burning time on redundant analysis instead of the real backlog.

## Suggested fix direction

Options to evaluate (not prescribing one):
- A CI check on PR merge (or a scheduled job) that flags PRs whose body contains `Refs #N`/`Related #N`/bare `(#N)` for an issue number that a human should confirm is intentionally left open vs. should have been a closing keyword.
- A periodic sweep (script or `/fixer`-adjacent skill) that cross-references merged commit messages containing an issue number against that issue's open/closed state, and prompts to close any that are stale.
- Simplest: a `commitlint`/PR-template reminder specifically calling out the closing-keyword requirement, since `Refs #N` is apparently a natural thing to write when the intent is just "this touches that issue" without realizing it silently fails to close it.

Filed per this repo's scope-discipline convention (CLAUDE.md) rather than addressed inline, since fixing the underlying process gap is out of scope for `/fixer`'s actual job (working the issue backlog).

carlos-alm:
## Summary

Closes #2181

Three issues (#1763, #1867, #1888) were already fully fixed and merged to `main` but stayed open because the merging PR used "Refs #N" instead of a closing keyword — which looks like a valid reference but does not trigger GitHub's auto-close on merge. Nothing previously caught this.

## Fix

A hard-failing CI gate isn't the right shape here: a non-closing reference (`Refs #N`, `Related #N`, a partial fix) is often intentional, and text alone can't distinguish that from a genuine mistake. Instead:

- New `.github/workflows/closing-keyword-check.yml`: scans the PR body for issue references using a non-closing phrase (`Refs #N`, `Related #N`, `related to #N`, `see #N`, or a bare `(#N)`) that have no closing keyword (`closes`/`fixes`/`resolves` + variants) for that same issue number, and posts (or clears, on later edits) a single informational PR comment. It never fails the build — it only makes the gap visible for a human to confirm.
- Uses `pull_request_target` rather than `pull_request`: the check only reads the trusted event payload's `body` field — no checkout, no code execution — so this is safe without the `workflow_run` split `codegraph-impact-comment.yml` uses for its riskier (code-executing) case, and it also works correctly on fork PRs, which get a read-only token under plain `pull_request`.
- Clarified the existing "Related issues" hint in `.github/PULL_REQUEST_TEMPLATE.md` to spell out that `Refs`/`Related` won't auto-close the issue — addressing the PR-template-reminder option directly, as defense in depth alongside the automated check.

## Test plan

- [x] Verified the regex/dedup logic locally against 8 cases (closes-only, refs-only, refs+closes for the same number, "related to", bare parens, lowercase "fixes", mixed multi-issue body, no references at all) — all produce the expected flagged set.
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` — valid YAML.
- [x] `actionlint .github/workflows/closing-keyword-check.yml` — zero findings.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -26,4 +26,6 @@ N/A
 
 ## Related issues
 
-<!-- Link related issues: Fixes #123, Closes #456 -->
+<!-- Use a closing keyword if this PR fully resolves the issue: Closes #123, Fixes #456.
+     "Refs #123" / "Related #123" will NOT auto-close the issue on merge — only use those
+     for a non-closing reference. -->
--- a/.github/workflows/closing-keyword-check.yml
+++ b/.github/workflows/closing-keyword-check.yml
@@ -52,7 +52,12 @@ jobs:
               repo: context.repo.repo,
               issue_number: prNumber,
             });
-            const existing = comments.find((c) => c.body.includes(marker));
+            // Restrict to the workflow's own bot-authored comments — matching on the
+            // marker text alone would let a PR participant plant it in their own
+            // comment and have a later edit/reopen event overwrite or delete it.
+            const existing = comments.find(
+              (c) => c.body.includes(marker) && c.user.login === 'github-actions[bot]'
+            );
 
             if (flagged.size === 0) {
               if (existing) {
--- a/.github/workflows/closing-keyword-check.yml
+++ b/.github/workflows/closing-keyword-check.yml
@@ -1,0 +1,90 @@
+# Reminds PR authors when a description references an issue with a non-closing
+# phrase ("Refs #N", "Related #N", bare "(#N)") but no closing keyword for that
+# same number — the pattern behind issue #2181, where merged PRs silently left
+# already-fixed issues open because "Refs #N" looks like a valid reference but
+# does not trigger GitHub's auto-close. Informational only: a non-closing
+# reference is often intentional (partial fix, related work), so this never
+# fails the build — it only makes the gap visible for a human to confirm.
+#
+# Uses pull_request_target (not pull_request) so the check also works on PRs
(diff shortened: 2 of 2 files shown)


carlos-alm:
@greptileai

Issue #1973, opened 2026-07-07
checkNoDeletedExportsInUse consumer refs also conflate file-level imports-type with symbol-level calls


carlos-alm:
## Context

While fixing #1830 (exports output `consumers[]` conflating file-level `imports-type` entries with symbol-level `calls` entries), I found the same underlying ambiguity in a separate, unrelated code path.

`checkNoDeletedExportsInUse` in `src/features/check.ts` runs this query to find external consumers of a to-be-deleted exported symbol:

```sql
SELECT DISTINCT caller.name, caller.file, caller.line
FROM edges e
JOIN nodes caller ON e.source_id = caller.id
WHERE e.target_id = ? AND e.kind IN ('calls', 'imports-type') AND caller.file != ?
```

For `imports-type` edges, `e.source_id` is the *importing file's* node (see `emitNamedSymbolEdges` in `build-edges.ts`), so `caller.line` is always `0` and `caller.name` equals `caller.file`. `src/presentation/check.ts` then renders this as `<file>:0` in the CI violation message, e.g.:

```
Config (interface) — file types.ts deleted but still used by 1 external consumer(s): consumer.ts:0
```

The fabricated `:0` line number is misleading the same way the `exports` CLI output was before #1830 — it looks like a real call-site line but there's no such line.

## Suggested fix

Apply the same fix pattern used for #1830 (add a `consumerKind: 'file' | 'symbol'` discriminator derived from `caller.kind`, and have `presentation/check.ts` render file-level consumers without a fabricated line number) to `checkNoDeletedExportsInUse` / `ConsumerRef` in `src/features/check.ts` and its formatter in `src/presentation/check.ts`.

Out of scope for #1830 itself since it's a separate feature (CI predicate output), not the `exportsData` output the issue is scoped to.

PR #2364, opened 2026-08-08
fix(exports): consumer discriminator misclassifies top-level calls

### LINKED ISSUE #2189: exports' consumer discriminator (#1830) misclassifies top-level calls sourced from the file node
## Summary

While fixing #1973 (the same \`consumerKind\` discriminator for \`checkNoDeletedExportsInUse\`), Greptile caught that keying the discriminator off the *source node's kind* (\`n.kind === 'file'\`) rather than the *edge kind* (\`e.kind === 'imports-type'\`) misclassifies a genuine top-level call as a type-only import.

\`findCaller\` (\`src/domain/graph/builder/call-resolver.ts\`) falls back to the file node itself as a call's source when there's no enclosing function/method AND no enclosing variable/constant binding (a bare statement at module scope, e.g. a language like Haskell where `main` is a top-level `bind`, or any language with a real top-level side-effecting call). That produces a genuine \`calls\` edge whose \`source_id\` is a file-kind node — structurally identical, from the source-node's-kind point of view, to an \`imports-type\` edge (which is *always* sourced from the importing file's own node, by construction — see \`emitNamedSymbolEdges\`).

\`src/domain/analysis/exports.ts\`'s \`consumersStmt\` (added for #1830) has exactly this latent bug:

\`\`\`ts
const consumersStmt = cachedStmt(
  _consumersStmtCache,
  db,
  \`SELECT n.name, n.file, n.line, n.kind AS sourceKind FROM edges e JOIN nodes n ON e.source_id = n.id
       WHERE e.target_id = ? AND e.kind IN ('calls', 'imports-type')\`,
);
// ...
consumerKind: c.sourceKind === 'file' ? ('file' as const) : ('symbol' as const),
\`\`\`

A real top-level call to an exported function would be misclassified \`consumerKind: 'file'\` and rendered as a type-only import in \`codegraph exports\` output, hiding a genuine caller.

## Fix applied in #1973 (for check.ts, not this file)

Select \`e.kind AS edgeKind\` instead of the source node's kind, and discriminate on \`edgeKind === 'imports-type'\` — an edge's own kind is the correct, unambiguous signal (\`imports-type\` edges are always file-sourced by construction; \`calls\` edges are always genuine calls regardless of what happens to be at their source, including the file-node fallback for top-level calls).

## Suggested fix

Apply the identical one-line change to \`exports.ts\`'s \`consumersStmt\` and its consumer-mapping call site.

Filed separately since \`exports.ts\` isn't part of #1973's diff — scope discipline.

carlos-alm:
## Summary

Closes #2189

`exports.ts`'s `consumersStmt` keyed `consumerKind` off the SOURCE NODE's kind (`n.kind === 'file'`), not the edge's own kind. `findCaller` (`domain/graph/builder/call-resolver.ts`) falls back to the file node as a call's source when there's no enclosing function/method or variable binding — a bare top-level statement — which produces a genuine `calls` edge sourced from a file-kind node, structurally identical from the source-node's-kind point of view to an `imports-type` edge (always file-sourced by construction). A real top-level call to an exported function was misclassified `consumerKind: 'file'` and rendered as a type-only import, hiding a genuine caller.

## Fix

Applies the identical fix already validated for the same discriminator in `src/features/check.ts`'s `checkNoDeletedExportsInUse` (via `src/db/repository/edges.ts`'s `findExternalConsumers`, #1973): select `e.kind AS edgeKind` instead of the source node's kind, and discriminate on `edgeKind === 'imports-type'`.

## Test plan

- [x] Added a regression test reproducing the exact scenario (an exported function called from a bare top-level statement, sourced from the importing file's own node) — verified it fails against the old source-kind logic (`consumerKind: 'file'` instead of `'symbol'`) and passes against the fix.
- [x] Full `tests/integration/exports.test.ts`: 24 passed.
- [x] `npx tsc --noEmit`, `npm run lint`: clean.
- [x] Full `npm test`: 289 files, 4631 passed.

### DIFF
--- a/src/domain/analysis/exports.ts
+++ b/src/domain/analysis/exports.ts
@@ -20,7 +20,7 @@ const _consumersStmtCache: StmtCache<{
   name: string;
   file: string;
   line: number;
-  sourceKind: string;
+  edgeKind: string;
 }> = new WeakMap();
 const _reexportsFromStmtCache: StmtCache<{ file: string }> = new WeakMap();
 const _reexportsToStmtCache: StmtCache<{ file: string }> = new WeakMap();
@@ -184,7 +184,7 @@ function exportsFileImpl(
   const consumersStmt = cachedStmt(
     _consumersStmtCache,
     db,
-    `SELECT n.name, n.file, n.line, n.kind AS sourceKind FROM edges e JOIN nodes n ON e.source_id = n.id
+    `SELECT n.name, n.file, n.line, e.kind AS edgeKind FROM edges e JOIN nodes n ON e.source_id = n.id
          WHERE e.target_id = ? AND e.kind IN ('calls', 'imports-type')`,
   );
   const reexportsFromStmt = cachedStmt(
@@ -232,7 +232,7 @@ function exportsFileImpl(
         name: string;
         file: string;
         line: number;
-        sourceKind: string;
+        edgeKind: string;
       }>;
       if (noTests) consumers = consumers.filter((c) => !isTestFile(c.file));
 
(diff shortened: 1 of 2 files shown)


carlos-alm:
Verified this is a real observation, but it's an inherited characteristic of the already-merged #1973 reference implementation this PR deliberately mirrors, not a regression introduced here. `edges.ts`'s `findExternalConsumers` (the #1973 fix) has the identical behavior — its own test (`tests/integration/check.test.ts`, `'does not misclassify a genuine top-level call sourced from a file node as a type-only import (#1973)'`) deliberately asserts only `file`/`consumerKind`, not `name`/`line`, for exactly this reason: `findCaller`'s file-node fallback for a bare top-level statement genuinely has no better name/line to report today.

Fixing it properly needs a change to `findCaller`/the call-resolution layer to synthesize a real caller symbol or call-site for that fallback case, which would need to land for both `checkNoDeletedExportsInUse` and `exportsData` identically since they share the same root cause — broader than this PR's scope (mirroring #1973's discriminator fix). Filed as #2365.

@greptileai

Issue #2369, opened 2026-08-08
follow-up: full CLI.md/MCP.md output-content diffing against real invocations


carlos-alm:
## Context

#2212 extended docs-drift protection to \`docs/examples/CLI.md\` with two tiers (PR TBD):

1. **Static command/flag validity** — every documented \`codegraph <cmd> ...\` line's base command, subcommand (for command groups like \`registry\`/\`snapshot\`), and \`--flag\`s are checked against the CURRENT CLI's own \`--help\` output. Already caught one real bug: \`plot\`'s \`--cluster-by\` flag was renamed to \`--cluster\` and the docs were never updated.
2. **Representative real execution** — a curated subset of ~15 commands, with this-repo-specific symbol/file names substituted for a small fixture project's equivalents, are actually run against a freshly-built graph and asserted to exit successfully.

## What's still missing

Neither tier verifies that a documented example's **output content/shape** still matches what the command actually produces — only that the command+flags are syntactically valid and (for the curated subset) don't crash. A command's output format could be restructured (a section renamed/removed, e.g. "Graph Quality:" → "Health Score:") and neither tier would catch it.

This is the harder half of #2212's original ask, deliberately deferred because it needs real design decisions:

- **Dynamic-content normalization**: CLI.md's examples are captured against codegraph's own (constantly growing) codebase, so exact counts/paths/timestamps in the documented output will NEVER match a live re-run — any content diff needs rules for what to normalize away vs. compare exactly.
- **Strictness level**: structural (same section labels/keys present) vs. exact-value — the original issue itself flagged this as an open design question, not something to decide unilaterally in a docs-test PR.
- **MCP.md tool-call comparison**: comparable idea for MCP tool responses (key-shape match against real tool output from an actual MCP server call), entirely unaddressed by #2212 — MCP.md's own protection remains JSON-syntax-validity only (#2005).

## Suggested next step

Build a small fixture-graph harness that runs each documented command/tool-call for real and diffs *structural labels only* (extracting header-like lines, normalizing away embedded filenames/symbol-names/digits) against the documented block — i.e., pick a normalization/strictness policy and implement it, rather than leaving it as an open question. Consider starting with just the commands whose documented output already uses clear markdown-style headers (\`context\`, \`audit --quick\`), since those are the easiest to normalize reliably.

Filed per this repo's scope-discipline convention — #2212's own PR intentionally stayed at the "cheap mechanical floor" tier (mirroring #2005's own scope), not the full content-diffing vision.

Issue #2005, opened 2026-07-08
docs/examples: no automated check keeps CLI.md/MCP.md example output in sync with real output


carlos-alm:
## Context

While fixing #1873 (stale `audit`/`diff-impact` example output blocks in `docs/examples/CLI.md` and `docs/examples/MCP.md`), I found that these example blocks are hand-transcribed and have no automated verification against the real CLI/MCP output. That's very likely why they drifted out of sync with the implementation in the first place (the `audit` output shape has been rewritten at least once since the docs were written, and the `diff-impact` MCP JSON contract changed entirely).

## Suggestion

Add a lightweight check (could be a vitest test, or a small script run in CI) that either:
- Extracts each fenced JSON block in `docs/examples/MCP.md` and asserts it parses as valid JSON (would have caught the `diff_impact`/`audit` shape drift immediately since the tool return values differ structurally from what was documented), and/or
- Runs the documented `codegraph <cmd>` invocations from `docs/examples/CLI.md` against a small fixture graph and diffs the real output against the documented block (catches full output-format drift, not just JSON validity).

This is a docs-quality/tooling improvement, not a bug fix — filing separately per repo scope-discipline rather than bundling into #1873's docs-only diff.

PR #2370, opened 2026-08-08
test(docs): extend drift protection to CLI.md command/flag validity

### LINKED ISSUE #2212: docs/examples: extend drift protection to full CLI.md output diffing (not just JSON validity)
## Context

#2005 added an automated JSON-syntax-validity check (`tests/unit/docs-examples-json-validity.test.ts`) for every fenced ` ```json ` block in `docs/examples/MCP.md`. That's a cheap, mechanical floor — it catches syntactically broken examples (it already caught one: an unquoted `{ "..." }` placeholder in the `batch_query` response block) — but it does **not** verify that any example's *content* still matches real `codegraph`/MCP output. A block can be perfectly valid JSON while describing a shape the tool no longer returns.

## What's still missing

The original issue (#2005) proposed two independent approaches; only the first (JSON-validity) has been implemented:

- `docs/examples/CLI.md`'s command-output blocks (mostly plain-text terminal output, not JSON) have no automated check at all — the JSON-validity approach doesn't apply to them (its one `` ```json `` block is a `{ ... }` elision placeholder, not literal JSON, and would need to be excluded or handled specially).
- Neither doc's examples are checked against **real** output from running the documented `codegraph <cmd>` invocations (or MCP tool calls) against a fixture graph.

## Suggested approach

Build a small fixture-graph harness (could reuse `tests/fixtures/sample-project` or a dedicated docs-example fixture) that:
- Runs each documented `codegraph <cmd>` invocation from `CLI.md` and diffs the real output against the documented block, tolerant of non-deterministic content (absolute paths, timestamps, line numbers that shift as the codebase evolves).
- Optionally does the same for MCP tool calls in `MCP.md`, comparing real tool-response shape (keys present, not necessarily exact values) against the documented JSON.

This is materially larger in scope than #2005's JSON-validity check — it requires designing normalization rules for dynamic content and deciding how strict the diff should be (structural/key-shape match vs. exact-value match). Filing separately per repo scope-discipline rather than expanding #2005's PR.

carlos-alm:
## Summary

Closes #2212

#2005 added JSON-syntax validity for `MCP.md`'s ` ```json ` blocks — a cheap, mechanical floor that doesn't verify content matches real output, only that it parses. `CLI.md`'s blocks are plain-text terminal output, so that exact check doesn't apply; this extends the same "cheap floor" philosophy with two tiers instead of full output-content diffing (deliberately out of scope here — see the filed follow-up below):

- **Static command/flag validity** (all ~75 documented invocation lines): every documented `codegraph <cmd> ...` line's base command, subcommand (for command groups like `registry`/`snapshot`), and `--flag`s are checked against the CURRENT CLI's own `--help` output. **Caught a real bug on first run**: `plot`'s documented `--cluster-by` flag was renamed to `--cluster` and the docs were never updated — fixed in this PR.
- **Representative real execution** (~15 curated commands): this-repo-specific symbol/file names substituted for the small fixture project also used by `tests/integration/cli.test.ts` (since `CLI.md`'s examples reference this repo's own internals, which don't exist in any small fixture), actually run against a freshly-built graph, asserted to exit successfully. Catches a command that parses fine but crashes when invoked, which the static check can't see.

## Scope boundary

Filed #2369 for the harder remaining half: real output-**content** diffing (structural label comparison with dynamic-content normalization) for both `CLI.md` and `MCP.md`. The original issue itself flagged that as needing real design decisions (normalization rules for counts/paths/timestamps, strictness level) rather than something to decide unilaterally in this PR — mirroring exactly how #2005 (JSON-validity only) preceded this PR (command/flag validity + successful execution, still not content verification).

## Test plan

- [x] New `tests/integration/docs-examples-cli-drift.test.ts`: 92 tests, all passing after fixing the `plot` flag bug it found.
- [x] `npx tsc --noEmit`, `npm run lint`: clean.
- [x] Full `npm test`: 290 files, 4723 passed.

### DIFF
--- a/docs/examples/CLI.md
+++ b/docs/examples/CLI.md
@@ -605,7 +605,7 @@ Generates an interactive HTML viewer powered by vis-network. Open in a browser f
 - Detail panel with metrics, callers, callees on node click
 
 ```bash
-codegraph plot --functions --color-by complexity --cluster-by community
+codegraph plot --functions --color-by complexity --cluster community
 ```
 
 ---
--- a/tests/integration/docs-examples-cli-drift.test.ts
+++ b/tests/integration/docs-examples-cli-drift.test.ts
@@ -1,0 +1,196 @@
+/**
+ * Extends #2005's docs-drift protection (originally JSON-syntax validity
+ * for docs/examples/MCP.md's ```json blocks) to docs/examples/CLI.md,
+ * whose blocks are plain-text terminal output, not JSON — a different kind
+ * of check is needed (#2212).
+ *
+ * Two tiers, deliberately NOT full output-content diffing (#2212 itself
+ * flags that as materially larger scope — it needs designing normalization
+ * rules for dynamic content like counts/paths/timestamps and deciding how
+ * strict a diff should be, which a docs-example test file shouldn't decide
+ * unilaterally):
+ *
+ *  - Tier 1 (every documented command, cheap/static): the base command (and,
+ *    for command groups like `registry`/`snapshot`, the subcommand) and
+ *    every `--flag` used in each ```bash block still exist in the CURRENT
+ *    CLI's own --help output. Catches the most common and highest-impact
+ *    form of drift — a renamed/removed command or flag — without needing to
+ *    run anything against a real graph.
+ *  - Tier 2 (a curated, representative subset, real execution): actually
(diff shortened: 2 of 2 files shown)


carlos-alm:
@greptileai

PR #2424, opened 2026-08-10
fix(test-filter): -T recognizes test/fixture directories and per-language conventions

### LINKED ISSUE #2256: roles --role dead -T: hand-authored test fixture directories aren't excluded, causing reachability noise
## Summary

While validating #2032's transitive-reachability dead-code fix by running `codegraph roles --role dead -T` on this repo's own `src/`, I found that the bulk of the newly-flagged-dead nodes (50 of 53 in the final delta) come from directories that are clearly test infrastructure but aren't excluded by `-T`:

- `tests/benchmarks/resolution/fixtures/**` — the hand-authored, per-language (34 languages) resolution-precision fixture projects documented in `CLAUDE.md`'s Test Structure section (`expected-edges.json` manifests).
- `tests/benchmarks/resolution/tracer/**` — the dynamic call tracers (`loader-hook.mjs`, `python-tracer.py`, `lua-tracer.lua`, `r-tracer.R`, `php-tracer.php`, etc.).
- `tests/fixtures/sample-project/**` — the ES-module sample fixture (`math.js`/`utils.js`/`index.js`).

None of these filenames match `TEST_FILE_PATTERNS` (`%.test.%`, `%.spec.%`, `%__test__%`, `%__tests__%`, `%.stories.%` in `src/graph/classifiers/roles.ts` / the mirrored Rust `TEST_FILE_PATTERNS`) — they're plain source files (`main.c`, `service.ml`, `validators.gleam`, `reflection.kt`, ...) that happen to live under a `tests/` directory, which the current filename-suffix-only heuristic doesn't check for.

## Why this matters

These fixtures are deliberately tiny, self-contained per-language snippets with no genuine external entry point recognized by codegraph's own root-detection (no CLI/MCP/exported-and-called-cross-file signal) — they exist purely to exercise parser extraction precision/recall. Before #2032, direct fan-in alone kept most of their internal call chains classified non-dead; #2032's stricter reachability check (correctly) now flags almost everything in these disconnected components as dead, since nothing in them is a confirmed-live root by any of codegraph's own criteria.

This isn't a bug in #2032's logic — it's a pre-existing gap in `-T`'s test-file detection becoming more visible now that dead-code classification is stricter. It affects any analysis that relies on `-T` to scope to "real" application code (`roles --role dead -T`, `triage -T`, `audit -T`, etc.), not just the new reachability pass.

## Suggested fix

Extend the test-file exclusion heuristic (`TEST_FILE_PATTERNS` / `testFilterSQL` in `src/db/index.ts` and the mirrored Rust `TEST_FILE_PATTERNS`/`test_file_filter_col` in `crates/codegraph-core/src/graph/classifiers/roles.rs`) to also recognize path-based test/fixture markers, e.g. any path segment matching `tests/`, `fixtures/`, `__fixtures__/`, or similar — needs care to avoid over-broadly excluding legitimately-shipped code that happens to live under a `test-utils`-style production directory.

## Related

- #2032 — the transitive-reachability fix whose self-build validation surfaced this

carlos-alm:
Closes #2256

## Problem

`-T`'s test-file exclusion only matched filename markers (`.test.`, `.spec.`, `__test__`, `__tests__`, `.stories.`) — it never recognized the dominant test/fixture layout convention used by most non-JS ecosystems: a `tests/`/`test/`/`fixtures/`/`__fixtures__/` directory, or Go's `_test.go` and Python's `test_*.py`/`*_test.py` filename suffixes.

On this repo's own self-build, 431 of 647 "dead -T" symbols were pure fixture noise from `tests/benchmarks/resolution/fixtures/**` (the hand-authored per-language resolution-precision fixtures) and `tests/fixtures/sample-project/**` — none of which match the filename-only heuristic despite living entirely under `tests/`.

The predicate was also independently duplicated across **9 call sites** (not the 2-3 named in the issue) — `src/infrastructure/test-filter.ts`, `src/db/query-builder.ts` (two separate spots: `testFilterSQL` and `NodeQuery.excludeTests`), `src/domain/search/search/filters.ts`, `src/features/ast.ts`, `src/features/complexity-query.ts`, `src/features/manifesto.ts`, `src/db/repository/in-memory-repository.ts` (two spots), plus the Rust mirrors in `crates/codegraph-core/src/graph/classifiers/roles.rs` and `crates/codegraph-core/src/db/repository/graph_read.rs`. Some had already drifted from each other before this fix.

## Fix

Consolidated into a single canonical pattern list per engine:
- TS: `src/infrastructure/test-filter.ts` — `TEST_FILE_LIKE_PATTERNS` (SQL LIKE templates) + `isTestFile()` (boolean predicate). Every other TS call site now delegates to one of these instead of hand-rolling its own copy.
- Rust: new `crates/codegraph-core/src/infrastructure/test_filter.rs`, mirroring the TS module (added to the `lib.rs` module-parity table). `roles.rs` and `graph_read.rs` both delegate to it now.

Path segments are bounded by `/` (or path start/end) so `tests/`/`test/` never matches `contests/`, `latest/`, or a production `test-utils/` directory — the exact false-positive risk the issue called out.

The Java `TestFoo.java`/`FooTest.java` filename convention is recognized precisely by `isTestFile`'s case-sensitive, camelCase-boundary-aware regex, but **deliberately excluded** from the SQL LIKE pattern list: SQLite's `LIKE` is case-insensitive by default, so a substring approximation would also match ordinary words merely containing "test" (`Latest.java`, `Contest.java`, `Protest.java`, `Testament.java`). The standard Maven/Gradle `src/test/java/` directory layout is already covered by the `test` path segment either way, so this is a deliberate, documented asymmetry rather than a gap.

While consolidating, found and fixed a latent LIKE-escaping bug: `_` is LIKE's own single-character wildcard, so the pre-existing `__test__`/`__tests__` patterns were unintentionally matching e.g. `"XXtestZZ"` (verified empirically against real SQLite). All literal underscores are now escaped with `ESCAPE '\'`.

## Testing

- New `tests/unit/test-filter.test.ts` (43 cases) + expanded `tests/unit/query-builder.test.ts`: filename markers, directory segments, Go/Python suffixes, explicit non-matches (`test-utils/`, `contests/`, `latest/`, `protest.py`), Java filename convention, and a parity check between `isTestFile` and the SQL filter.
- New Rust unit tests in `test_filter.rs` mirror the same battery.
- Self-build validation (this repo, 1049 files): `roles --role dead -T` dropped from 647 to 289, with **zero** remaining dead symbols under any `tests?/`/`fixtures/`/`__fixtures__/` directory (was 431).
- Full suite: 4899 JS tests + 891 Rust tests pass. `tsc --noEmit`, `npm run lint`, `cargo fmt --check` all clean.
- Dual-engine parity: `node scripts/parity-compare.mjs` — all 42 fixtures `PARITY OK`.

### DIFF
--- a/crates/codegraph-core/src/db/repository/graph_read.rs
+++ b/crates/codegraph-core/src/db/repository/graph_read.rs
@@ -63,25 +63,17 @@ fn push_file_filter(
     *idx += files.len();
 }
 
-/// Check if a file path looks like a test file (mirrors `isTestFile` in JS).
+/// Check if a file path looks like a test file. Delegates to the shared
+/// `infrastructure::test_filter` module (single source of truth, #2256);
+/// mirrors TS `isTestFile`.
 fn is_test_file(file: &str) -> bool {
-    file.contains(".test.")
-        || file.contains(".spec.")
-        || file.contains("__test__")
-        || file.contains("__tests__")
-        || file.contains(".stories.")
+    crate::infrastructure::test_filter::is_test_file(file)
 }
 
-/// Build test-file exclusion clauses for a column.
+/// Build test-file exclusion clauses for a column. Delegates to the shared
+/// `infrastructure::test_filter` module (single source of truth, #2256).
 fn test_filter_clauses(column: &str) -> String {
-    format!(
-        "AND {col} NOT LIKE '%.test.%' \
-         AND {col} NOT LIKE '%.spec.%' \
-         AND {col} NOT LIKE '%__test__%' \
-         AND {col} NOT LIKE '%__tests__%' \
-         AND {col} NOT LIKE '%.stories.%'",
-        col = column,
-    )
+    crate::infrastructure::test_filter::test_file_filter_col(column)
 }
 
 /// Read a full NativeNodeRow from a rusqlite Row by column name.
--- a/crates/codegraph-core/src/graph/classifiers/roles.rs
+++ b/crates/codegraph-core/src/graph/classifiers/roles.rs
@@ -61,14 +61,6 @@ const ENTRY_PATH_PATTERNS: &[&str] = &[
 /// Mirrors `COMMANDER_DISPATCH_NAMES` in `graph/classifiers/roles.ts`.
 const COMMANDER_DISPATCH_NAMES: &[&str] = &["execute", "validate"];
 
-const TEST_FILE_PATTERNS: &[&str] = &[
-    "%.test.%",
-    "%.spec.%",
-    "%__test__%",
-    "%__tests__%",
-    "%.stories.%",
-];
-
 // ── Output types ─────────────────────────────────────────────────────
 
 #[napi(object)]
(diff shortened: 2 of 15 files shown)


carlos-alm:
@greptileai

Issue #2032, opened 2026-07-08
roles --role dead: non-transitive fan-in means a function called only by another dead function is never flagged dead


carlos-alm:
## Summary

`codegraph roles --role dead` (and the underlying `classifyRoles`/`classify_node` in both engines) treats "has at least one inbound `calls` edge" as sufficient evidence of liveness — it never checks whether that caller is itself reachable from anything live. This means a genuinely-unreachable function that is called only by another genuinely-unreachable function is never flagged dead, because the direct fan-in count is > 0 regardless of the caller's own reachability.

## Concrete repro

```ts
function computeHelper(x: number): number {
  return x + 1;
}

function makeThing(): { doThing: () => number } {
  return {
    doThing: () => computeHelper(1),
  };
}

export function useThing(): number {
  const t = makeThing();
  return 42; // t.doThing() is never called anywhere
}
```

`computeHelper` has exactly one caller — the closure assigned to `doThing` — so its fan-in is 1 and it is classified `core`/non-dead, even though `doThing` (and therefore `computeHelper`) is never actually invoked by anything reachable from `useThing`'s real control flow.

This is the residual mechanism behind the historical #1770/#1895 investigation: even after #1895 fixes the object-literal-property value-ref liveness check (bare-identifier/shorthand property values now require independent invocation evidence), a function whose *only* caller is itself dead still isn't flagged, because `roles --role dead` never verifies that a caller is reachable — it only checks that a caller exists at all.

## Suggested fix direction

Replace (or augment) the direct fan-in check with proper reachability analysis: compute the set of nodes reachable via `calls` edges from confirmed-live roots (`entry`-classified nodes, exported public API, framework dispatch points), and classify anything outside that reachable set as dead — regardless of whether it happens to have some inbound edge from another also-unreachable node. This is a substantial change to the dead-code classifier's core algorithm (in both `src/graph/classifiers/roles.ts` and `crates/codegraph-core/src/graph/classifiers/roles.rs`), needs its own design pass (worklist/BFS from roots vs. the current single-pass per-node classification), and should be scoped as its own issue rather than folded into a narrower fix.

## Related

- #1895 — object-literal property-value references counted as liveness without checking if the property is invoked (narrower, now fixed for the bare-identifier/shorthand value-ref mechanism)
- #1770 — the concrete `Partition` dead-code removal that originally surfaced this gap

PR #2444, opened 2026-08-10
fix(resolver): this-bound EventEmitter callback registrations get a real calls edge

### LINKED ISSUE #2259: Method-kind event-callback registrations (EventEmitter-style) aren't recognized as reachability roots
## Summary

#2032 added a transitive-reachability check to the dead-code classifier. To avoid over-promoting genuinely-dead methods to \"root\" status (a gap Greptile flagged during that PR's review), `isInterfaceDispatchMethodRoot` (`src/graph/classifiers/roles.ts`, mirrored as `is_interface_dispatch_method_root` in `crates/codegraph-core/src/graph/classifiers/roles.rs`) only treats a `method`-kind, fan-in-0, hasActiveFileSiblings-rescued method as an unconditional reachability root when its bare name matches an actual `interface`/`type` declaration member somewhere in the codebase (e.g. `enterNode` matching `interface Visitor { enterNode?(...): ...; }`).

This correctly narrows the false-positive class Greptile identified (an ordinary unused class method with no relation to any interface). But it also means a **method-kind callback registered with a runtime event emitter** (Node's `Worker`/`EventEmitter` `.on('event', cb)` pattern, not a TypeScript-declared interface) no longer qualifies as a root, since there's no local interface/type declaration for codegraph to match against.

## Confirmed real instance (self-build)

`WasmWorkerPool.onMessage` in `src/domain/wasm-worker-pool.ts:276` is a genuine, live method (presumably wired up via `worker.on('message', ...)` or similar) with `fanIn === 0` — nothing in the graph resolves a call to it, since Node's built-in `EventEmitter` registration isn't a TypeScript interface member. Post-#2032, it is correctly still classified `leaf` itself (via the existing `hasActiveFileSiblings` rescue, untouched by the reachability downgrade), but it is no longer promoted to root status, so its own callees are wrongly flagged dead:

- `deserializeResult`, `deserializeCoreFields`, `deserializeBindingFields`, `deserializeMapFields` (all in `wasm-worker-pool.ts`) — called only by `onMessage`, now `dead-unresolved`.

Checked: `onMessage` does not repeat across multiple files in this codebase (ruled out a "same name in ≥2 files" heuristic as an easy fix — it wouldn't help here, since this is a singular occurrence, structurally indistinguishable from a genuinely dead singular method by name-repetition alone).

## Suggested fix

One of:
1. Give well-known Node.js/framework event-registration patterns (`.on('event', methodRef)`, `EventEmitter` subclass method binding, etc.) a real `calls`-style edge from the registration call site to the bound method, the same way #1771/#1895 did for object-literal value-refs — the most precise fix, but real extraction work.
2. Recognize a narrower, still-safe root signal for methods bound as a callback argument to a recognized emitter API (`.on(`, `.addEventListener(`, `.once(`) even without a matching interface declaration — cheaper, but needs care to avoid reopening the exact "ordinary unused method" gap #2032 just closed.

Either fix needs the equivalent change in both `src/graph/classifiers/roles.ts` and `crates/codegraph-core/src/graph/classifiers/roles.rs` for dual-engine parity.

## Related

- #2032 — introduced the reachability check and `isInterfaceDispatchMethodRoot`'s name-match requirement
- #2257 — a structurally similar but distinct narrow gap (bare function-value logical-or-fallback references, not method-kind event callbacks)

carlos-alm:
## Summary

Closes #2259

`extractCallbackDefinition` already creates a synthetic `event:${eventName}` definition for `.on()`/`.once()`/`.addEventListener()` registrations, spanning just the callback's own body, so the callback gets an entry-point node for root classification. `findEnclosingCallable` picked this synthetic definition as the call's ATTRIBUTED CALLER purely because its span is narrower than the real enclosing method's — but a synthetic definition has no class/`this` context of its own, so a `this.handler()` call inside it (e.g. `w.on('message', (msg) => this.onMessage(msg))` inside a class method) could never resolve through any strategy in the this/self/super cascade, leaving the handler's own callees wrongly classified dead once #2032's reachability downgrade landed.

Confirmed real instance: `WasmWorkerPool.onMessage` in `src/domain/wasm-worker-pool.ts` — its callees (`deserializeResult` etc.) had no path back to a root.

## Fix

Prefer a REAL (non-synthetic) enclosing callable over a synthetic `route:`/`event:`/`command:`-prefixed placeholder for caller attribution, falling back to the synthetic one only when no real callable also encloses the call — the common case of a route/event/command registered directly at module scope, where the synthetic definition is the only candidate and behavior is unchanged.

Mirrored in the Rust engine (`build_edges.rs`); `FRAMEWORK_ENTRY_PREFIXES` made visible from `roles.rs`/`roles.ts` for reuse.

## Test plan

- [x] New unit test on `findCaller`'s tie-breaking (real callable wins; synthetic still wins when it's the only candidate)
- [x] New dual-engine (WASM + native) integration test confirming the real edge is created and the handler's callees stay reachable
- [x] Full test suite passes (4971 tests)
- [x] `cargo test --lib` passes (921 tests)
- [x] Dual-engine parity confirmed (42/42 fixtures)
- [x] No new circular dependencies introduced
- [x] Verified against the real `WasmWorkerPool.onMessage` case in this repo's own `src/`

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
@@ -944,6 +944,7 @@ fn emit_pts_alias_edges<'a>(
             alias_ctx.imported_original_names,
             alias_ctx.namespace_imports,
             &mut alias_confidence_override,
+            None,
         );
         sort_targets_by_confidence(
             &mut alias_targets,
@@ -1357,7 +1358,7 @@ fn process_file<'a>(
             }
         }
 
-        let (caller_id, caller_name) =
+        let (caller_id, caller_name, enclosing_class_hint) =
             find_enclosing_caller(&fc.defs_with_ids, call.line, fc.file_node_id);
         let is_dynamic = if call.dynamic.unwrap_or(false) {
             1u32
@@ -1381,6 +1382,7 @@ fn process_file<'a>(
             &fc.imported_original_names,
             &fc.namespace_imports,
             &mut confidence_override,
+            enclosing_class_hint,
         );
         // #1771/#1784: value-ref references (object-literal property values,
         // Lua builtin reassignment, `instanceof ClassName`) resolve against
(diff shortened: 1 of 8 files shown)


carlos-alm:
@greptileai

Issue #2257, opened 2026-08-04
Function-value logical-or-fallback references lack a real calls edge, unlike object-literal value-refs (#1771)


carlos-alm:
## Summary

#1771 gave object-literal property-value references (`{ resolve: someFn }`-style dispatch tables) a real `calls` edge (`dynamic=1`, `dynamicKind: 'value-ref'`) once invocation evidence is confirmed (#1895). The classifier's `hasActiveFileSiblings`+`fanOut > 0` rescue in `classifyUnreferencedNode` (`src/graph/classifiers/roles.ts`) still has a documented, explicitly-acknowledged fallback for a *different*, not-yet-fixed value-reference shape: a bare function referenced as a logical-or default, e.g.

```ts
const fetchFn = options._fetchLatest || fetchLatestVersion;
```

(the real, pre-existing example is `fetchLatestVersion` in `src/infrastructure/update-check.ts`). This pattern still produces **no edge at all** — the function is only kept out of `--role dead` output via the heuristic rescue, not a real edge, unlike the object-literal case.

## Why this now matters more

#2032 added a transitive-reachability check: a `function`/`method` is only exempted from the new downgrade pass if it's a confirmed-live root (exported, framework-dispatched, Commander dispatch, or a `method`-kind interface-dispatch rescue — see `isLiveRoot`/`isInterfaceDispatchMethodRoot`). Deliberately, a **`function`-kind logical-or-fallback rescue is NOT treated as a root** (see `applyReachabilityDowngrade`'s doc comment) — promoting it to root status would silently rescue genuinely-dead intermediate functions, defeating #2032's own fix in the common case.

One side effect, confirmed on this repo's own self-build: `fetchLatestVersion` has `fanIn === 0` itself (rescued only via the heuristic, not exported), so `collectResponseBody` — called only by `fetchLatestVersion` — is now flagged `dead-unresolved`, even though `fetchLatestVersion` is genuinely invoked in production via the logical-or pattern. A second, structurally identical case: `resolveCalleeNameAndReceiverType` in `src/domain/graph/resolver/ts-resolver.ts`, whose only caller `visit` is itself only rescued the same way.

## Suggested fix

[...]

Issue #1895, opened 2026-07-06
roles --role dead: object-literal property-value references count as liveness without checking if the property is ever invoked (false negative)


carlos-alm:
## Summary

`codegraph roles --role dead` treats a function being referenced as a **value inside an object literal** (e.g. a factory function's returned object) as sufficient evidence of liveness — without tracing whether the resulting object's property is ever actually *invoked* by a consumer. This produces false negatives: genuinely-unreachable functions are not flagged as dead when they're wired into an object literal whose property nobody calls.

## Concrete instance (already fixed)

Found and fixed in #1770: `src/graph/algorithms/leiden/partition.ts`'s `makePartition()` factory returned an object literal like:

```ts
return {
  // ...
  deltaCPM: (v, newC, gamma) => computeDeltaCPM(s, v, newC, gamma),
  deltaModularityDirected: (v, newC, gamma) => computeDeltaModularityDirected(s, v, newC, gamma),
  deltaModularityUndirected: (v, newC, gamma) => computeDeltaModularityUndirected(s, v, newC, gamma),
  getCandidateCommunityCount: (): number => s.candidateCommunityCount,
  // ...
};
```

`computeDeltaCPM`, `computeDeltaModularityDirected`, `computeDeltaModularityUndirected`, and the `getCandidateCommunityCount` closure were genuinely unreachable — nothing anywhere in `src/`/`tests/`/`scripts` ever called `partition.deltaCPM(...)`, `partition.deltaModularityDirected(...)`, `partition.deltaModularityUndirected(...)`, or `partition.getCandidateCommunityCount()` — verified both by repo-wide grep and empirically (replacing each function body with a `throw` and running the full test suite: all 3549 tests still passed, proving zero test or code path reached them). Confirmed via `codegraph roles --role dead -T --file src/graph/algorithms/leiden/partition.ts --json`: none of these were flagged as dead, because each function is referenced as an object-literal property value inside `makePartition`'s return statement, which the detector accepts as sufficient liveness evidence on its own.

## Why this happens

[...]

PR #2445, opened 2026-08-10
fix(resolver): computed dispatch-table access gets invocation evidence

### LINKED ISSUE #2260: Computed-property (bracket-access) dispatch-table lookups lack a real calls edge, unlike dot-property value-refs
## Summary

#1771/#1895 gave object-literal property-value references (`{ resolve: someFn }`-style dispatch tables accessed via `x.keyExpr(...)`) a real `calls` edge once invocation evidence via member-call syntax is confirmed. That mechanism does not cover the **computed/bracket-access** dispatch-table idiom:

```ts
const GROOVY_NODE_HANDLERS: Record<string, (node, ctx) => void> = {
  interface_definition: handleGroovyInterfaceDecl,
  interface_declaration: handleGroovyInterfaceDecl,
  // ...
};

function walkGroovyNode(node, ctx) {
  const handler = GROOVY_NODE_HANDLERS[node.type];
  if (handler) handler(node, ctx);
  // ...
}
```

`handleGroovyInterfaceDecl` (and every other handler in `GROOVY_NODE_HANDLERS`) has `fanIn === 0` — no `calls`/`imports-type` edge points to it at all — because #1895's invocation-evidence check specifically looks for literal member-call syntax (`x.keyExpr(...)`), not a computed lookup (`table[key](...)`) followed by a call through the resulting local variable.

This is a **very common idiom in this codebase** — the comment at `src/extractors/groovy.ts:47-49` even says "Lookup table keyed on node.type, replacing a linear switch dispatch," implying multiple of the 34 language extractors use this pattern for AST-node-type dispatch (tree-sitter's own giant switch/if-chain refactored into a table).

## Why this now matters more

#2032 added a transitive-reachability check to the dead-code classifier. `handleGroovyInterfaceDecl` was previously kept out of `--role dead` output only because its FILE (`src/extractors/groovy.ts`) was incorrectly marked "exported" by a since-fixed bug (see #2032's PR: a named reexport of one function, `export { extractGroovySymbols } from './groovy.js'` in `src/extractors/index.ts`, was wrongly marking every OTHER symbol in that file as reachable too). Once that bug was fixed, `handleGroovyInterfaceDecl`'s real `fanIn === 0` became visible, and — since it's not itself a confirmed-live root by any of #2032's criteria — it and its cascade of internal helpers (`collectGroovyParentInterfaces`, `getFirstStringArgGroovy`, `extractGroovyParams`, `extractGroovyClassMembers`, all called only from within `handleGroovyInterfaceDecl`'s own body) now show up as `dead-unresolved`, even though they are genuinely live (dispatched via the lookup table).

## Suggested fix

Extend the #1771/#1895 value-ref edge mechanism to recognize computed-property dispatch-table lookups: `const handler = TABLE[key]; ...handler(...)` (or the inline form `TABLE[key](...)`) should get the same treatment as `TABLE.key(...)` once invocation evidence is confirmed for the resulting call. Needs the equivalent change in both the TS extractors/resolver and the mirrored Rust engine, per this project's dual-engine parity invariant.

## Known affected symbols (as of #2032)

- `src/extractors/groovy.ts`: `handleGroovyInterfaceDecl` and its downstream helpers (`collectGroovyParentInterfaces`, `getFirstStringArgGroovy`, `extractGroovyParams`, `extractGroovyClassMembers`)
- Likely other language extractors using the same `NODE_HANDLERS[node.type]` lookup-table idiom — not exhaustively audited here; worth a `codegraph roles --role dead -T` sweep across `src/extractors/*.ts` once this is fixed, to confirm the blast radius.

## Related

- #1771, #1895 — the dot-property value-ref mechanism this issue proposes extending to computed/bracket access
- #2257 — a structurally similar but distinct narrow gap (bare function-value logical-or-fallback references)
- #2259 — another structurally similar but distinct gap (EventEmitter-style method callback registrations)
- #2032 — the transitive-reachability fix whose reexport-scoping correction surfaced this

carlos-alm:
## Summary

`{ resolve: someFn }` dot-property dispatch-table references already get liveness-gated `calls` edges once #1895 confirms a real `.resolve(...)` invocation somewhere. That mechanism never covered the computed/bracket-access idiom:

```js
const handler = TABLE[node.type];
if (handler) handler(node, ctx);
```

`src/extractors/groovy.ts`'s `GROOVY_NODE_HANDLERS` dispatch table (and every handler it points to) has `fanIn === 0` under this pattern — nothing pointed to any handler via a syntactic edge — so once #2032's reachability downgrade landed, each handler and its own callees were wrongly classified dead, despite being genuinely reachable through the table.

## Fix

New `collectComputedDispatchTableEvidence`/`handle_computed_dispatch_table_evidence` (TS + Rust) recognizes `const handler = TABLE[computedExpr]` followed by a later call-site reference to `handler` in the same enclosing block — reusing #2257's local, position-scoped liveness-scanning machinery (`hasLaterReferenceInEnclosingBlock`/`has_later_reference_in_enclosing_block`), extended with a new `requireCallSite`/`require_call_site` parameter so the leaf match is restricted to identifiers that are specifically the *callee* of a call expression, matching #1895's "invoked via call syntax" precision.

Because a computed key can't statically name one specific property the way a dot access can, evidence is credited to the **whole table** (keyed on the table's own variable name) rather than per-key — the existing keyExpr gate at `resolveFallbackTargets`/`resolve_call_targets` now passes when either the per-key evidence *or* the table-level computed-access evidence is present.

## Also fixed: a durable-table parity bug found during verification

While verifying this change end-to-end, `tests/integration/issue-1852-watch-cha-pts-dynamic-sink.test.ts`'s object-rest-param suite started failing — a full native build produced an extra, incorrect `calls` edge that an incremental rebuild of the identical source correctly omitted.

Root cause: `persist_invoked_property_names` (`import_edges.rs`) — the native orchestrator's writer for the durable `invoked_property_names` table (#2087) — has its own inline copy of the "what counts as invoked-via-member-call-syntax" filter, rather than calling the shared `collect_invoked_property_names`. That inline copy was missing the `dynamic_kind != "value-ref"` exclusion this PR's own gate change depends on, so a value-ref call's own bare reference polluted the durable evidence table it later gets checked against — self-confirming its own liveness on every subsequent build. Fixed by applying the same exclusion there, matching the already-correct in-memory `collect_invoked_property_names` and the TS-side `persistInvokedPropertyNames` (which already reuses the shared, correctly-scoped helper).

## Testing

- New extractor unit tests (7 TS + 6 Rust) covering: table-name recorded when later called, not recorded when only referenced, not recorded for string-literal keys, not recorded across a shadowing nested scope, resolves through parenthesized/as-const wrappers, isolates evidence per-table (two tables, only one has call evidence), and a pathological 300-deep nesting stack-safety check.
- New dual-engine integration test (`tests/integration/issue-2260-computed-dispatch-table-evidence.test.ts`) against a Groovy-style dispatch-table fixture, asserting the edge exists on both WASM and native engines and that the handler + its callees are not classified dead.
- Full suite: 5008 tests passing (WASM + native).
- `node scripts/parity-compare.mjs`: 42/42 fixtures identical across engines.
- `npm run lint`: clean.

Closes #2260

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
@@ -2243,6 +2243,7 @@ fn build_and_insert_call_edges(
             array_callback_bindings: non_empty(&symbols.array_callback_bindings),
             object_rest_param_bindings: non_empty(&symbols.object_rest_param_bindings),
             object_prop_bindings: non_empty(&symbols.object_prop_bindings),
+            computed_dispatch_table_evidence: non_empty(&symbols.computed_dispatch_table_evidence),
         });
     }
 
--- a/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs
@@ -150,6 +150,12 @@ pub struct FileEdgeInput {
     /// Phase 8.3f: object-property bindings.
     #[napi(js_name = "objectPropBindings")]
     pub object_prop_bindings: Option<Vec<ObjectPropBinding>>,
+    /// Table names (issue #2260) with confirmed LOCAL computed-invocation
+    /// evidence: `const handler = TABLE[computedExpr]; ...; handler(...)`.
+    /// Mirrors `ExtractorOutput.computedDispatchTableEvidence` in
+    /// `src/types.ts` — see its doc comment for the full rationale.
+    #[napi(js_name = "computedDispatchTableEvidence")]
+    pub computed_dispatch_table_evidence: Option<Vec<String>>,
 }
 
 #[napi(object)]
@@ -194,6 +200,14 @@ struct EdgeContext<'a> {
     /// liveness rationale. Owned (not borrowed) since the extra evidence
     /// comes from outside `files`' own lifetime.
     invoked_property_names: HashSet<String>,
+    /// Table names (issue #2260) with confirmed LOCAL computed-invocation
+    /// evidence across every file in this build pass — the alternate
+    /// liveness pathway for a computed/bracket-access dispatch table
+    /// (`const handler = TABLE[computedExpr]; ...; handler(...)`), where a
+    /// computed key can't name a specific property statically the way
+    /// `TABLE.key(...)` can, so evidence is credited to the whole table
+    /// rather than per-key. See `collect_computed_dispatch_table_evidence`.
+    computed_dispatch_table_evidence: HashSet<String>,
     /// CHA + RTA typed-dispatch context (#1949): interface/class name →
     /// concrete classes that implement or extend it, built once per build
     /// pass from every file's `classes` (`extends`/`implements`). Used by
@@ -261,6 +275,7 @@ impl<'a> EdgeContext<'a> {
                 files,
                 extra_invoked_property_names,
             ),
+            computed_dispatch_table_evidence: collect_computed_dispatch_table_evidence(files),
             cha_implementors: cha.implementors,
             cha_implementors_by_file: cha.implementors_by_file,
             cha_parents: cha.parents,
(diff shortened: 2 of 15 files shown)


carlos-alm:
@greptileai

carlos-alm:
@greptileai

carlos-alm:
@greptileai

Issue #2045, opened 2026-07-08
jvm-tracer.sh: java/groovy per-method traceCall() injection regex also matches if/while/for/catch blocks


carlos-alm:
## Bug

While validating the fix for #1913, I got the java and groovy dynamic-call-tracer sed injections to actually execute on BSD sed for the first time (they previously failed with a hard syntax error before even reaching this logic). That surfaced a separate, pre-existing precision bug: the per-method `traceCall()` injection regex in `tests/benchmarks/resolution/tracer/jvm-tracer.sh` matches **any** line ending in `) {` (optionally with whitespace) that isn't a `class `/`interface ` declaration — not just method signatures.

```bash
sedi_append_unless '/\)[[:space:]]*\{$/' '/class |interface /' \
    '        CallTracer.traceCall();' "$javafile"
```

Unlike the kotlin/scala tracers (which anchor on the `fun `/`def ` keyword), java and groovy's per-method pattern has no such anchor, so it also matches `if (...) {`, `while (...) {`, `for (...) {`, `switch (...) {`, and `catch (...) {` lines.

Example (from `tests/benchmarks/resolution/fixtures/java/UserService.java` after injection):

```java
public void createUser(String id, String name, String email) {
    CallTracer.traceCall();
    boolean valid = validator.validateUser(name, email);
    if (!valid) {
    CallTracer.traceCall();          // <-- spurious: same enclosing method, not a real call site
        throw new IllegalArgumentException("Invalid user data");
    }
    ...
}
```

## Impact

Currently harmless at runtime: `CallTracer.traceCall()` records `stack[2]` (the caller's caller) as the edge source, which is the *enclosing method* regardless of whether the call happens at the top of the method body or inside a nested `if`/`while` block, so the spurious call produces an edge with the same key as the one already recorded at method entry — deduplicated by the `seen` Set in `CallTracer.java`/`CallTracer.groovy`. No incorrect edges reach the JSON output today.

It's still worth fixing because:
- It's fragile — any future refactor of `CallTracer`/dedup logic could turn silent duplicates into duplicate or incorrect edges.
- It does unnecessary extra work (redundant `traceCall()` invocations, extra bookkeeping) on every control-flow block in every method.

## Suggested fix

[...]

PR #2449, opened 2026-08-10
fix(tracer): jvm-tracer.sh excludes anonymous-class/closure bodies from traceCall() injection

### LINKED ISSUE #2272: jvm-tracer.sh: java/groovy per-method traceCall() injection can also match anonymous-class/closure bodies
## Bug

Follow-up from #2045. The java/groovy per-method `traceCall()` injection in `tests/benchmarks/resolution/tracer/jvm-tracer.sh` matches any line ending in `) {` that isn't excluded by the negate pattern:

```bash
sedi_append_unless '/\)[[:space:]]*\{$/' '/class |interface |if |while |for |switch |catch |synchronized |try /' \
    '        CallTracer.traceCall();' "$javafile"
```

#2045 fixed the control-flow-keyword false positives (`if`/`while`/`for`/`switch`/`catch`/`synchronized`/`try-with-resources`). A related but distinct shape is still unguarded:

- **Java anonymous class instantiation**, e.g. `new Runnable() {` — matches the outer `) {` pattern, isn't excluded (no literal `class `/`interface ` on that line), and injecting a bare statement directly into a class body (rather than inside one of its method bodies) is invalid Java and would fail compilation if such a fixture existed.
- **Groovy trailing-closure method calls with explicit parens**, e.g. `list.findAll() {` or a DSL-style `withLock() {` — also matches the outer pattern and isn't excluded; the injected `traceCall()` would land inside a closure literal rather than a real method body.

## Impact

No current fixture (`tests/benchmarks/resolution/fixtures/java/*.java`, `tests/benchmarks/resolution/fixtures/groovy/*.groovy`) contains either shape today, so this doesn't currently break the benchmark. It's a latent correctness/robustness gap in the same injection helper, worth tracking separately since fixing it requires different handling (Java: the anonymous-class case is a hard compile error, not a silently-deduped runtime no-op like the #2045 case; Groovy: needs to distinguish a real method declaration from a trailing-closure call expression, which the current negate-list approach can't cleanly express).

## Suggested fix

Investigate whether a more structural approach (e.g. requiring a `<modifiers> <type> <name>(` shape before the `) {`, or tracking brace depth relative to the last `class `/`interface ` line) is needed here, since negate-list extension has diminishing returns for this class of false positive.

carlos-alm:
## Summary

Follow-up to #2045. The java/groovy per-method `traceCall()` injection in `tests/benchmarks/resolution/tracer/jvm-tracer.sh` matches any line ending in `) {` that isn't excluded by the negate pattern. #2045 fixed the control-flow-keyword false positives. Two related shapes still shared that pattern and weren't excluded:

- **Java anonymous class instantiation**, e.g. `new Runnable() {` — injecting a bare statement directly into a class body, rather than inside one of its method bodies, is invalid Java.
- **Groovy trailing-closure method calls with explicit parens**, e.g. `list.findAll() {` or a DSL-style `lock.withLock() {` — the injected `traceCall()` would land inside a closure literal rather than a real method body.

No current fixture triggers either shape, so this doesn't currently break the benchmark — it's a latent robustness gap in the same injection helper.

## Fix

Extends the existing negate-list (matching #2045's own approach — the issue explicitly noted "negate-list extension has diminishing returns," but these two additions are narrow and unambiguous, so a full structural rewrite — brace-depth tracking relative to the last `class`/`interface` line — isn't warranted yet):

- `new[[:space:]]+[A-Za-z_$]` excludes an anonymous-class opening. Java has no default parameter values, so a method's own signature can never itself contain `new SomeType(` — any line matching the outer pattern that also contains this is unambiguously an anonymous-class body, not a method signature.
- `\.[A-Za-z_][A-Za-zA-Z0-9_]*\([^()]*\)[[:space:]]*\{[[:space:]]*$` (Groovy only) excludes a trailing-closure call. A real Groovy method declaration is never written as `<qualifier>.<name>(...) {`, and a dotted, fully-qualified return type (`java.util.List getUsers() {`) still has a space — not a dot — directly before the method name, so it isn't caught by this exclusion.

Both additions are applied to the Java branch; the closure-call exclusion is Groovy-only (Java has no closure-call syntax to exclude).

## Testing

Verified directly against representative snippets (shown below) that:
- The anonymous class's own opening is skipped, but its own inner method is still correctly traced.
- A trailing-closure call is skipped, but the enclosing real method — and a method with a fully-qualified return type — are still correctly traced.
- Confirmed each new test actually catches the regression: reverted the corresponding negate-pattern fragment, confirmed the test fails, then restored it.

Full `javac`/`groovyc` compilation could not be verified end-to-end in this environment (no JRE available locally). Verified instead at the sed-transformation level directly via the shared `tracer-common.sh` helpers — exactly what `tracer-common.test.ts` already does, and for the same reason (toolchain-independent, runs identically regardless of which compilers happen to be installed).

- New tests in `tests/benchmarks/resolution/tracer/tracer-common.test.ts`: 2 new cases (Java anonymous class, Groovy trailing closure), all 7 tests in the file passing.
- Full suite: 5039 tests passing (WASM + native) — after rebuilding the native addon from source, since the npm-installed one in this worktree was stale relative to several recent native-behavior-changing merges and caused unrelated native/parity test failures.
- `node scripts/parity-compare.mjs`: 42/42 fixtures identical across engines.
- `npm run lint` / `shellcheck`: clean (shellcheck's existing info-level notices are all pre-existing and unrelated to this change).

Closes #2272

### DIFF
--- a/tests/benchmarks/resolution/tracer/jvm-tracer.sh
+++ b/tests/benchmarks/resolution/tracer/jvm-tracer.sh
@@ -155,13 +155,17 @@ case "$LANG" in
             # `new[[:space:]]` (#2272) additionally excludes an anonymous class
             # instantiation (`new Runnable() {`), which also shares the "...) {"
             # shape: Java has no default parameter values, so a method's own
-            # signature can never itself contain `new SomeType(` -- any line
-            # matching the outer pattern that also contains `new <Identifier>`
-            # is unambiguously an anonymous-class body opening, not a method
-            # signature (injecting into it would be invalid Java -- a bare
-            # statement directly in a class body -- rather than a distinct
-            # method body to trace).
-            sedi_append_unless '/\)[[:space:]]*\{$/' '/class[[:space:]]|interface[[:space:]]|if[[:space:]]|while[[:space:]]|for[[:space:]]|switch[[:space:]]|catch[[:space:]]|synchronized[[:space:]]|try[[:space:]]|new[[:space:]]+[A-Za-z_$]/' \
+            # signature can never itself contain the token `new` followed by
+            # whitespace -- any line matching the outer pattern that also
+            # contains it is unambiguously an anonymous-class body opening, not
+            # a method signature (injecting into it would be invalid Java -- a
+            # bare statement directly in a class body -- rather than a distinct
+            # method body to trace). Deliberately NOT anchored to an identifier
+            # immediately after the whitespace (`new[[:space:]]+[A-Za-z_$]`)
+            # -- that missed `new /* comment */ Runnable() {} ` (Greptile
+            # review, PR #2449); the bare `new` keyword followed by whitespace
+            # is already the unambiguous signal on its own.
+            sedi_append_unless '/\)[[:space:]]*\{$/' '/class[[:space:]]|interface[[:space:]]|if[[:space:]]|while[[:space:]]|for[[:space:]]|switch[[:space:]]|catch[[:space:]]|synchronized[[:space:]]|try[[:space:]]|new[[:space:]]/' \
                 '        CallTracer.traceCall();' "$javafile"
         done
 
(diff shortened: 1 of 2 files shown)


carlos-alm:
@greptileai

carlos-alm:
@greptileai

PR #2450, opened 2026-08-10
fix(hooks): guard-git.sh checkout -- block no longer false-positives on flags

### LINKED ISSUE #2280: guard-git.sh: checkout -- block pattern false-positives on detach/track/orphan flags and unreached branches
## Summary

`.claude/hooks/guard-git.sh`'s block for reverting files via `git checkout -- <file>` uses a regex that matches `checkout` followed by whitespace then `--`.

This matches *any* `git checkout --flag`, not just the file-reverting form it's meant to catch. Legitimate flags like `--detach`, `--track`, `--orphan`, `--ours`/`--theirs` also start with `--` and get denied even though they don't revert working-tree changes.

Worse, the hook greps the **raw command string** passed to Bash, not the code path that will actually execute. A multi-line script with a conditional branch containing `git checkout --detach origin/main` gets blocked even when that branch's condition is false and the line never runs — the hook has no way to know that, since it's a static text scan over the whole script (in fact, this very issue body triggered the same false positive on first attempt, since it quoted the offending git command literally).

## Repro

Encountered while running `/fixer` (issue #1923 branch-cut step, per the skill's own boilerplate): a script containing a conditional `git checkout --detach origin/main` fallback (only reached if a stale local branch already exists) was blocked outright, even though the branch didn't exist yet and that line never executed.

## Suggested fix

Narrow the regex to require a path-like token after `--` (not another flag) — e.g. require `--` to be followed by whitespace then a non-`-`-prefixed token, or a bare `--` immediately before end-of-segment/end-of-string. At minimum, exclude common non-destructive flags (`--detach`, `--track`, `--orphan`, `--no-track`, `--guess`).

carlos-alm:
## Summary

Closes #2280

`.claude/hooks/guard-git.sh`'s block for `git checkout -- <file>` (reverting files) matched any `checkout --<anything>`, denying legitimate flags that merely share the same prefix — `--detach`, `--track`, `--orphan`, `--no-track`, `--guess`, `--ours`, `--theirs`. None of these revert working-tree changes the way a bare `--` pathspec separator does.

## Fix

Require the trailing `--` to be its own token (immediately followed by whitespace or end-of-segment/string) rather than the start of a longer flag name — the minimal, precise narrowing the issue itself suggested. Applied identically to the `docs/examples/claude-code-hooks/guard-git.sh` mirror copy, since a byte-identical-sync test (#2105) enforces the two staying in lockstep.

The hook's static whole-script text scan (no awareness of untaken conditional branches) is left as-is — that's an inherent tradeoff of the cheap, no-AST design already used throughout this file for every other check, not something in scope for this fix.

## Test plan

- [x] New unit tests in `tests/unit/hook-guard-git-clean.test.ts` cover: still blocks the real revert form (`checkout -- file.txt`, multi-file, bare trailing `--`); allows `--detach`, `--track`, `--orphan`, `--no-track`, `--guess`, `--ours`, `--theirs`; still blocks a real revert chained after an allowed `--detach` invocation.
- [x] Verified the new tests actually catch the regression (temporarily re-applied the old regex and confirmed `--detach` was wrongly denied).
- [x] `shellcheck` on `guard-git.sh`: clean.
- [x] Full test suite (native addon rebuilt from source in this worktree): 311 files / 5047 tests passed.
- [x] `npm run lint`: clean.
- [x] Dual-engine parity (`scripts/parity-compare.mjs`): clean (unaffected — this change touches no engine code).
- [x] `codegraph diff-impact --staged`: no function-level changes detected.

### DIFF
--- a/.claude/hooks/guard-git.sh
+++ b/.claude/hooks/guard-git.sh
@@ -97,13 +97,24 @@ if echo "$NCOMMAND" | grep -qE '(^|[[:space:]]|&&[[:space:]]*)git[[:space:]]+res
 fi
 
 # git checkout -- <file> (reverting files). Requires "--" to be its own
-# token (immediately followed by whitespace or end-of-segment/string), not
-# a longer flag that merely starts with "--" (#2280) — otherwise this also
-# denied legitimate flags sharing the same "checkout --" prefix, like
-# --detach, --track, --orphan, --no-track, --guess, --ours, --theirs, none
-# of which revert working-tree changes the way a bare "--" pathspec
-# separator does.
-if echo "$NCOMMAND" | grep -qE '(^|[[:space:]]|&&[[:space:]]*)git[[:space:]]+checkout[[:space:]]+--([[:space:]]|$)'; then
+# token, not a longer flag that merely starts with "--" (#2280) —
+# otherwise this also denied legitimate flags sharing the same
+# "checkout --" prefix, like --detach, --track, --orphan, --no-track,
+# --guess, --ours, --theirs, none of which revert working-tree changes
+# the way a bare "--" pathspec separator does.
+#
+# Deliberately NOT anchored on literal whitespace-or-end-of-line after
+# "--" (i.e. NOT `--([[:space:]]|$)`): every one of those legitimate
+# flag names starts with a plain letter right after "--", so "anything
+# that isn't a letter" is the correct, safe boundary — and it must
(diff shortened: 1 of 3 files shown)


carlos-alm:
@greptileai

carlos-alm:
@greptileai

carlos-alm:
@greptileai

Issue #1922, opened 2026-07-06
WASM complexity engine skips entire file when every function has a dotted name (Lua module pattern, Go receiver-only files)


carlos-alm:
## Summary

`hasFuncBody()` in `src/ast-analysis/engine.ts` excludes any definition whose `name` contains a dot (`!d.name.includes('.')`), intended to filter out bodyless TS/interface method signatures extracted with dotted names (e.g. `InterfaceName.methodName`). This heuristic is also applied as the **file-level gate** deciding whether the WASM complexity visitor runs at all:

```js
function hasFuncBody(d) {
  return (
    (d.kind === 'function' || d.kind === 'method') &&
    d.line > 0 &&
    d.endLine != null &&
    d.endLine > d.line &&
    !d.name.includes('.')
  );
}
```

```js
const needsWasmComplexity = defs.some((d) => hasFuncBody(d) && !d.complexity);
```

For **module-table languages** (most visibly Lua's idiomatic `local M = {}; function M.foo() ... end; return M` pattern), every top-level function in a file can legitimately have a dotted name and a real multi-line body — these are not signature stubs. When *every* definition in a file is dotted, `defs.some(...)` is false for the whole file, so the WASM complexity visitor never runs and **no function in that file gets complexity data**, even though each one has a real, non-trivial body.

Once the visitor *does* run (triggered by at least one non-dotted def in the same file), `storeComplexityResults` maps results back to definitions purely by line number — it does not re-apply the dotted-name filter — so individual dotted-name functions are computed correctly. The bug is specifically the file-level "does this file need complexity at all" gate.

## Evidence

Discovered while fixing #1782 (Lua missing complexity rules entirely). After adding Lua to `COMPLEXITY_RULES`/`HALSTEAD_RULES`, comparing native vs WASM-built graphs for the same fixtures:

```
$ codegraph complexity --file tests/benchmarks/resolution/fixtures/lua/repository.lua --health -T --json --engine native
functions: 4   (M.delete, M.count, M.save, M.find_by_id)

$ codegraph complexity --file tests/benchmarks/resolution/fixtures/lua/repository.lua --health -T --json --engine wasm
functions: 0
```

Same for `tests/benchmarks/resolution/fixtures/lua/service.lua` (M.create_user, M.get_user, M.remove_user, M.summary) — 4 via native, 0 via WASM.

[...]

Issue #1850, opened 2026-07-06
Duplicated CFG/complexity result-merging logic between ast-analysis/engine.ts and domain/wasm-worker-entry.ts risks drift


carlos-alm:
## Summary

While root-causing #1743 (wrong cyclomatic complexity from the CFG-derived override), I found the buggy logic existed in **two independently-maintained copies**:

1. `src/ast-analysis/engine.ts` — `storeComplexityResults`, `storeCfgResults`, `storeNativeCfgResults`, `overrideCyclomaticFromCfg`, `reconcileCfgCyclomatic`, plus small helpers `indexByLine`/`matchResultToDef`/`hasFuncBody`.
2. `src/domain/wasm-worker-entry.ts` — a near-verbatim duplicate of `hasFuncBody`, `indexByLine`, `matchResultToDef`, `overrideCyclomaticFromCfg`, `storeComplexityResults`, `storeCfgResults` (see the file's own comment: `// ── Per-file visitor setup (db-free version of engine.ts::setupVisitors) ────`).

Both copies had the identical bug (CFG block/edge count overriding the correct AST-derived cyclomatic). Fixing #1743 required editing both files in lockstep — nothing enforces that two independent copies of ~100 lines of non-trivial matching/merging logic stay in sync, and it's easy to fix one and forget the other (which is effectively what happened here, since only one of the two had ever been touched by the PR that introduced the bug's "parity" framing).

## Why they're separate today

`wasm-worker-entry.ts` runs inside a `worker_threads` worker and is intentionally "db-free" / doesn't import from `./parser.js` (worker-safety, per its own docblock). That constraint applies to the *parser/grammar* globals, not to the pure, DB-free result-merging functions (`indexByLine`, `matchResultToDef`, `storeComplexityResults`, `storeCfgResults`, `hasFuncBody`) — those take plain data in and return plain data out, with no DB or parser-global dependency, so they look safe to share.

## Suggested fix direction

Extract the shared, DB-free portion (line-indexing, result-matching, `storeComplexityResults`/`storeCfgResults`, `hasFuncBody`) into a small shared module (e.g. `src/ast-analysis/apply-results.ts`) that both `ast-analysis/engine.ts` and `domain/wasm-worker-entry.ts` import, instead of maintaining two copies. Leave the DB-touching and native-specific pieces (`storeNativeCfgResults`, `reconcileCfgCyclomatic` equivalents) in `engine.ts` only, since the worker never sees native data.

[...]

PR #2452, opened 2026-08-11
fix(ast-analysis): hasFuncBody no longer requires a multi-line span

### LINKED ISSUE #2285: WASM complexity visitor never runs for a file where every function is single-line
## Summary

`setupComplexityVisitorForFile` (`src/ast-analysis/engine.ts`) and the analogous `initWasmParsersIfNeeded`/`setupCfgVisitorForFile` gates decide whether to run the WASM complexity/CFG visitor for a file using:

```ts
const needsWasmComplexity = defs.some((d) => hasFuncBody(d) && !d.complexity);
```

`hasFuncBody` requires `d.endLine > d.line` (a multi-line span) in addition to `!d.bodyless`. If **every** function/method definition in a file happens to have its entire body on one line (a common style for simple getters, one-line guard returns, or C# expression-bodied members `=> expr;`), every definition fails `hasFuncBody`, `needsWasmComplexity` evaluates to `false` for the whole file, and the complexity/CFG visitor is never added to the walk at all — the file gets **zero** complexity/CFG data under the WASM engine, even though every function in it has a real, bodied, computable implementation.

## Reproduction

```csharp
interface IRepo2 {
    bool Save(string id, int value);
    bool SaveOneLine(string id, int value) => value >= 0;
}
class Repo2 : IRepo2 {
    public bool Save(string id, int value) { return true; }
    public bool SaveOneLine(string id, int value) => value >= 0;
}
```

```
$ codegraph where --file Repo2.cs
Symbols: IRepo2:1, IRepo2.Save:2, IRepo2.SaveOneLine:3, Repo2:5, Repo2.Save:6, ..., Repo2.SaveOneLine:7, ...

$ codegraph complexity --engine wasm --json
functions: 0   for anything in Repo2.cs
```

All four real definitions (two abstract signatures aside) are single-line, so the file-level gate never fires and none of them — including the two genuinely bodied methods — get complexity data.

This is a **pre-existing, independent bug** from #2055 (confirmed present before any of that issue's fix — `setupComplexityVisitorForFile` was not touched by that PR). It's the same `hasFuncBody` `endLine > line` heuristic, but misapplied at the file-scoping layer rather than the per-definition merge layer #2055 fixed.

## Suggested fix

The file-level gate only needs to know "does this file contain at least one definition that plausibly has a body and is missing complexity" — the `endLine > line` heuristic is the wrong tool for that, for the same reason #2055 found it wrong for the per-definition merge: it's a proxy for "multi-line", not "has a body". Consider gating on `!d.bodyless && !d.complexity` directly (mirroring the `def.bodyless !== true` fix in #2055) instead of the stricter `hasFuncBody`, or auditing whether `hasFuncBody`'s `endLine > line` requirement is still needed anywhere now that `bodyless` is a reliable direct signal from every extractor (added in #1922).

## Discovery context

Found while stress-testing edge cases for #2055 (WASM complexity wrongly computed for bodyless interface methods) — that PR's own fix is unaffected by this; this is a distinct, pre-existing latent bug in the file-level visitor-setup gate, not introduced by it.

carlos-alm:
## Summary

Closes #2285

`hasFuncBody(d)` required `endLine > line` in addition to `!bodyless` — a proxy for "has a body" from before `bodyless` existed as a reliable direct signal from every extractor (#1922). When every function/method definition in a file happened to have its entire body on one line (a getter, a guard return, a C# expression-bodied member `=> expr;`), every one failed that requirement, so every "does this file need a WASM/native complexity or CFG pass" gate that calls `hasFuncBody` evaluated `false` for the whole file — the file got **zero** complexity/CFG data, even though every function in it has a real, bodied, computable implementation.

## Fix

`storeComplexityResults`/`storeCfgResults` had already worked around this by checking `bodyless` directly instead of calling `hasFuncBody()` (the #2055 fix-of-a-fix), but every *file-level* gate across `ast-analysis/engine.ts`, `features/complexity.ts`, `features/cfg.ts`, and `domain/wasm-worker-entry.ts` still called the broken `hasFuncBody` directly (confirmed via `codegraph fn-impact hasFuncBody` — ~10 call sites across those 4 files, all with the identical bug).

Rather than patching each call site individually with a near-duplicate inline check (the codebase already learned that lesson once — see `apply-results.ts`'s own history, #1850/#1743), this fixes the shared predicate itself: `endLine > line` is removed entirely, since `bodyless` is now the sole signal every caller actually needs. This fixes all ~10 call sites at once, and lets `storeComplexityResults`/`storeCfgResults` collapse back to calling `hasFuncBody()` again instead of duplicating its logic inline.

Pre-existing, independent bug from #2055 (confirmed present before that issue's own fix — the file-level gate was not touched by that PR). The native engine's own inline complexity/CFG computation has no such gate and is unaffected — verified manually against this issue's exact C# repro (native already computed complexity correctly for both single-line methods) — so no native/Rust-side change is needed.

## Test plan

- [x] New integration test (`tests/integration/issue-2285-wasm-complexity-single-line-functions.test.ts`) reproduces the issue's exact C# fixture end-to-end via the WASM engine: both single-line methods get real complexity + CFG data, the bodyless interface stub still correctly gets none.
- [x] Verified the new test (and the underlying bug) via direct CLI repro before/after, and by temporarily reverting the fix and confirming both the CLI repro and the new test fail exactly as the issue describes (`functions: []`), then restoring.
- [x] Updated `tests/unit/apply-results.test.ts`'s `hasFuncBody` suite for the new semantics (a single-line, non-bodyless definition is now `true`).
- [x] `codegraph fn-impact hasFuncBody -T`: confirms the blast radius is exactly the 4 files' gates identified during investigation, nothing missed.
- [x] Full test suite (native addon rebuilt from source in this worktree, dist rebuilt after each source change since the WASM worker thread always loads from compiled `dist/`, not `src/`): 313 files / 5074 tests passed.
- [x] `npm run lint`: clean.
- [x] `codegraph diff-impact --staged`: 3 functions changed, 36 transitive callers across 5 files — matches the intended shared-predicate fix exactly.

### DIFF
--- a/crates/codegraph-core/src/extractors/swift.rs
+++ b/crates/codegraph-core/src/extractors/swift.rs
@@ -346,6 +346,42 @@ fn match_swift_node(node: &Node, source: &[u8], symbols: &mut FileSymbols, _dept
             }
         }
 
+        // A protocol member signature is its own distinct node type — NOT
+        // `function_declaration` — confirmed via tree-sitter-swift's own
+        // node-types.json: `protocol_function_declaration` has no `body`
+        // field at all, while `function_declaration`'s `body` field is
+        // required. Without this arm, every protocol method was silently
+        // dropped from the graph entirely (issue #2285, Greptile review, PR
+        // #2452). `bodyless: Some(true)` unconditionally, since the grammar
+        // guarantees this node type never has a body; skip complexity/CFG
+        // for the same reason `handle_interface_decl` (csharp.rs) does for
+        // interface methods.
+        "protocol_function_declaration" => {
+            let name_node =
+                find_child(node, "simple_identifier").or_else(|| node.child_by_field_name("name"));
+            if let Some(name_node) = name_node {
+                let parent_proto = find_swift_parent_class(node, source);
+                let name = node_text(&name_node, source);
+                let full_name = match &parent_proto {
+                    Some(proto) => format!("{}.{}", proto, name),
+                    None => name.to_string(),
+                };
+                symbols.definitions.push(Definition {
+                    name: full_name,
+                    kind: "method".to_string(),
+                    line: start_line(node),
+                    end_line: Some(end_line(node)),
+                    decorators: None,
+                    complexity: None,
+                    cfg: None,
+                    children: None,
+                    bodyless: Some(true),
+                    content_hash: None,
+                    accessor_kind: None,
+                });
+            }
+        }
+
         "import_declaration" => {
             if let Some(id_node) = find_child(node, "identifier") {
                 let path = node_text(&id_node, source).to_string();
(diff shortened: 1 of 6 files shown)


carlos-alm:
@greptileai

PR #2456, opened 2026-08-11
fix(ast-analysis): LOC comment detection tracks block-comment state

### LINKED ISSUE #2287: LOC comment-line detection misclassifies code lines starting with a comment-prefix character (e.g. pointer deref *ptr =)
## Summary

`computeLOCMetrics()` (TS: `src/ast-analysis/metrics.ts`; native: the LOC-counting logic in `crates/codegraph-core/src/ast_analysis/complexity.rs`) classifies a line as a comment purely by checking whether the *trimmed* line text starts with any of a language's comment prefixes (`trimmed.startsWith(p)`). For any language whose comment prefixes include a bare `*` (used to match `/** ... */` Javadoc-style continuation lines: `* some doc text`), this also matches any **executable** line that happens to start with `*` after trimming — most notably pointer-dereference assignment (`*ptr = 5;`), which several supported languages allow as a statement.

There is no block-comment-state tracking (i.e. "are we currently inside an unterminated `/* ... */` block") — it's a pure per-line, stateless prefix check.

## Reproduction

This already reproduces on **`main`**, for **Rust**, on both engines — completely unrelated to any recent PR:

```rust
fn deref_heavy(ptr: *mut i32) -> i32 {
    unsafe {
        *ptr = 5;
        *ptr = *ptr + 1;
        return *ptr;
    }
}
```

```
$ codegraph complexity --health -T --json --engine wasm
{ "name": "deref_heavy", "loc": 7, "sloc": 5 }   ← 2 lines wrongly counted as comments
$ codegraph complexity --health -T --json --engine native
{ "name": "deref_heavy", "loc": 7, "sloc": 5 }   ← identical, same bug on both engines
```

Both `*ptr = 5;` and `*ptr = *ptr + 1;` are counted as comment lines, undercounting `sloc` and skewing the Maintainability Index input. Rust, C#, and Go all currently use the 4-entry `["//", "/*", "*", "*/"]` prefix list (needed for `/** */`-style doc comments) and all support a `*`-prefixed pointer-dereference/assignment statement shape, so all three are exposed to this same class of misclassification today, independent of language.

## Scope note

Found via Greptile review on #2286 (which extended the same 4-entry prefix list to c/cpp/objc/kotlin/swift/scala for cross-engine/cross-language *consistency* — before that PR, those languages used an inconsistent 2-entry list, which is what #2058 was scoped to fix). Confirmed via the Rust repro above that this exact misclassification already existed on `main`, unrelated to #2286's own changes, for every language that already had the 4-entry list — so it's a pre-existing, shared limitation of the per-line stateless heuristic itself, not something #2286 introduced. Filing as its own issue since fixing it properly means making comment detection block-comment-state-aware (tracking whether a line is inside an unterminated `/* ... */`, rather than trusting a bare `*` prefix), which is a real algorithmic change affecting every C-style/JSDoc-style language on both engines — out of scope for a prefix-list consistency fix.

## Suggested fix

Track block-comment state while scanning a function's lines: once a `/*` is seen without a matching `*/` on the same line, treat every subsequent line as a comment continuation (including bare `*`-prefixed ones) until a line containing `*/` is seen — rather than trusting a bare `*` prefix unconditionally on every line. This must be mirrored identically in both `src/ast-analysis/metrics.ts` and the native Rust LOC-counting logic to preserve engine parity.

carlos-alm:
## Summary

Closes #2287

`computeLOCMetrics()` (TS) and the LOC-counting logic in `compute_all_metrics()` (native) classified a line as a comment purely by checking whether the trimmed line starts with any of a language's comment prefixes. For languages using the 4-entry prefix list (`//`, `/*`, `*`, `*/` — needed to match Javadoc-style continuation lines), a bare `*` prefix matched *unconditionally*, so a pointer-dereference assignment (`*ptr = 5;`) — valid in Rust, C, C#, and Go — was wrongly counted as a comment line, undercounting `sloc` and skewing the Maintainability Index input.

## Fix

Track block-comment state across lines instead of trusting a bare prefix character: a line is only a block-comment continuation while a genuine block-opening line has been seen and the block hasn't closed yet. Split the flat prefix list into:
- single-line prefixes (`//`, `#`, `--`) — trusted unconditionally, as before
- a per-language block-comment-support flag — state-tracked, closing the misclassification

Mirrored identically in both engines (`src/ast-analysis/metrics.ts` / `crates/codegraph-core/src/ast_analysis/complexity.rs`), matching the issue's own scope note.

## Test plan

- [x] New unit tests in `tests/unit/complexity.test.ts`: the issue's own Rust pointer-dereference repro (`comment_lines: 0`), a genuine multi-line block comment still tracked correctly, and a JS-native repro (a multiplication-continuation line starting with `*`) since JS has no pointer-dereference syntax.
- [x] New Rust unit tests in `complexity.rs`: mirrored pointer-dereference and block-comment-tracking cases, plus `is_block_comment_lang` coverage for the full language list.
- [x] New dual-engine integration test (`tests/integration/issue-2287-loc-comment-block-state.test.ts`): the issue's exact repro end-to-end via `buildGraph`, both WASM and native.
- [x] Verified the fix and tests by temporarily reverting `metrics.ts` and confirming the new test fails with the exact wrong `comment_lines: 2` the issue describes, then restoring.
- [x] Full test suite (native addon + dist rebuilt from source in this worktree): 314 files / 5107 tests passed.
- [x] `npm run lint` / `cargo fmt --check`: clean.
- [x] `codegraph diff-impact --staged`: 1 function changed (`computeLOCMetrics`), 11 transitive callers across 5 files — matches the shared-predicate fix scope exactly.

### DIFF
--- a/crates/codegraph-core/src/ast_analysis/complexity.rs
+++ b/crates/codegraph-core/src/ast_analysis/complexity.rs
@@ -2142,22 +2142,28 @@ pub fn halstead_rules(lang_id: &str) -> Option<&'static HalsteadRules> {
     }
 }
 
-/// Comment line prefixes per language, used for LOC metrics.
-pub fn comment_prefixes(lang_id: &str) -> &'static [&'static str] {
+/// Single-line comment prefixes per language, used for LOC metrics.
+///
+/// Deliberately excludes `/*`/`*/`/a bare `*` continuation marker: a bare
+/// `*` also opens a pointer-dereference assignment (`*ptr = 5;`) in every
+/// block-comment language below that supports one, so trusting it as a
+/// standalone comment signal misclassified real code as a comment (issue
+/// #2287). Block-comment lines are instead recognized via explicit
+/// `/* ... */` state tracking in `compute_all_metrics`, scoped to languages
+/// where [`is_block_comment_lang`] returns true.
+pub fn line_comment_prefixes(lang_id: &str) -> &'static [&'static str] {
     match lang_id {
-        // c/cpp/cuda/objc/kotlin/swift/scala all use the same `/** ... */`
-        // block-comment style as JS/Java/C# — the 2-entry list omitted
-        // bare `*`/`*/` continuation lines, undercounting commentLines for
-        // any multi-line Javadoc-style comment (issue #2058).
-        "javascript" | "typescript" | "tsx" | "go" | "rust" | "java" | "csharp" | "c" | "cpp"
-        | "cuda" | "objc" | "kotlin" | "swift" | "scala" | "groovy" => &["//", "/*", "*", "*/"],
         "python" | "ruby" | "r" => &["#"],
-        "php" => &["//", "#", "/*", "*", "*/"],
+        "php" => &["//", "#"],
         "bash" => &["#"],
         "lua" => &["--"],
-        "zig" => &["//"],
-        _ => &["//", "/*", "*", "*/"],
+        _ => &["//"],
     }
+}
+
+/// Languages using `/** ... */`-style block comments (issue #2058, #2287).
+pub fn is_block_comment_lang(lang_id: &str) -> bool {
+    !matches!(lang_id, "python" | "ruby" | "r" | "bash" | "lua" | "zig")
 }
 
 // ─── Merged Single-Pass: Complexity + Halstead + LOC + MI ─────────────────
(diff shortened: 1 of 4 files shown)


carlos-alm:
@greptileai

Issue #2387, opened 2026-08-09
bug(python): no imports edges emitted, and aliased module calls (import x as y; y.f()) resolve to nothing


carlos-alm:
## Summary

Two distinct Python resolution defects, both confirmed on a minimal fixture and independently observed across four Optave Python repos during the org rollout:

1. **Python produces no `imports` edges at all.** File-level dependency analysis (`deps`, `impact`, `map` fan-in/out) is therefore blind for Python.
2. **`import mod as alias` + `alias.func()` produces no `calls` edge**, so the callee is classified `dead-unresolved`. The `from mod import func` form resolves correctly.

## Reproduction

```python
# lib.py
def strip_block(text):
    return text.strip()

def helper():
    return 1
```

```python
# consumer.py
import lib as L
from lib import helper

def run(out):
    helper()
    return L.strip_block(out)
```

```console
$ codegraph build
Native build orchestrator completed: 7 nodes, 8 edges, 2 files

$ # edge kinds in the resulting graph:
    5  contains
    2  parameter_of
    1  calls          <- only helper(); L.strip_block() produced nothing
                      <- and there is no `imports` edge of any kind

$ codegraph where strip_block
f strip_block [dead-unresolved]  lib.py:1
  No uses found

$ codegraph where helper
f helper [core]  lib.py:4  (exported)
  Used in: consumer.py:4
```

`strip_block` is called on line 6 of `consumer.py`. It is reported dead.

For contrast, this repo's own TypeScript self-build produces `imports 1958` and `imports-type 1351` edges, so the `imports` edge kind works — it just never fires for Python.

## Independent confirmation across real repos

| Repo | Observation |
|---|---|
| `data-analytics-pipeline-svc` | 413 Python files, 21,802 edges, **zero `imports` edges**. All 3,092 `calls` edges resolved via `ts-native` only — no contribution from import-aware resolution. `deps` reports `Imports (0)` on files with 8 real `from pipeline...` imports. Uses the PyPA-standard "src layout" (`pyproject.toml` → `packages = ["src/pipeline"]`, `pythonpath = ["src","scripts"]`). |
| `ops-development-tool` | `import blocked_label_sync as bls` then `bls.strip_block(...)` — cross-file caller missing from `codegraph context`. `deps` reports `Imports (0)`; `impact` reports no dependents; `where <module>` finds no symbol. |
| `data-retrieval-storage-svc` | 332 Python files, caller coverage 27.0%. |
[...]

Issue #2217, opened 2026-08-02
Rust crate-root discovery doesn't parse Cargo.toml [[bin]]/[[example]]/[[test]]/[[bench]] path overrides


carlos-alm:
## Context

#2007 added `crate::`/`self::`/`super::` resolution to both engines, including (in a follow-up commit on that PR) recognizing standalone Cargo targets by directory-name convention (`src/bin/foo.rs`, `examples/foo.rs`, `tests/foo.rs`, `benches/foo.rs` each being treated as their own independent crate root).

Greptile correctly noted this convention-based approach cannot identify a target declared in `Cargo.toml` at a **custom** path, e.g.:

```toml
[[bin]]
name = "tool"
path = "custom/location/tool.rs"
```

A file at `custom/location/tool.rs` compiles as its own independent crate (same as a conventional `src/bin/tool.rs`), but `findRustCrateRoot`/`find_rust_crate_root` has no way to know this without reading `Cargo.toml` — it would walk up `custom/location/`'s ancestor chain looking for `main.rs`/`lib.rs` and, if one exists elsewhere in the project, wrongly attribute the custom target's `crate::` paths to that unrelated crate.

## Why this is out of scope for #2007

This is the same class of limitation as the `#[path]` attribute override case already documented in `resolveRustUsePath`'s doc comment (a convention-based, known-files-only resolver cannot model overrides that require reading additional source/metadata beyond the file tree) — not a regression, but an inherent boundary of the current design. Closing it requires:

- Parsing `Cargo.toml` (and any path-dependency/workspace-member `Cargo.toml`s) for `[[bin]]`/`[[example]]`/`[[test]]`/`[[bench]]` sections with an explicit `path = "..."` field.
- Building a custom-target-path → crate-root map from that, and consulting it in `findRustCrateRoot`/`find_rust_crate_root` before falling back to directory-convention.
- A TOML parser dependency on both the TS and Rust sides if one isn't already available for this purpose.

This is a materially different mechanism (metadata parsing, not file-tree convention) from #2007's resolver-logic fix — filing separately per repo scope-discipline.

PR #2458, opened 2026-08-11
fix(watcher): watch mode refreshes package.json exports/workspace caches

### LINKED ISSUE #2290: Watch mode never refreshes package.json exports/workspace caches during incremental rebuilds (both engines)
## Summary

Both engines' incremental/watch-mode rebuild path only populates the `package.json` `exports` cache and the workspace-detection map **once**, at the start of a full build (`setWorkspaces()` in resolve.ts / `pipeline_setup` + the `resolve_imports` batch FFI entry point in resolve.rs) — and never refreshes either cache for the lifetime of a watch session. If a dependency or workspace package's `package.json` changes (its `exports` field, or its workspace membership) while `codegraph build --watch` is running, subsequent incremental single-file rebuilds keep resolving imports against the stale, originally-cached data until the process restarts (i.e. the next full build).

## Confirmed symmetric across engines

- **TS/WASM**: `src/domain/graph/watcher.ts` → `rebuildFile()` (`src/domain/graph/builder/incremental.ts`) has zero calls to `detectWorkspaces()`/`setWorkspaces()`/`clearExportsCache()`. `_exportsCache`/`_workspaceCache` are populated once by the full-build pipeline and never touched again during watch mode.
- **Native**: the single-import FFI entry point used by the incremental rebuild path (`resolve_import` in `lib.rs`) intentionally does not call `reset_workspace_resolved_paths()` or `clear_exports_cache()` — only the batch entry point (`resolve_imports`) and `pipeline_setup` do, both full-build-only call sites.

Both engines share the exact same characteristic, so this is not a native/WASM divergence — a native port mirroring this (rather than clearing on every single incremental call, which would defeat the purpose of the cache and force a `package.json` re-read/re-parse on every file save) is the *correct* parity behavior given the existing design.

## Impact

Low-to-moderate: only affects long-running `--watch` sessions where a dependency's `package.json` changes mid-session (e.g. `npm install` of an updated package, or editing a workspace package's `exports` field) without the user restarting the watcher. The graph will resolve against stale exports/workspace data until the next full rebuild.

## Suggested fix

Watch mode would need to detect `package.json` changes specifically (in addition to source file changes it already watches) and trigger either a full rebuild or a targeted cache invalidation (`clearExportsCache()`/`clear_exports_cache()`, and re-running `detectWorkspaces()`/re-supplying the workspace map) when one changes. This is a watcher-level architecture change affecting both engines equally, not a resolver-level fix.

## Discovery context

Found via Greptile review on #2060 (native package.json `exports`-field resolution port) — flagged against the native single-import entry point specifically, but confirmed via direct inspection that the WASM/JS engine has the identical limitation in its own incremental rebuild path. Filing as its own issue since fixing it means changing watch mode's change-detection scope for both engines, not something in scope for a resolver-parity port.

carlos-alm:
## Summary

Closes #2290

Both engines' incremental rebuild path only populated the `package.json` `exports` cache and the workspace-detection map **once**, at the start of a full build, and never refreshed either for the lifetime of a watch session. If a dependency's or workspace package's `package.json` changed while `codegraph build --watch` was running, subsequent incremental rebuilds kept resolving imports against the stale, originally-cached data until the process restarted.

## Fix

Two changes, deliberately split by cost:

- **`clearExportsCache()`** (cheap: a plain `Map#clear`, now also clearing the native side via a new `clear_exports_cache` NAPI export) is called **unconditionally** in `rebuildFile`, once per rebuild of any file — mirroring the existing Cargo.toml/pyproject.toml precedent (#2217, #2387). A dependency's `exports` field can change independent of any workspace membership, including a plain `node_modules` package the watcher never observes at all (by design).
- **Workspace-map re-detection** (`detectWorkspaces()`/`setWorkspaces()`, potentially expensive in a large monorepo since it re-globs and re-reads every workspace package's manifest) is **gated** on the watcher actually observing a `package.json` change inside the watched tree — detected in both the native `fs.watch` callback and the polling loop, to avoid a per-keystroke re-scan cost.

The native engine has no workspace cache of its own — it's passed the map fresh from the TypeScript side (`getWorkspacesForNative`) on every call — so refreshing it on the TS side alone is sufficient for both engines.

## Test plan

- [x] New unit tests (`tests/unit/watcher-workspace-refresh.test.ts`): a workspace package's `exports` field change, a new workspace package added mid-session, and a plain non-workspace dependency's `exports` change — each locks in the exact stale-resolution bug (asserting the OLD result persists without a refresh) before proving the fix (asserting the NEW result appears after calling `refreshWorkspaceAndExportsCaches`). Plus dedicated tests for the shared `diffMtimes` polling-diff helper.
- [x] Verified every new test by temporarily no-op'ing `refreshWorkspaceAndExportsCaches` and confirming all 3 behavior tests fail with the exact stale/missing result the issue describes, then restoring.
- [x] Existing watcher tests (`watcher-flush.test.ts`, `watcher-rebuild.test.ts`) still pass — no regression to `rebuildFile`'s existing behavior.
- [x] Full test suite (native addon + dist rebuilt from source in this worktree): 316 files / 5118 tests passed.
- [x] `npm run lint` / `cargo fmt --check`: clean.
- [x] `codegraph diff-impact --staged`: 10 functions changed, 8 callers affected across 3 files — matches the watcher-focused scope exactly.

### DIFF
--- a/crates/codegraph-core/src/lib.rs
+++ b/crates/codegraph-core/src/lib.rs
@@ -184,6 +184,19 @@ pub fn clear_python_import_roots_cache() {
     domain::graph::resolve::clear_python_import_roots_cache();
 }
 
+/// Clear the native package.json `exports` cache (issue #2290) — called from
+/// the JS-side `clearExportsCache()` (`resolve.ts`) for the same reason as
+/// the Cargo/Python caches above: `resolve_import`, the single-import entry
+/// point `rebuildFile` uses, does not self-clear per build the way the batch
+/// `resolve_imports` entry point does. Unlike the workspace *map* (which
+/// native never caches itself — it's passed in fresh on every call from the
+/// JS side's own `_workspaceCache`), this `exports` cache is internal to the
+/// native resolver and needs its own explicit clear.
+#[napi]
+pub fn clear_exports_cache() {
+    domain::graph::resolve::clear_exports_cache();
+}
+
 /// Compute proximity-based confidence for call resolution.
 #[napi]
 pub fn compute_confidence(
--- a/src/domain/graph/builder/incremental.ts
+++ b/src/domain/graph/builder/incremental.ts
@@ -28,6 +28,7 @@ import type {
 import { parseFileIncremental } from '../../parser.js';
 import {
   clearCargoTargetOverridesCache,
+  clearExportsCache,
   clearPythonImportRootsCache,
   computeConfidence,
   resolveImportPath,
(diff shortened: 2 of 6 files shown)


carlos-alm:
@greptileai

Issue #2074, opened 2026-07-09
update-graph.sh hook fallback silently no-ops when dist/cli.js hasn't been built yet


carlos-alm:
Found while investigating a Greptile P2 comment on PR #1980.

**The issue:** `.claude/hooks/update-graph.sh`'s fallback branch (used when the `codegraph` binary isn't on `PATH`) now correctly points at `dist/cli.js` (fixed by #1980, which corrected a stale `src/cli.js` reference). But on a genuinely fresh checkout — before `npm run build` has ever run — `dist/cli.js` doesn't exist yet either. Since stderr is redirected to `/dev/null` and the exit code is swallowed (`2>/dev/null && BUILD_OK=1 || true`), the hook silently no-ops with no visible error, identical in symptom to the bug #1980 fixed (graph never rebuilds, contributor gets no feedback).

**Why this is separate from #1980:** #1980's fix is complete and correct for its stated scope (pointing the fallback at the right file once built). This is a narrower, pre-existing edge case — the hook's own header comment states it's "informational only, never blocks," so some silent-failure tolerance is by design. The specific gap is diagnosability: a contributor with a fresh clone gets no signal that they need to run `npm install`/`npm run build` first.

**Suggested fix:** before invoking `node .../dist/cli.js`, check whether the file exists; if not, either skip with a one-line hint (e.g. `echo "[codegraph] run npm run build to enable graph auto-rebuild" >&2`, still non-blocking) or fall back further. Should also cover the case where `node_modules`/`better-sqlite3` native bindings aren't installed yet.

Issue #2075, opened 2026-07-09
sweep skill's Greptile-reply verification silently breaks under zsh


carlos-alm:
The \`sweep\` skill's Step 2g verification script (and the equivalent inline pattern reused ad hoc while running \`/resolve\`) iterates Greptile comment IDs with:

\`\`\`bash
greptile_comment_ids=$(echo "$all_comments" | jq -r '...')
for cid in $greptile_comment_ids; do
  ...
done
\`\`\`

This relies on the shell word-splitting an unquoted multi-line variable on newlines inside a \`for x in $var\` expansion. bash does this by default (IFS includes \n). **zsh does not** — unquoted parameter expansion in zsh is not split on IFS unless \`SH_WORD_SPLIT\` is set. Since this repo's default interactive shell is zsh (confirmed via \`ps -p $$ -o comm=\` → \`/bin/zsh\`), running this exact snippet under zsh collapses all comment IDs into a single loop iteration where \`cid\` is the literal multi-line string \`"id1\nid2\n..."\`. The subsequent \`jq\` call then fails to parse (a real parse error hits stderr), the reply-count lookup silently comes back empty, and the surrounding \`if [ "$reply_count" -eq 0 ]\` guard fails its own numeric comparison — so the "unanswered" array never gets populated and the script reports **all comments answered even when they are not**.

Reproduced directly while resolving PR #1982: the first version of the check printed a visible \`jq: parse error\` yet still concluded "All Greptile comments have replies — safe to re-trigger." Rewriting the loop as \`while IFS= read -r cid; do ... done < <(jq -r ...)\` (portable across bash and zsh) fixed it and correctly counted 1 reply per comment.

**Fix:** update \`.claude/skills/sweep/SKILL.md\`'s Step 2g script (and audit the rest of that skill, plus \`resolve\`'s skill file, for the same \`for x in $(...)\`-over-multiline-output pattern) to use a \`while read\`-based loop instead of \`for x in $multiline_var\`.

**Impact:** this is a false-negative on a safety check whose entire purpose is to block a premature Greptile re-trigger — the exact failure mode the check exists to prevent.

PR #2467, opened 2026-08-11
fix(skills): fixer step 2g defends against a moved queue head

### LINKED ISSUE #2304: fixer skill: orchestrator/sub-agent race on shared state.json+queue.json+worktree corrupts queue (dropped issue #2075 record)
## What happened

While executing issue #2074 as a dispatched /fixer sub-agent (per SKILL.md Phase 2's "Dispatching each issue to a sub-agent" model), the orchestrating session appears to have independently reconciled issue #2074 into `.codegraph/fixer/state.json` / `queue.json` — and performed worktree cleanup (`git checkout main` + fast-forward merge of `origin/main`) in the **same shared worktree** — *while the dispatched sub-agent (this session) was still actively working* (polling post-merge CI for PR #2303, per invariant I8).

Evidence from `git reflog --date=iso` in the shared worktree:

```
6e3a4f09 2026-08-06 09:07:04 -0600: merge origin/main: Fast-forward
70406b7e 2026-08-06 09:07:03 -0600: checkout: moving from fix/issue-2074 to main
4a0173ba 2026-08-06 08:06:11 -0600: commit: test(hooks): skip update-graph diagnostics suite on win32
26ab0420 2026-08-06 07:59:40 -0600: commit: fix(hooks): surface build failures in update-graph.sh ... (#2074)
```

PR #2303 (`Closes #2074`) merged at `2026-08-06T14:39:28Z` (UTC). The `checkout main` + `Fast-forward merge origin/main` at `2026-08-06 09:07:03 -0600` (`= 15:07:03Z`) happened **~28 minutes after the merge, while this sub-agent was still inside step 2f-bis** (foreground-polling the post-merge `CI` workflow run to completion, per invariant I8 — that polling alone took well over that window given the matrix job durations).

## Impact — actual corruption, not just a timing curiosity

Because the orchestrator's own reconciliation (or some other concurrent actor sharing the worktree) had already:
1. appended `{issue: 2074, status: "merged", pr: 2303}` to `state.json`, and
2. shifted `queue.json`'s head past `2074` to `2075`,

...this sub-agent's own step 2g — which (per SKILL.md) reads `ISSUE=$(jq -r '.[0].issue' .codegraph/fixer/queue.json)` fresh at write time rather than using a value captured earlier — read the *already-shifted* queue head (`2075`) and appended a **second, bogus** record: `{issue: 2075, status: "merged", pr: 2303}`. Issue #2075 had never been touched by this sub-agent (no branch, no commits, no PR — confirmed via `gh pr list --head fix/issue-2075` and `gh issue view 2075`, both empty/unrelated). Left uncorrected, this would have permanently marked issue #2075 as done and dropped it from the queue without ever being worked.

This sub-agent caught and repaired the corruption before finishing (removed the bogus `#2075` entry from `state.json`, restored `#2075` to the front of `queue.json`), but the underlying race that produced it is still present in the skill.

## Root cause

SKILL.md's dispatch model (`Phase 2 — Solve and Merge Loop`, "Dispatching each issue to a sub-agent") states the orchestrator "only reads state back off disk between dispatches" and shares one worktree with the sub-agent it dispatches (`no isolation parameter`). In practice here, something wrote to `state.json`/`queue.json` and mutated the shared worktree's git state (`checkout main`, fast-forward merge) **before** the dispatched sub-agent's own step 2g ran — i.e., a second writer acted on the same on-disk queue/state and the same working tree that the "currently dispatched, still-running" sub-agent owns for the duration of its run. The skill has no lock/guard preventing this: 2g's own read of `queue.json`'s head is not compared against anything the sub-agent captured earlier (unlike `dispatching-issue`, which step 2a *does* capture up front specifically to survive `queue.json` being shifted by 2g itself — but that safeguard assumes only 2g itself shifts it, not a second, concurrent writer).

## Suggested fix

Either:
- Make the orchestrator's reconciliation-while-waiting strictly read-only until the dispatched sub-agent has actually returned (no writes to `state.json`/`queue.json`, no `git checkout`/`merge` in the shared worktree) — the sub-agent's own return is already the documented signal for "the orchestrator may now validate/reconcile"; or
- Make step 2g defend against a moved queue head the same way step 2a already defends against it: compare the queue head it's about to record against `.codegraph/fixer/dispatching-issue` (captured before dispatch) and refuse to write / hard-stop if they disagree, instead of trusting `queue.json[0]` blindly at write time.

## Where this was found

Discovered and manually repaired while running the fixer skill for issue #2074 in a dispatched sub-agent (this session). No data was lost — issue #2075 has been restored to the front of `queue.json` and its bogus `state.json` entry removed — but the underlying race is a latent bug that could silently drop a real issue from a future run's queue without any sub-agent around to notice and repair it.

carlos-alm:
## Summary
- Step 2g's state-recording bash block read `queue.json[0]` fresh at write time with no check that it was still the issue this dispatch was actually working on. If some other writer shifted `queue.json` past this issue while 2a-2g was still running for it (e.g. during 2f-bis's I8 post-merge-CI polling), 2g would silently record its outcome under whatever issue now sat at the queue's head — marking an untouched issue as done and dropping it from the queue with no trace of the mistake. Observed once in practice (#2075 dropped this way while a sub-agent was mid-run on #2074) and manually repaired at the time; the underlying race was still present in the skill.
- 2g now compares `queue.json[0]` against `.codegraph/fixer/dispatching-issue` (captured before this issue was ever dispatched, per "Dispatching each issue to a sub-agent") and hard-stops instead of writing when they disagree — mirroring the same defense the validation/reconciliation block already applies after a dispatch returns.

Closes #2304

## Test plan
- [x] New unit tests `tests/unit/fixer-skill-2g-queue-guard-2304.test.ts` — extracts the real 2g bash block from SKILL.md (not a reimplementation) and executes it against a sandboxed `.codegraph/fixer/` directory, covering: the moved-queue-head guard firing, the happy path still working, and the guard staying opt-in when `dispatching-issue` is absent. Confirmed to fail against the pre-fix skill text (revert-and-confirm) and pass against the fix.
- [x] Existing `tests/unit/fixer-skill-2g-queue-guard-2229.test.ts` still passes unchanged.
- [x] Full test suite (320 files / 5140 tests): passing.
- [x] `npm run lint`: clean.

### DIFF
--- a/.claude/skills/fixer/SKILL.md
+++ b/.claude/skills/fixer/SKILL.md
@@ -1148,11 +1148,21 @@ fi
 
 TMP_QUEUE=$(mktemp "${TMPDIR:-/tmp}/fixer-queue.XXXXXXXXXX")
 trap 'rm -f "$TMP_QUEUE"' EXIT
-# Guarded the same way: if this fails after state.json already succeeded above,
-# queue.json[0] is still this (already-resolved) issue — exactly the case Phase 2's
-# own resume-detection note (above step 2a) already filters out on the next run.
-if ! jq '.[1:]' .codegraph/fixer/queue.json > "$TMP_QUEUE"; then
-  echo "ERROR: failed to build updated queue.json — issue #$ISSUE was already recorded as $STATUS in state.json. The next run's resume-detection will filter this already-resolved queue head."
+# Re-validate the head in the same jq invocation that performs the shift
+# (Greptile review, PR #2467): the moved-queue-head guard above and this
+# shift are two separate reads of queue.json, with the state.json write
+# happening in between — a writer that moves the head during that window
+# would otherwise make this unconditional `.[1:]` drop whatever untouched
+# issue is now at the head instead of the one whose outcome was just
+# recorded. Checking identity and shifting in one jq call collapses that
+# gap to the same single-read-then-atomic-rename pattern already used for
+# every other queue.json/state.json write in this skill.
+if ! jq --argjson issue "$ISSUE" '
+    if (.[0].issue // null) == $issue then .[1:]
+    else error("queue.json head changed to \(.[0].issue // "empty") — expected #\($issue)")
+    end
+  ' .codegraph/fixer/queue.json > "$TMP_QUEUE"; then
+  echo "ERROR: failed to build updated queue.json for issue #$ISSUE — either the write itself failed, or the queue's head moved away from #$ISSUE between recording the outcome and shifting the queue (see the jq error above). Issue #$ISSUE was already recorded as $STATUS in state.json — investigate queue.json before proceeding rather than retrying blind."
   exit 1
 fi
 if ! mv "$TMP_QUEUE" .codegraph/fixer/queue.json; then
--- a/.claude/skills/fixer/SKILL.md
+++ b/.claude/skills/fixer/SKILL.md
@@ -1092,6 +1092,20 @@ if [ -z "$ISSUE" ] || [ "$ISSUE" = "null" ]; then
   echo "ERROR: queue is empty — nothing to record. The loop should have exited already."
   exit 1
 fi
+# Defend against a moved queue head (issue #2304): if some other writer shifted
+# queue.json past this issue while 2a-2g was still running for it, queue.json[0]
+# is no longer this issue — it's whatever comes next. Recording under that wrong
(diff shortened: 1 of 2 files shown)


carlos-alm:
@greptileai

PR #2469, opened 2026-08-12
fix(scripts): gc-worktrees.sh uses git branch -D for name+SHA confirmed merges

### LINKED ISSUE #2310: gc-worktrees.sh's --prune-branches uses git branch -d, which refuses genuinely squash-merged branches

carlos-alm:
## Summary
- `--prune-branches` used plain `git branch -d` for every branch `classify()` had already deemed "settled" via authoritative GitHub PR state. `-d`'s own internal check is an ancestry test — under a squash/rebase-merge workflow (this repo's own, confirmed live — #2309) a genuinely-merged branch's tip is never an ancestor of the squash commit on the default branch, so `-d` silently refused to delete it and the branch survived every future `--prune-branches` run indefinitely.
- Switching to `-D` unconditionally for every "settled" branch would introduce a worse bug: `classify()` previously matched a branch to a settled PR by name only. A branch name can be reused after its earlier PR merged or closed — matching by name alone would hand a brand-new, never-merged branch's real commits to `-D`. `classify()` now requires an exact name-**and**-current-tip-SHA match against a MERGED PR before a branch is `-D`-eligible, mirroring housekeep's Phase 4a fix for the identical gap (Greptile review, PR #2311). Every other "settled" case (closed-without-merging, a reused name, or the ancestry-test fallback) still only gets `-d`, preserving its refusal-if-unmerged safety net.
- Also corrects a stale `housekeep/SKILL.md` comment that attributed this name+SHA gap to issue #2310 while describing it as unresolved — it's resolved here, in the same PR.

Closes #2310

## Test plan
- [x] New unit tests `tests/unit/gc-worktrees-classify-2310.test.ts` — extracts the real `classify()` function from the script (not a reimplementation) and exercises it with controlled inputs: open PR, exact name+SHA merge match, name-reuse mismatch, closed-without-merging, no PR (orphan), and both ancestor-test-fallback branches. Confirmed to fail against the pre-fix script (4/8 cases wrongly returned "safe for -D") and pass against the fix.
- [x] `tests/unit/docs-skills-mirror-sync.test.ts` (housekeep skill doc) still passes.
- [x] `bash -n` syntax check: clean.
- [x] Full test suite (320 files / 5141 tests): passing.
- [x] `npm run lint`: clean.

### DIFF
--- a/.claude/scripts/gc-worktrees.sh
+++ b/.claude/scripts/gc-worktrees.sh
@@ -152,13 +152,58 @@ classify() {
   return 1
 }
 
+# Delete a branch classify() already ruled settled, choosing -D vs -d by $3's verdict.
+# Split out of flush() so it can be extracted and unit-tested in isolation, the same way
+# classify() itself is (tests/unit/gc-worktrees-classify-2310.test.ts).
+#
+# $1 = branch name, $2 = HEAD sha captured from `git worktree list --porcelain` BEFORE this
+# call, $3 = classify()'s verdict for this branch (3 or 0 — flush() never calls this for 1/2).
+prune_branch() {
+  local branch="$1" head="$2" verdict="$3"
+  [ "$prune_branches" -eq 1 ] && [ -n "$branch" ] || return 0
+  if [ "$verdict" -eq 3 ]; then
+    # -d, not -D: this branch was classified "settled" but without an exact name+SHA
+    # match against a MERGED PR — either it's a CLOSED-without-merging PR (no landed
+    # commit exists to pin against; deleting its branch with -D could destroy real,
+    # never-merged work), a MERGED PR whose branch name has since been reused for
+    # different commits, or the git-ancestry-test fallback (no gh available, or a
+    # detached worktree). None of these have a more-authoritative signal than git's own
+    # ancestry check, so -d's refusal-if-unmerged behavior is exactly the safety net to
+    # keep here — unlike the exact-merge-confirmed branches below.
+    if git -C "$main_root" branch -d "$branch" >/dev/null 2>&1; then echo "  branch deleted   $branch"; fi
+    return 0
+  fi
+  # -D, not -d: classify() already confirmed this branch's CURRENT tip SHA matches a
+  # MERGED PR's exact recorded head via authoritative GitHub PR state, not git
+  # ancestry. Under a squash/rebase-merge workflow (this repo's own, confirmed live —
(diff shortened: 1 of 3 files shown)


carlos-alm:
@greptileai Addressed the TOCTOU finding: branch deletion now happens inline in a new `prune_branch()` helper called immediately after each branch's own worktree removal (not deferred to a batch loop after the full walk), and the `-D` path re-verifies the branch's current tip SHA against the one `classify()` confirmed immediately before deleting — refusing if it changed. `prune_branch()` is extracted and unit-tested the same way `classify()` already is. Please re-review.

carlos-alm:
@greptileai Addressed: replaced the separate `rev-parse` check + `git branch -D` with a single atomic `git update-ref -d refs/heads/<branch> <sha>` call, so there's no longer a window between confirming the SHA and deleting the ref — git performs the compare-and-delete as one operation. Tests updated to match. Please re-review.

Issue #2471, opened 2026-08-13
analyze_complexity_standalone / build_cfg_standalone report "<anonymous>" for Julia (and R) functions


carlos-alm:
## What

The generic `function_name()` helper in `crates/codegraph-core/src/ast_analysis/engine.rs` (used by the NAPI-exposed `analyzeComplexity`/`buildCfgAnalysis` standalone entry points) resolves a function's display name via:

```rust
fn function_name(node: &Node, source: &[u8]) -> String {
    node.child_by_field_name("name")
        .map(|n| n.utf8_text(source).unwrap_or("<anonymous>").to_string())
        .unwrap_or_else(|| "<anonymous>".to_string())
}
```

This only checks a direct `name` field on the function node itself. It silently falls back to `"<anonymous>"` for any language whose function name is nested deeper in the tree rather than a direct field — which is true for:

- **Julia**: `function_definition` has no `name` field; the name is nested under `signature` → `call_expression` → `child(0)` (see `extractors/julia.rs`'s `signature_call` helper, which already handles this correctly for the main extraction path).
- **R**: `function_definition` also has no `name` field; the name comes from the enclosing `binary_operator`'s LHS (see `extractors/r_lang.rs` and the comment on `dataflowR` in `src/ast-analysis/rules/b3.ts`).

Both languages' main extractors (`JuliaExtractor`, and R's extractor) already implement the correct, non-trivial name resolution — but the generic standalone-analysis helper doesn't reuse that logic, so `native.analyzeComplexity(source, path, 'julia')` / `'r'` returns correct `complexity`/`halstead` metrics but `name: "<anonymous>"` for every function.

## Impact

- Direct callers of the NAPI `analyzeComplexity`/`buildCfgAnalysis` functions for Julia/R source get an unusable `name` field.
- The JS-side native-standalone-analysis fallback path (`src/ast-analysis/engine.ts`'s `runNativeAnalysis` → `storeNativeComplexityResults`/`storeNativeCfgResults`) matches native results back onto `Definition`s primarily **by line number**, with `name` only as a tie-breaker when multiple functions share a line — so this particular fallback path is NOT visibly broken in practice today, but any other consumer relying on the standalone functions' own `name` field is.

## Where discovered

[...]

Issue #2470, opened 2026-08-13
complexity: pattern-matching branch-counting design (haskell/ocaml/fsharp/gleam/clojure/erlang) + verilog + dart (#2312 split)


carlos-alm:
## Context

Split from #2312, which originally bundled five distinct pieces of work. #2312 now covers only the two well-defined engine fixes (Julia operator-by-text matching, Solidity generic-wrapper transparency) plus wiring `COMPLEXITY_RULES`/`HALSTEAD_RULES` for julia and solidity. This issue tracks the remaining three items from #2312's original scope, none of which are mechanical rules ports:

### 1. Pattern-matching branch-counting design (haskell, ocaml, ocaml-interface, fsharp, fsharp-signature, gleam, clojure, erlang)

These languages need a design decision on what counts as a "branch" for pattern-match/guard constructs (an N-arm `match`/`case` expression, guard clauses, etc.) before any `ComplexityRules`/`HalsteadRules` can be written. Per #2312's original body: "needs its own design doc given 'what counts as a branch for an N-arm pattern match' needs to be decided once and applied consistently... rather than re-litigated per language." This should land as a design doc first, then mechanical rules ports for each of the eight languages once the design is settled.

### 2. Verilog

No existing dataflow-rule investigation to build on (unlike r/julia/groovy/solidity, which had prior dataflow work to extend). Needs its own from-scratch grammar investigation — HDL `always`/`case`/`if` semantics differ meaningfully from imperative control flow and don't obviously map onto the existing `ComplexityRules`/`HalsteadRules` shape.

### 3. Dart

Blocked on #2182 (function body isn't a child of `function_signature`/`method_signature` in tree-sitter-dart). Do not attempt until that's fixed.

## Suggested approach

Tackle the pattern-matching design question as its own design doc first (item 1), since the decision needs to be made once and applied to all eight affected languages rather than re-derived per language. Verilog (item 2) and Dart (item 3, once unblocked) can proceed independently once that design work isn't a prerequisite for them specifically.

PR #2472, opened 2026-08-13
fix(ast-analysis): add Julia/Solidity complexity+Halstead rules with dual-engine fixes

### LINKED ISSUE #2312: complexity: Julia and Solidity need engine-level work, not a rules port (#1923 follow-up)
## Context

Follow-up to #1923. That issue's remaining scope (post tier-1 and ObjC, and now R/Groovy) is: dart (blocked on #2182), zig — done, haskell, ocaml, ocaml-interface, fsharp, fsharp-signature, gleam, clojure, julia, erlang, solidity, verilog.

While implementing R and Groovy, two of the remaining languages turned out to need an actual engine change in `src/ast-analysis/visitors/complexity-visitor.ts` (and the mirrored `crates/codegraph-core/src/ast_analysis/complexity.rs` `walk()`), not just a `ComplexityRules`/`HalsteadRules` config port:

### Julia — operators are unclassifiable by node type

tree-sitter-julia wraps **every** binary operator (`+`, `-`, `>`, `==`, `&&`, `||`, ...) in one generic named `operator` leaf node — confirmed by parsing `x > 0 && y > 0`: the tree is `binary_expression(identifier, operator ">", integer_literal)` where the `operator` node's *type* is literally `"operator"` for every operator kind; only its `.text` distinguishes `">"` from `"&&"` from `"+"`.

`classifyLogicalOp()` (TS) / `is_logical_op()` (Rust) currently match on `node.child(1)?.type` against `logicalOperators`. For Julia this would either:
- match nothing (if `logicalOperators` lists `"&&"`/`"||"`, since the type is always `"operator"`), or
- match everything (if `logicalOperators` lists `"operator"`), incorrectly treating every arithmetic/comparison expression as a logical operator and corrupting cyclomatic/cognitive complexity.

Needs: a rules field (e.g. `logicalOperatorsByText: boolean` or a distinct matcher) so the classifier can compare `.text` instead of `.type` when a language's grammar wraps operators generically. Must be mirrored in both engines.

### Solidity — every statement/expression is wrapped generically

tree-sitter-solidity wraps every statement in a generic `statement` node and every expression in a generic `expression` node — confirmed by parsing an if/else-if/else chain: `if_statement`'s "then" branch and its "else if"/"else" branch are **both** reached via a field named `body` (not `consequence`/`alternative`), each pointing to a `statement` node whose *only* child is the real `block_statement` or nested `if_statement`.

This breaks both else-if detection patterns the engine currently supports:
- Pattern A (`else_clause` wrapper, e.g. JS/C#/Rust) — there is no `else_clause` node.
- Pattern C (`alternative` field, e.g. Go/Java) — there is no `alternative` field; both branches share the field name `body`, so `childForFieldName('alternative')` never matches.

`isElseIfNonNesting()` / `classifyPlainElse()` (TS) and their Rust mirrors check `node.parent?.type` and `node.parent?.childForFieldName(...)` directly — since the real nested `if_statement`/block is a *grandchild* (through the generic `statement` wrapper), not a child, these checks silently return false for every else-if link, so each one is scored as if the `else if` had its own extra nesting level, over-counting cognitive complexity for any Solidity function with an else-if chain.

Needs: either (a) a "transparent single-child wrapper" concept in the shared walk so parent/field checks see through generic `statement`/`expression` nodes, or (b) a Solidity-specific field name (`body`) with disambiguation logic that also handles the double-`body`-field case. This is a real engine gap, not just a Solidity oddity — any future grammar with the same "supertype aliasing via a generic wrapper" shape (common in grammars generated from ungrammar/ASDL-style specs) will hit the identical bug.

## Scope

This issue tracks:
1. The two engine enhancements above (Julia operator-text matching, Solidity wrapper-transparency), landed in whichever engine module and mirrored to the other.
2. Wiring `COMPLEXITY_RULES`/`HALSTEAD_RULES` for julia and solidity once the engine supports their grammars correctly.
3. The remaining functional/pattern-matching languages #1923 flagged as needing "real design thought": haskell, ocaml, ocaml-interface, fsharp, fsharp-signature, gleam, clojure, erlang — these need a design decision on what counts as a "branch" for pattern-match/guard constructs before any rules can be written.
4. verilog — no existing dataflow-rule investigation to build on (unlike r/julia/groovy/solidity), needs its own from-scratch grammar investigation (HDL `always`/`case`/`if` semantics differ from imperative control flow).
5. dart — blocked on #2182 (function body isn't a child of `function_signature`/`method_signature` in tree-sitter-dart), do not attempt until that's fixed.

## Suggested approach

Do the two engine fixes first (Julia, Solidity) since they're reusable infra, verified with hand-built fixtures the same way #1923's prior passes did. Then treat julia/solidity as tier-1-equivalent (mechanical rules port once the engine supports them). Tackle the pattern-matching language design question separately — likely its own design doc given "what counts as a branch for an N-arm pattern match" needs to be decided once and applied consistently across haskell/ocaml/fsharp/gleam/clojure/erlang rather than re-litigated per language.

carlos-alm:
## Summary

Julia and Solidity had no `ComplexityRules`/`HalsteadRules` entries in either engine. Wiring them up surfaced two structural bugs that would have silently corrupted metrics for both languages:

- **Julia**: every binary operator (`+`, `-`, `>`, `&&`, `||`, ...) is wrapped in one generic `operator` leaf node — only the token's literal `.text` distinguishes which operator it is, so classifying logical operators and Halstead operator vocabulary by node `.type` alone is impossible. Adds an opt-in `logicalOperatorsByText`/`operatorLeafTypesByText` flag (mirrored in TS and Rust) that switches the comparison to text when set; every other language's `.type`-based comparison is unchanged.
- **Solidity**: `if_statement` has no `else_clause` node and no `alternative` field — both the then- and else-branch bodies are reached via the same `body` field, each wrapped in a generic `statement` supertype-alias node, with a bare `else` token in between. The existing else-if detection (Patterns A/B/C) never recognized this shape, so every Solidity else-if scored as a fresh nested branch instead of a flat +1. A separate `expression` wrapper also broke the same-operator-sequence check for chained `&&`/`||`. Adds a new Pattern D (`transparentWrapperTypes` + `elseKeywordType`, mirrored in Rust) that walks through transparent wrapper nodes to find the real structural parent/sibling.

Both fixes are applied identically across all TS/WASM complexity paths and the Rust native engine, plus wire `compute_all_metrics` into `SolidityExtractor`'s native extractor.

## Scope note

This issue originally bundled five distinct pieces of work. Items 3-5 (pattern-matching branch-counting design for haskell/ocaml/fsharp/gleam/clojure/erlang, verilog, and dart-blocked-on-#2182) need separate design/investigation work and are split into #2470. This PR covers only the two engine fixes plus rules wiring for julia/solidity.

## Out-of-scope finding

Filed #2471: `analyze_complexity_standalone`/`build_cfg_standalone`'s generic `function_name()` helper returns `"<anonymous>"` for Julia and R (pre-existing, unrelated to this fix's correctness).

## Test plan
- [x] Full test suite (`npm test`) — 321 files / 5174 tests pass, zero behavior change for any other language
- [x] `cargo fmt --check` / `cargo test --release` — 988 passed, 0 failed
- [x] Dual-engine parity (`scripts/parity-compare.mjs --langs julia,solidity`) — OK
- [x] New hand-authored fixtures for both languages covering: if/elseif/else, chained logical operators (both bugs' regression cases), while/for loops, try/catch, Halstead vocabulary
- [x] Regression-guard sanity check: disabling each fix individually makes the corresponding new test fail as expected

Closes #2312

### DIFF
--- a/crates/codegraph-core/src/ast_analysis/complexity.rs
+++ b/crates/codegraph-core/src/ast_analysis/complexity.rs
@@ -46,6 +46,12 @@ pub struct LangRules {
     /// (wrapper-unwrapped) node's immediate parent's PRECEDING SIBLING is
     /// this keyword kind — see [`detect_else_if`]/[`is_pattern_d_else`].
     pub else_keyword_type: Option<&'static str>,
+    /// Node kinds to skip when walking backward from a transparent-wrapper
+    /// node looking for the preceding `else_keyword_type` sibling (e.g.
+    /// Solidity's `comment` — `else /* note */ if (...)` still counts as an
+    /// else-if even though a comment node sits between the `else` token and
+    /// the wrapper). Empty for every language without `else_keyword_type` set.
+    pub comment_types: &'static [&'static str],
 }
 
 impl LangRules {
@@ -113,6 +119,7 @@ pub static JS_TS_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 pub static PYTHON_RULES: LangRules = LangRules {
@@ -146,6 +153,7 @@ pub static PYTHON_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 pub static GO_RULES: LangRules = LangRules {
@@ -181,6 +189,7 @@ pub static GO_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 pub static RUST_LANG_RULES: LangRules = LangRules {
@@ -216,6 +225,7 @@ pub static RUST_LANG_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 pub static JAVA_RULES: LangRules = LangRules {
@@ -255,6 +265,7 @@ pub static JAVA_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 pub static CSHARP_RULES: LangRules = LangRules {
@@ -297,6 +308,7 @@ pub static CSHARP_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 pub static RUBY_RULES: LangRules = LangRules {
@@ -335,6 +347,7 @@ pub static RUBY_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 pub static PHP_RULES: LangRules = LangRules {
@@ -378,6 +391,7 @@ pub static PHP_RULES: LangRules = LangRules {
     logical_operators_by_text: false,
     transparent_wrapper_types: &[],
     else_keyword_type: None,
+    comment_types: &[],
 };
 
 // tree-sitter-c's if_statement wraps its else branch in a real `else_clause`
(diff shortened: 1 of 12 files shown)


carlos-alm:
@greptileai Addressed: added \`commentTypes\`/\`comment_types\` to skip over comment siblings when Pattern D walks backward from the transparent wrapper looking for the \`else\` keyword, so \`else /* note */ if (...)\` and \`else\n  // note\n  if (...)\` are no longer misclassified as fresh nested branches. Verified by disabling the fix and confirming it reproduces exactly the inflated cognitive/nesting numbers you predicted. New regression tests in both engines. Please re-review.

PR #2473, opened 2026-08-13
feat(cha): nudge codegraph info when an interface gains its first instantiated implementor

### LINKED ISSUE #2315: incremental CHA post-pass: caller with zero existing dispatch edges never revisited on first implementor (#2078 follow-up)
## Context

Follow-up to #2078 (fixed in #2314), flagged by Greptile during that PR's review.

`findChaSiblingCallerFiles` (in `applyChaDispatchPostPass`, `src/domain/graph/builder/incremental.ts`) discovers additional caller files to revisit by finding existing `technique IN ('cha', 'super-dispatch')` edges pointing at some *other* implementor of a touched interface, then re-parsing those callers' files.

That heuristic requires **at least one existing dispatch edge to seed discovery**. If a caller declares a parameter/variable typed as interface `I`, but at the time its own file was last parsed `I` had **zero instantiated implementors** (so `resolveChaTargets` produced no edges at all for that call site), the caller has no `cha`/`super-dispatch` edge anywhere in the DB. When a later incremental rebuild adds `I`'s *first* instantiated implementor, `findChaSiblingCallerFiles` has nothing to search from — the caller is never discovered, and its dispatch edge to the new implementor stays silently missing until a full rebuild runs.

## Why this needs a different design (not a quick follow-on to #2314's fix)

Every other gap in this area (the original #2078 report, and the transitive-ancestry gap #2314 also fixed) could be closed by searching **existing edges** more thoroughly. This one can't: by construction, there is no edge to search from. Closing it requires one of:

1. **Persist unresolved-but-typed call sites.** Add a DB-level record (new table or column) for "this call site's static receiver type is `I`" independent of whether any edge was ever created for it — then query that instead of `edges`. This is a schema change with build-time and storage cost for every language/every typed call site, not just the CHA-relevant ones.
2. **Re-scan every file for calls typed against the touched interface.** Correct without a schema change, but is exactly the "materially different (and more expensive) post-pass design" the original #2078 report explicitly deferred — it reintroduces an O(all files) cost to the single-file incremental rebuild path, which is the thing incremental rebuilds exist to avoid.
3. **Accept the gap as a documented incremental-mode limitation**, relying on a full rebuild (already recommended after large structural changes, per the README's "if you suspect stale analysis data" guidance) to catch it, and scope this issue to just that documentation + a `codegraph doctor`/build-health check that can flag "an interface gained its first instantiated implementor since the last full build" as a nudge to rebuild.

## Suggested approach

Start with option 3 (cheap, honest, no perf regression risk) and only escalate to option 1 or 2 if a real workload shows this gap matters in practice — matching this repo's convention (see #1936) of not building expensive speculative machinery ahead of a driving need. If picked up, benchmark option 2's actual incremental-rebuild cost first; it may be acceptable if scoped to only the touched interface's *candidate* typed parameters (via a narrower AST query) rather than a full re-parse of every file.

carlos-alm:
## Summary

The incremental CHA post-pass (`findChaSiblingCallerFiles`, `src/domain/graph/builder/incremental.ts`) discovers callers to revisit by following EXISTING `cha`/`super-dispatch` edges from some other implementor of a touched interface. If an interface had zero instantiated implementors when a caller's file was last parsed, that caller has no such edge anywhere in the DB — so when a later incremental rebuild gives the interface its first instantiated implementor, there's nothing to search from, and the caller is never revisited. Its dispatch edge to the new implementor can stay silently missing until a full rebuild.

Per the issue, closing this gap outright needs either a schema change (persist unresolved-but-typed call sites) or an O(all-files) rescan per touched interface — both explicitly deferred pending a proven workload need. This PR implements the recommended cheap alternative instead:

- Snapshots which interfaces/base classes have zero instantiated implementors into a new `build_meta` key (`cha_zero_implementor_interfaces`) at the end of every FULL build — both the JS/WASM `finalize()` path and `native-orchestrator.ts`'s all-Rust fast path (which bypasses `finalize()` entirely when it succeeds — confirmed empirically, hence the second call site).
- `codegraph info` compares that snapshot against a fresh read of the graph and prints a nudge (`Consider: codegraph build --no-incremental`) when one of those interfaces has since gained an instantiated implementor.
- Documents the limitation in `docs/guides/incremental-builds.md` and README's "when to run a full rebuild" section.

This is a diagnostic only — it does not close the underlying gap, per the issue's own scoping.

## Test plan
- [x] Full test suite (`npm test`) — 322 files / 5192 tests pass
- [x] New unit tests for `deriveZeroImplementorInterfaces` (pure function)
- [x] New integration test: full build (interface with an uninstantiated implementor) → incremental rebuild adding a self-instantiating implementor → `codegraph info` asserts the nudge fires, for both engines
- [x] Confirmed `built_at`/`engine`/`codegraph_version`/etc. are unchanged — purely additive

Closes #2315

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,6 +597,7 @@ codegraph build --no-incremental   # Force full rebuild
 
 - **After large refactors** (renames, moves, deleted files) — the reverse-dependency cascade handles most cases, but a full rebuild ensures nothing is stale
 - **If you suspect stale analysis data** — complexity or dataflow results for files you didn't directly edit won't update incrementally
+- **If `codegraph info` nudges you** — it flags when an interface gained its first instantiated implementor since the last full build, a narrow CHA/RTA dispatch-edge gap incremental rebuilds can't self-heal (see [incremental-builds.md](docs/guides/incremental-builds.md#known-limitation-an-interfaces-first-instantiated-implementor))
 - **Periodically** — if you rely heavily on `complexity`, `dataflow`, `roles --role dead`, or `communities` queries, run a full rebuild weekly or after major merges
 - **After upgrading codegraph** — engine, schema, or version changes trigger an automatic full rebuild, but if you skip versions you may want to force one
 
--- a/docs/guides/incremental-builds.md
+++ b/docs/guides/incremental-builds.md
@@ -41,6 +41,28 @@ The most important thing to understand: **complexity, dataflow, and CFG data for
 
 For most workflows — querying callers, checking impact, finding dead code — incremental builds are perfectly accurate. The staleness only matters for analysis-heavy queries (`complexity`, `dataflow`, `cfg`) on files that weren't directly modified.
 
+### Known limitation: an interface's first instantiated implementor
+
+There is one narrower, structural gap around virtual-dispatch resolution
+(CHA/RTA — the `cha`/`super-dispatch` edge techniques). The incremental CHA
+post-pass discovers callers to revisit by following *existing* `cha`/
+`super-dispatch` edges from some other implementor of a touched interface. If
+an interface had **zero instantiated implementors** when a caller's file was
+last parsed, that caller has no such edge anywhere in the graph — so when a
+later incremental rebuild gives the interface its *first* instantiated
+implementor, there's nothing for the post-pass to search from, and the
+caller is never revisited. Its dispatch edge to the new implementor stays
+silently missing until a full rebuild runs.
+
+`codegraph info` detects this automatically: it snapshots which interfaces
+have zero instantiated implementors at the end of every full build, and on
+a later run compares that snapshot against the current graph. If one of
+those interfaces has since gained an implementor, `info` prints a nudge to
+run `codegraph build --no-incremental` (issue #2315). This is a diagnostic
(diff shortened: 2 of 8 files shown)


carlos-alm:
@greptileai Addressed all three findings:

1. **Transitive descendants**: added `collectTransitiveImplementors()`, a BFS mirroring `resolveChaTargets`'s own traversal (keeps walking non-instantiated intermediate classes since they may have instantiated subclasses).
2. **Bare-name collisions**: each `implementorsByFile` entry is now checked as its own disambiguated `${name}|${file}` root; bare names are used only as the same "file identity unknown" fallback `resolveChaTargets` itself falls back to. The write side (`deriveZeroImplementorInterfaces`) and read side (`codegraph info`'s nudge, via the new `hasInstantiatedImplementor`) now share the exact same derivation so they can't disagree.
3. **Inaccurate comment**: corrected the native-orchestrator.ts comment — `ctx.db` can still be a `NativeDbProxy` at that call site, which is fine since the code only relies on the shared `BetterSqlite3Database` interface.

New unit tests cover both the transitive and cross-file-collision cases. Please re-review.

Issue #2476, opened 2026-08-13
Dart: null-aware (conditional) method calls (a?.b()) are never extracted, in either engine


carlos-alm:
## Context

Discovered while implementing #2319 (Dart typeMap population + `call.receiver` wiring for call resolution), while enumerating the selector/call-expression grammar shapes Dart's two extractors need to handle.

Dart's null-aware member access (`a?.b()`) uses a distinct grammar node — `conditional_assignable_selector` (`?.`) — instead of the `unconditional_assignable_selector` (`.`) node ordinary member access/calls use. Confirmed both engines silently extract ZERO calls for this shape:

```dart
void f() {
  a?.b();
}
```

- WASM (`src/extractors/dart.ts`): `resolveDartSelectorCall`'s (formerly `resolveDartSelectorMethodName`) Layout A/B checks only look for `unconditional_assignable_selector`, never `conditional_assignable_selector` — confirmed via `extractDartSymbols(...).calls` returning `[]` for the snippet above.
- Native (`crates/codegraph-core/src/extractors/dart.rs`, tree-sitter-dart 0.2): confirmed the SAME empty-calls result via a scratch unit test during #2319's investigation — tree-sitter-dart 0.2 appears to represent `a?.b()` differently from a plain `.` call in a way `handle_dart_call_expression`'s `member_expression` branch doesn't currently match either (needs its own grammar investigation to confirm the exact node shape 0.2 produces for this).

This is a pure extraction gap (no call is ever recorded at all, so no false edges and no benchmark regression), separate from and unrelated to #2319's typeMap work. Not present in any current resolution-benchmark fixture, so it doesn't move any recall numbers — filed for visibility since it's a real, silent gap in real-world Dart code (null-aware access is common/idiomatic).

## Suggested fix direction

1. Confirm the native engine's exact grammar shape for `a?.b()` (tree-sitter-dart 0.2) via a scratch parse test (same technique used during #2319 — see that issue/PR for the parse-dump harness).
2. WASM: add a `conditional_assignable_selector` branch alongside `unconditional_assignable_selector` in `resolveDartSelectorCall` (mirroring the exact same identifier-extraction logic).
3. Native: add the equivalent handling to `handle_dart_call_expression` / `handle_dart_selector` once the grammar shape is confirmed.
4. Add a fixture case (or extend an existing one) to the Dart resolution-benchmark fixtures so this gap surfaces in `receiver-typed` (or a new `null-aware` mode) recall going forward.

Issue #2475, opened 2026-08-13
Dart: class field 'children' list never populated (dead identifier-scan in extractDartClassMembers)


carlos-alm:
## Context

Discovered while implementing #2319 (Dart typeMap population for field/parameter types).

`src/extractors/dart.ts`'s `extractDartClassMembers` has a "Field declarations" loop meant to push each field into the class definition's `children: SubDeclaration[]` array (used to list a class's properties, e.g. for `codegraph show ClassName`-style output):

```ts
// Field declarations
for (let j = 0; j < member.childCount; j++) {
  const decl = member.child(j);
  if (decl?.type === 'identifier') {
    children.push({ name: decl.text, kind: 'property', line: member.startPosition.row + 1 });
    break;
  }
}
```

This scans `member`'s (a `declaration` node) DIRECT children for a node of type `identifier`. But for every real Dart field-declaration shape (`final Foo x;`, `Foo x;`, `late Foo x;`, `Foo? x;`, `final Foo a, b;`), the field's identifier is nested TWO levels deep — `declaration -> initialized_identifier_list -> initialized_identifier -> identifier` — never a direct child of `declaration` itself. Dart requires every field to carry a `var`/`final`/`const`/`late`/type modifier, so there is no legal field-declaration shape where `identifier` IS a direct child of `declaration`.

Confirmed empirically: parsing `class UserService { final UserRepository _repo; UserService(this._repo); }` and checking `symbols.definitions.find(d => d.name === 'UserService').children` returns `undefined` — the field is silently never recorded.

Net effect: Dart class definitions' `children` array is **always empty for real code** — this loop is dead code that has presumably never fired since Dart support was added.

## Suggested fix direction

Rewrite the loop to walk `initialized_identifier_list -> initialized_identifier -> identifier`, mirroring the (already-correct) same-shape traversal added in #2319's `handleDartFieldDeclTypeMap`/`extractDartDeclaredTypeName`. Also verify/fix the Rust mirror in `crates/codegraph-core/src/extractors/dart.rs`'s `extract_dart_class_methods`, which has the analogous gap (it currently `continue`s on any class_member that isn't a method/bodyless-signature, without ever recording a field child either).

Low-severity — doesn't affect call resolution or the resolution benchmark, only the `children` listing on Dart class definitions — but worth fixing since it's presumably surprising to anyone using `codegraph show` or similar against Dart code today.

Issue #2474, opened 2026-08-13
Dart: local-variable constructor-call typing not seeded (svc.method() unresolved)


carlos-alm:
## Context

Follow-up to #2319, discovered while verifying that fix's resolution-benchmark improvement.

#2319 added typeMap population for Dart's explicitly-typed field declarations and `this.field` constructor-shorthand params, moving the resolution benchmark's `receiver-typed` category from 0/8 to 4/8 (50% recall; overall dart recall 61.9% → 81.0%). The remaining 4/8 false negatives are all local-variable method calls in `tests/benchmarks/resolution/fixtures/dart/main.dart`:

```dart
var repo = UserRepository();
var svc = UserService(repo);   // <-- svc's type is never seeded

svc.createUser('1', 'Alice', 'alice@example.com');  // unresolved
svc.getUser('1');                                    // unresolved
svc.removeUser('2');                                 // unresolved
svc.summary();                                        // unresolved
```

`svc`'s type (`UserService`) is knowable with full certainty from its constructor-call initializer (`UserService(repo)`) — the same "assign a constructor call to a local variable" pattern every other language extractor in this codebase already treats as a distinct, high-confidence (1.0), always-implemented category (e.g. JS/TS's `handleVarDeclaratorTypeMap` for `const p = new Ctor()`, and its Rust mirror `handle_var_declarator_type_map`). Dart's `src/extractors/dart.ts` and `crates/codegraph-core/src/extractors/dart.rs` have no equivalent for `local_variable_declaration` (WASM grammar) / `initialized_variable_definition` (native grammar) nodes.

This was deliberately left out of #2319's scope — that issue's own scope explicitly excluded "type inference from initializers without an explicit type annotation (`var x = Foo();`)" as a blanket exclusion (not limited to fields), to keep the change narrowly scoped to fields/constructor-shorthand params.

## Suggested fix direction

Add a Dart-specific `local_variable_declaration` / `initialized_variable_definition` handler (mirroring JS's `handleVarDeclaratorTypeMap`'s constructor-call branch, confidence 1.0) in both `src/extractors/dart.ts` and `crates/codegraph-core/src/extractors/dart.rs`, seeding `typeMap[varName] = ctorType` when the initializer is a bare/keyword-less constructor call (`var svc = UserService(repo);`). Verify the exact grammar shapes for both engines before implementing — confirmed via tree-sitter-dart parsing during #2319's investigation:

[...]

Issue #1323, opened 2026-06-05
fix(resolver): scope this.prop typeMap key to enclosing class to prevent false edges in multi-class files


carlos-alm:
Deferred from PR #1314 review (Greptile P1).

## Problem

In `handlePropWriteTypeMap` (src/extractors/javascript.ts), constructor property assignments like `this.logger = new Logger(...)` are stored in the typeMap as `this.logger` — without a class-scope qualifier.

When a single file defines two or more classes that both assign the same property name via `this.prop = new DifferentCtor()`, only the first write survives (setTypeMapEntry discards equal-confidence duplicates). Callers of the second class then resolve `this.prop` to the wrong type, producing spurious call graph edges.

## Impact

- Before the PR that seeded this: neither edge was resolved (no false positive)
- After the PR: one class resolves correctly, the other incorrectly — a false positive
- The JS precision gate is at 1.0, so this will surface as a benchmark failure if a multi-class fixture exercises the pattern

## Fix

Key the typeMap entry as `ClassName.prop` (e.g. `Logger.logger`) instead of `this.prop`, where ClassName is the class currently being walked. This requires passing the current class context to `handlePropWriteTypeMap`.

## Example

```js
class A {
  constructor() { this.service = new ServiceA(); }
  doA() { this.service.methodA(); }  // should resolve to ServiceA.methodA
}
class B {
  constructor() { this.service = new ServiceB(); }
  doB() { this.service.methodB(); }  // currently resolves to ServiceA.methodB (wrong!)
}
```

Deferred from PR #1314 review.

PR #2477, opened 2026-08-13
fix(dart): populate typeMap for field/parameter types

### LINKED ISSUE #2319: Dart: no typeMap population for field/parameter types — receiver-typed calls unresolved (#2082 follow-up)

carlos-alm:
## Summary

Dart's TS and Rust extractors never populated `typeMap` at all, so receiver-typed method calls on a class field (`_repo.findById(id)`, where `_repo`'s type comes from a typed field declaration or a `this.field` constructor-shorthand param) could never resolve — confirmed by the resolution benchmark: Dart's "receiver-typed" category sat at 0/8 (0.0% recall) while "static" and "same-file" were both 100%.

Seeds `typeMap` from explicitly-typed field declarations and `this.field` constructor-shorthand params (including the inline-typed shorthand variant, `Ctor(Type this.field)`), mirroring `handleFieldDefTypeMap`'s class-scoped primary + bare-fallback key convention already used for JS/TS. Also wires `call.receiver` extraction into both engines' call-site handling, which Dart never set at all before — the existing receiver-based call resolver had nothing to look up.

Moves Dart's resolution-benchmark receiver-typed category from 0/8 to 4/8 (overall recall 61.9% → 81.0%, precision unchanged at 100%).

## Out of scope (filed separately)

- #2474 — local-variable constructor-call typing (`var svc = UserService(repo)`) not seeded; the remaining 4/8 receiver-typed gap
- #2475 — dead field-`children` scan in `extractDartClassMembers`
- #2476 — null-aware calls (`a?.b()`) never extracted in either engine

## Test plan
- [x] Full test suite (`npm test`) — 322 files / 5211 tests pass
- [x] `cargo fmt --check` / `cargo test --release` — 1003 passed, 0 failed
- [x] Dual-engine parity (`scripts/parity-compare.mjs --langs dart`) — OK
- [x] 17 new unit tests covering field-decl variants (final/late/nullable/generic/multi-identifier), `this.field` shorthand (with and without inline type), and the receiver-wiring change
- [x] Resolution-benchmark threshold ratcheted 0.6 → 0.75 (actual: 0.81), test passes

Closes #2319

### DIFF
--- a/crates/codegraph-core/src/extractors/dart.rs
+++ b/crates/codegraph-core/src/extractors/dart.rs
@@ -462,15 +462,27 @@ fn handle_dart_type_alias(node: &Node, source: &[u8], symbols: &mut FileSymbols)
 /// a primary class-scoped key (`ClassName.field`, confidence 0.9) so two
 /// classes with identically-named fields of different types don't overwrite
 /// each other's entry, plus lower-confidence bare-name fallbacks (`field`,
-/// `this.field`, confidence 0.6) for when the resolver doesn't have a
-/// callerClass in scope. Unlike JS/TS (where every field read requires an
-/// explicit `this.` prefix), idiomatic Dart reads a field with a bare
-/// identifier (`_repo.findById()` inside the SAME class means `this._repo`
-/// implicitly) — the bare fallback key is therefore what call resolution
-/// (`resolveReceiverTypeName` in `src/domain/graph/resolver/strategy.ts`)
-/// actually consults for that shape today; the class-scoped primary key is
-/// seeded for convention-consistency and to serve any future/explicit
-/// `this.field` access (#2319). Mirrors `seedDartFieldTypeMapEntry` in
+/// `this.field`, confidence 0.6) for callers the resolver can't attribute to
+/// a specific class at all.
+///
+/// Idiomatic Dart reads a field with a bare identifier — `_repo.findById()`
+/// inside the SAME class means `this._repo` implicitly, unlike JS/TS, which
+/// requires an explicit `this.` prefix for every field read. This crate's
+/// receiver extraction (`find_dart_selector_receiver` / the `object` field
+/// read in `handle_dart_call_expression`) normalises that implicit shape by
+/// emitting the receiver text itself as `this.<name>` (matching the JS/TS
+/// convention textually), so `resolve_call_targets_core` in
+/// `build_edges.rs` treats a bare Dart field receiver exactly like a JS/TS
+/// `this.field` one: it strips the `this.` prefix and tries the
+/// class-scoped key (`ClassName.field`) FIRST, before ever falling back to
+/// these bare/`this.`-prefixed keys. That is what actually prevents two
+/// classes in the same file from cross-contaminating each other's
+/// same-named field's method resolution (#2319 follow-up on PR #2477's
+/// Greptile finding — see `find_dart_selector_receiver`'s and
+/// `handle_dart_call_expression`'s own doc comments for the
+/// extraction-side half of this fix). The bare/`this.`-prefixed keys seeded
+/// here remain as the fallback for any caller the resolver can't scope to a
+/// class at all. Mirrors `seedDartFieldTypeMapEntry` in
 /// `src/extractors/dart.ts`.
 fn seed_dart_field_type_map_entry(
     symbols: &mut FileSymbols,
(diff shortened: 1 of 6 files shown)


carlos-alm:
@greptileai Addressed: Dart's receiver extraction now emits a bare-identifier receiver as `this.<name>` instead of the bare name — textually matching JS/TS's own `this.field` shape, so the resolver's existing class-scoped-key-first lookup (originally added for #1323/#1458) applies to Dart with zero resolver changes. Applied to both engines' actual live extraction path (confirmed `handle_dart_call_expression` in the Rust extractor is the real live path for the pinned grammar, not the older `find_dart_selector_receiver` — mirrored both for parity). New dual-engine integration test reproduces the exact two-class same-field-name collision and confirms both classes now resolve to their own field's type. Please re-review.

carlos-alm:
@greptileai Addressed: Dart's receiver extraction now detects when a bare-identifier receiver is shadowed by a same-named parameter of the enclosing function/method (excluding the `this.field` constructor-shorthand shape, which doesn't introduce a new binding) — when shadowed, it emits the bare name instead of `this.<name>`, skipping the class-scoped field lookup entirely. A function-scoped typeMap entry is also seeded for the parameter's own type (mirroring the existing `setScopedTypeMapEntry`/#2235 mechanism already used by JS/TS) — this was necessary because skipping the prefix alone still fell through to the field's own bare fallback key. Local-variable shadowing (not just parameter shadowing) is a materially bigger problem, deliberately out of scope, and tracked in #2478. New dual-engine test coverage confirms the shadowed call resolves to the parameter's type, never the field's, and the existing cross-class-collision fix/test still passes. Please re-review.

Issue #1771, opened 2026-07-04
Dispatch-table function references (resolve: fn) inconsistently flagged dead-unresolved depending on unrelated fanOut


carlos-alm:
## Context

Found during Titan-run grind phase 24 (`/titan-grind --phase 24`) while triaging dead symbols in `src/ast-analysis/visitor-utils.ts` (forge commit ce843532, which introduced a `PARAM_NODE_HANDLERS: ParamNodeHandler[]` node-type -> handler dispatch table for `resolveParamNode`).

## Finding

`codegraph roles --role dead -T --file src/ast-analysis/visitor-utils.ts --json` flags 4 of the 5 handler functions referenced by the table as `dead-unresolved`, despite all 5 being referenced identically:

```ts
const PARAM_NODE_HANDLERS: ParamNodeHandler[] = [
  { matches: ..., resolve: (node) => ({ names: [node.text] }) },
  { matches: ..., resolve: resolveWrapperParam },        // flagged dead-unresolved
  { matches: ..., resolve: resolveDefaultParam },        // flagged dead-unresolved
  { matches: ..., resolve: resolveRestParam },           // flagged dead-unresolved
  { matches: ..., resolve: resolveObjectDestructParam }, // NOT flagged (role: leaf)
  { matches: ..., resolve: resolveArrayDestructParam },  // flagged dead-unresolved
];
```

All five are invoked identically at runtime via `handler.resolve(node, rules)` inside `resolveParamNode`'s loop. `fn-impact` confirms all five have `totalDependents: 0 / direct: 0 / transitive: 0` (i.e. codegraph creates **zero** graph edges for any of them when referenced as an object-literal property value) -- yet only `resolveObjectDestructParam` gets rescued from the dead bucket.

## Root cause

`src/graph/classifiers/roles.ts` `classifyUnreferencedNode()`:

```ts
if (node.kind === 'function' && node.fanOut > 0) {
  // Functions referenced as logical-or fallback defaults ... appear as
  // value references, not call sites, so no call edge is produced. We
  // require fanOut > 0 as evidence that the function is non-trivial
  return 'leaf';
}
```

[...]

Issue #1741, opened 2026-07-02
bug(resolver): identifier-arg dynamic call resolution fabricates false edges/cycles on name collision (e.g. communities.ts phantom cycle)


carlos-alm:
## Summary

`extractCallbackReferenceCalls` (and its native mirror) unconditionally emits a `dynamic` call edge for **every bare identifier argument** passed to any call expression, with no gating on the callee — unlike `member_expression` args, which are gated by `CALLBACK_ACCEPTING_CALLEES`/`HTTP_VERB_CALLEES` (fixed in #1190). When the identifier's name happens to collide with an unrelated exported function elsewhere in the repo, the resolver's global-fallback confidence scoring (0.5, same-grandparent-dir heuristic) binds the two together, fabricating a call edge — and, if it closes a loop, a phantom architectural cycle.

This is not hypothetical: it just produced a false cycle in codegraph's own source during a Titan Paradigm SYNC/FORGE run.

## Repro (on this repo)

`src/features/communities.ts` has:
```ts
function analyzeDrift(communities: CommunityObject[], communityDirs: ...): DriftResult {
  ...
  const mergeCandidates = findMergeCandidates(communities);   // line 136 — `communities` is a parameter, passed as a plain arg
  ...
}

export function communitiesData(...): Record<string, unknown> {
  const { communities, communityDirs } = buildCommunityObjects(...);
  const { splitCandidates, mergeCandidates, driftScore } = analyzeDrift(communities, communityDirs); // line 208 — `communities` is a local var, passed as a plain arg
  ...
}
```

Neither function contains a `communities(...)` call expression anywhere — `communities` is only ever a parameter/local-variable name here. But the only *function* named `communities` in the repo is the unrelated CLI entry point `export function communities(...)` in `src/presentation/communities.ts:89`.

`codegraph cycles --functions -T --json` reports:
```
analyzeDrift|src/features/communities.ts
  -> communitiesData|src/features/communities.ts
  -> communities|src/presentation/communities.ts
  -> (back to analyzeDrift)
```

Querying the DB directly confirms both edges closing the loop are fabricated:
```
analyzeDrift    -> communities   kind=calls confidence=0.5 dynamic=1 technique=ts-native dynamic_kind=null
communitiesData -> communities   kind=calls confidence=0.5 dynamic=1 technique=ts-native dynamic_kind=null
```//
[...]

PR #2529, opened 2026-08-15
fix(js/ts): resolve JSX elements and call-arg identifiers as references (#2389)

### LINKED ISSUE #2389: bug(js/ts): JSX elements and call-argument identifiers produce no reference edge, causing false dead-code in React and NestJS
## Summary

Two more members of the known "value-position reference produces no edge" family (#2257, #2260), both confirmed on minimal fixtures. Each causes systematic false dead-code on a framework Optave uses heavily.

1. **A JSX element is not a reference to its component.** `<Header />` produces no edge to `Header`.
2. **An identifier passed as a call argument is not a use.** `Factory.create(AppModule)` produces no edge to `AppModule`.

## Reproduction 1 — JSX

```jsx
// comp.jsx
export function Header({ title }) { return <h1>{title}</h1>; }
export function plainFn() { return 42; }
```

```jsx
// app.jsx
import { Header, plainFn } from './comp.jsx';
export function App() {
  const n = plainFn();
  return <Header title={"x" + n} />;
}
```

```console
$ codegraph where Header
f Header [entry]  comp.jsx:1
  No uses found          <- used at app.jsx:4 as <Header />

$ codegraph where plainFn
f plainFn [core]  comp.jsx:3  (exported)
  Used in: app.jsx:2      <- ordinary call resolves fine
```

The only difference between the two symbols is call syntax versus JSX syntax.

## Reproduction 2 — identifier as call argument

```ts
// mod.ts
export class AppModule {}
export function helper() { return 1; }
```

```ts
// main.ts
import { AppModule, helper } from './mod.js';
const Factory = { create(m: unknown) { return m; } };
export function bootstrap() {
  helper();
  return Factory.create(AppModule);
}
```

```console
$ codegraph where AppModule
* AppModule [leaf]  mod.ts:1
  No uses found          <- passed as an argument at main.ts:5

$ codegraph where helper
f helper [core]  mod.ts:2  (exported)
  Used in: main.ts:3
```

Note this is not a decorator-metadata or reflection limitation — it is a plain identifier in argument position.

## Impact

Both patterns are structural, not edge cases:

- **Every React codebase.** A component referenced only as JSX — which is the normal way to use one — reads as dead. On `client-analytics-reports-poc`, spot-checking 3 of 135 `dead-unresolved` symbols found all three (`Icon`, `Header`, `RingStat`) used exclusively as JSX elements. It also depresses caller coverage across the board.
- **Every NestJS backend.** `NestFactory.create(AppModule)` and the module/controller registration pattern rely on passing classes as values. On `interface-autopilot-web-app`, nearly every `*.module.ts` / `*.controller.ts` export reads as dead; `interface-analytics-web-app` reports 1206 dead symbols heavily attributed to this.

Optave has four React or NestJS repos in the current rollout, and their `.codegraph/basics.md` health baselines have to carry a "dead-code count is not meaningful here" caveat because of these two defects.

## Suggested fix

Treat a JSX opening/self-closing element name as a reference to the resolved binding, and treat a bare identifier in call-argument position as a value reference — the same treatment #2257 and #2260 propose for logical-or fallbacks and computed-property dispatch lookups. Given four sibling issues now, a single "identifier in value position" resolution pass may be a better fix than four targeted patches.

Mirror in `crates/codegraph-core/` for engine parity.

## Related

- #2257 — function-value logical-or-fallback references lack a calls edge
- #2260 — computed-property dispatch-table lookups lack a calls edge
- #2259 — event-callback registrations aren't reachability roots
- #2177 — false dead-code from unresolved cross-directory calls

carlos-alm:
## Summary

Two members of the "value-position reference produces no edge" family (#2257, #2260):

1. **A JSX element is not a reference to its component.** `<Header />` produced no edge to `Header`.
2. **An identifier passed as a call argument is not a use.** `Factory.create(AppModule)` produced no edge to `AppModule`.

Both are structural, not edge cases: every React component used only as JSX read as dead, and NestJS's module/controller registration pattern (`NestFactory.create(AppModule)`) relies entirely on the second pattern — depressing caller coverage and inflating dead-code counts across every React/NestJS repo in the org rollout.

## Changes

Both patterns are extracted as `value-ref` dynamic calls — the existing mechanism already used for object-literal property values, `instanceof` operands, and logical-or/ternary fallbacks (#1771/#1895/#2257):

- **JSX element reference** (`handleJsxElementRef`): a JSX opening/self-closing element's tag name is credited as a reference to the component it renders, gated on JSX's own capitalization convention — lowercase (`<div>`, `<span>`) is an intrinsic DOM element, never a symbol reference; a `member_expression` name (`<Namespace.Component />`) credits the base object identifier.
- **Call-argument identifier reference** (`extractCallArgumentIdentifierRefs`): a **capitalized** bare identifier passed as a call argument is credited as a reference to whatever it names.

### Why capitalized-only for call arguments

This restriction isn't stylistic — it's load-bearing. Issue **#1741** is an existing regression guard proving that crediting an arbitrary lowercase DATA argument (e.g. `analyzeDrift(communities, communityDirs)`) as any kind of reference risks the global-fallback resolver binding it to an unrelated same-named function elsewhere in the repo, fabricating a call edge and, transitively, a phantom cycle. My first pass at this fix did exactly that and broke 13 existing tests locally before I caught it. A class/component reference passed by value is overwhelmingly PascalCase in JS/TS convention, so restricting to capitalized identifiers satisfies #2389's request without reopening #1741 — verified all 13 previously-broken tests pass again with this restriction in place.

### Engine parity

Mirrored in `crates/codegraph-core/src/extractors/javascript.rs`.

The WASM query-based extraction path required new tree-sitter query patterns scoped to *only* the `javascript`/`tsx` grammars — plain `.ts` (no JSX) has no JSX node types at all, so folding them into the shared TypeScript pattern set would throw at `Query()` compile time and break all `.ts` parsing. These patterns are duplicated in both `src/domain/parser.ts` and the isolated `wasm-worker-entry.ts` (which intentionally keeps its own copy to avoid importing `parser.ts`'s non-worker-safe process-global caches) — I initially only updated `parser.ts` and the fix silently didn't work through the real WASM worker pool. A new query-vs-walk parity test case for a capitalized JSX tag catches exactly this class of omission; verified it fails against the pre-fix `wasm-worker-entry.ts` before confirming it passes with the fix.

## Verification

- `cargo fmt -- --check` / `cargo clippy --workspace --all-targets -- -D warnings`: pass
- `npx vitest run tests/parsers/javascript.test.ts`: 382/382 pass (8 new tests added, all 13 previously-regressed callback-extraction tests still pass with the capitalization restriction)
- `npx vitest run tests/engines/query-walk-parity.test.ts`: 25/25 pass (3 new JSX/call-argument parity cases added; verified they fail against the pre-fix `wasm-worker-entry.ts` and pass against the fix)
- `cargo test --package codegraph-core extractors::javascript`: 299/299 pass (6 new Rust tests added, including a regression guard mirroring #1741)
- `npx vitest run tests/parsers/ tests/engines/`: 1120/1128 pass (8 skipped, pre-existing)
- `npx vitest run tests/integration/`: 1483/1487 pass (2 skipped, 2 todo, pre-existing)
- `npm run lint`: pass
- Manually reproduced both of the issue's exact fixtures against a freshly built native addon (`napi build`) and the WASM engine: `Header`/`AppModule` now both show "Used in: ..." on both engines, with identical node/edge counts

Closes #2389

### DIFF
--- a/crates/codegraph-core/src/extractors/javascript.rs
+++ b/crates/codegraph-core/src/extractors/javascript.rs
@@ -1756,6 +1756,9 @@ fn match_js_node(
         "enum_declaration" => handle_enum_decl(node, source, symbols),
         "lexical_declaration" | "variable_declaration" => handle_var_decl(node, source, symbols),
         "call_expression" => handle_call_expr(node, source, symbols, callback_param_shapes),
+        "jsx_opening_element" | "jsx_self_closing_element" => {
+            handle_jsx_element_ref(node, source, &mut symbols.calls)
+        }
         "new_expression" => handle_new_expr(node, source, symbols),
         "decorator" => handle_decorator(node, source, symbols),
         "import_statement" => handle_import_stmt(node, source, symbols),
@@ -2874,6 +2877,9 @@ fn handle_call_expr(
                 callback_param_shapes,
                 &mut symbols.calls,
             );
+            symbols
+                .calls
+                .extend(extract_call_argument_identifier_refs(node, source));
             return;
         }
     }
@@ -2884,6 +2890,9 @@ fn handle_call_expr(
         symbols.definitions.push(cb_def);
     }
     extract_callback_reference_calls(node, source, callback_param_shapes, &mut symbols.calls);
+    symbols
+        .calls
+        .extend(extract_call_argument_identifier_refs(node, source));
 }
 
 fn handle_new_expr(node: &Node, source: &[u8], symbols: &mut FileSymbols) {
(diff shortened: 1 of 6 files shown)


carlos-alm:
@greptileai please re-review — addressed the query-path callee-shape coverage gap by routing call-argument value-ref extraction through a generic (call_expression) capture instead of duplicating it per callee shape.

Issue #2392, opened 2026-08-09
bug(roles): Python entrypoints (__main__ guard, __main__.py) not classified as entry


carlos-alm:
## Summary

Role classification does not recognize either of Python's two canonical entrypoint conventions — the `if __name__ == "__main__":` guard, and a `__main__.py` module. Functions invoked that way are classified `utility`, and repos whose only entrypoints are Python report zero `entry` symbols.

## Reproduction

```python
# app.py
def helper():
    return 1

def main():
    return helper()

if __name__ == "__main__":
    main()
```

```python
# pkg/__main__.py
def run():
    return 2

if __name__ == "__main__":
    run()
```

```console
$ codegraph build
Native build orchestrator completed: 6 nodes, 7 edges, 2 files

$ codegraph roles --role entry
No classified symbols found. Run "codegraph build" first.

$ codegraph where main
f main [utility]  app.py:4
  Used in: app.py:0
```

Both `main` and `run` are genuine program entrypoints. Neither is classified `entry`.

Note `Used in: app.py:0` — the module-scope call is recorded at synthetic line 0. If root detection requires a caller at a real source position, that synthetic location may be why the guard-invoked call doesn't promote the function to a root.

(The "Run codegraph build first" text on a fully-built graph is a separate defect, #2390.)

## Confirmed on real repos

- `data-ingestion-pipe` — textbook `if __name__ == "__main__": main()` at `app/oio.py:1786`; zero `entry` symbols repo-wide, `main()` tagged `utility`.
- `data-retrieval-storage-svc` — `src/data_plane/__main__.py` is the documented container entrypoint (`python -m data_plane`); not classified `entry`. Its CDK entry `infra/cdk/bin/data-plane.ts` lands in `dead-leaf`.
- `data-analytics-pipeline-svc`, `ops-development-tool` — both report zero `entry` symbols.

## Impact

Two consequences beyond the empty `entry` listing:

1. **`codegraph roles --role entry` is unusable for orienting in a Python repo** — which is precisely the "where does this start?" question a new agent asks first. The `.codegraph/basics.md` files being written across the org have to source entrypoints by reading `pyproject.toml` and the source by hand instead.
2. **It likely feeds the dead-code inflation.** If entrypoints aren't roots, transitive reachability from them never runs, so everything reachable only from `main()` is a candidate for a dead verdict. That would compound #2387 (Python emits no `imports` edges) on the same codebases.

## Suggested fix

[...]

PR #2535, opened 2026-08-16
fix(roles): recognize pyproject.toml console-script entrypoints

### LINKED ISSUE #2408: feat(roles): recognize pyproject [project.scripts] console-script entrypoints
## Summary

Follow-up to #2392, which was explicit that this is the third Python entrypoint convention and a natural next step.

#2392 covered the two source-level conventions — the `if __name__ == "__main__":` guard and a `__main__.py` module — by flagging the call site in the extractor and setting `nodes.entrypoint` on whatever it resolves to. A console script declared in `pyproject.toml` has no call site at all, so it is still missed:

```toml
[project.scripts]
ingest = "pipeline.cli:main"
```

`pipeline/cli.py`'s `main` is the installed command's entrypoint, but nothing in the repo calls it, so it classifies by fan-in shape (or lands in `dead-*`) rather than `entry`.

## Suggested scope

Parse `[project.scripts]` and `[project.gui-scripts]` from `pyproject.toml`, resolve each `module:attr` target through the same Python module resolution added in #2387 (which already handles the src-layout and pyproject-declared roots), and set `nodes.entrypoint` on the resolved symbol.

Poetry's `[tool.poetry.scripts]` uses the same `module:attr` value shape and is worth covering in the same pass.

Mirror in `crates/codegraph-core/` — the flag is read by both engines' role classifiers.

## Related

- #2392 — the two source-level conventions, which introduced `nodes.entrypoint`
- #2387 — the Python module resolution this would reuse to resolve `module:attr`

carlos-alm:
## Summary

Closes #2408

A Python project whose only program entrypoint is a packaging-declared console script (`[project.scripts]`, `[project.gui-scripts]`, or `[tool.poetry.scripts]` in `pyproject.toml`) had neither an `if __name__ == "__main__":` guard nor a `__main__.py`, so #2392's role classification still reported zero `entry` symbols for it — exactly the "where does this start?" question a new agent asks first.

- `pyproject.toml` is re-parsed fresh on every build (no per-file evidence table needed, unlike guard-call attribution) — it's a single, cheap-to-reread file, so attribution runs unconditionally each build and self-corrects when the declared scripts change.
- Attribution is scoped to `entrypoint_source_file = 'pyproject.toml'`, so it takes precedence over — but never clobbers — a target already attributed by guard-call evidence.
- Resolution reuses the existing `resolvePythonImportPath`/`resolve_python_import_path` machinery via a synthetic `<rootDir>/pyproject.toml` anchor, so `src/` layouts, flat layouts, and `pyproject.toml`-configured import roots all resolve correctly with no new resolution logic.
- Mirrored in both engines: TS (`resolve.ts`, `builder/entrypoints.ts`, `builder/stages/build-edges.ts`, `builder/incremental.ts`) and Rust (`resolve.rs`, `builder/entrypoints.rs`, `builder/pipeline.rs`).
- While building this out I found and removed a stale per-`root_dir` cache (`_pyprojectScriptsCache` / `pyproject_scripts_cache`) that had no repeat-call benefit (this resolver runs at most once per build) but did introduce real staleness across repeated `buildGraph`/native-pipeline calls in the same process — removed rather than patched, since the function is already cheap enough to just re-read every call.
- Also fixed an unrelated stray NUL byte that had crept into a template literal in `resolve.ts`, which was silently making every `grep`/`ugrep` search against that file return nothing (file treated as binary).

## Test plan
- [x] New Rust unit tests: `resolve_pyproject_script_entrypoints` (6 tests: src-layout resolution, gui/poetry tables, table-shaped-entry skip, dedup, missing/no-scripts, unresolvable module) and `apply_pyproject_script_attribution` (6 tests: marks, clears stale, no-clobber, precedence, empty case)
- [x] New TS unit tests mirroring the above in `tests/unit/resolve.test.ts`
- [x] New dual-engine (`wasm`/`native`) integration test `tests/integration/issue-2408-pyproject-script-entrypoints.test.ts`: classification for all three script tables, non-clobbering, precedence, and incremental-rebuild staleness clearing
- [x] Manually verified native ⇄ WASM parity on a scratch fixture (`entrypoint=1`, `role='entry'`, `entrypoint_source_file='pyproject.toml'` on both engines)
- [x] Revert-verify: confirmed every new test fails against the pre-fix source, then passes again once restored
- [x] Full suite green: `cargo test` (1056 passed), `cargo clippy -- -D warnings`, `cargo fmt --check`, `npm test` (5351 passed), `npm run lint`, `codegraph diff-impact --staged -T`

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/entrypoints.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/entrypoints.rs
@@ -43,6 +43,7 @@
 use rusqlite::Connection;
 use std::collections::{BTreeMap, HashMap, HashSet};
 
+use crate::domain::graph::resolve::resolve_pyproject_script_entrypoints;
 use crate::types::FileSymbols;
 
 /// Replace each reparsed Python file's persisted entrypoint-call evidence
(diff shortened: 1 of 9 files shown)


carlos-alm:
@greptileai

PR #2539, opened 2026-08-16
test(docs): guard README runtime-dependency claims against package.json drift

### LINKED ISSUE #2417: CI: guard README dependency counts against package.json/Cargo.toml drift
## Problem

The README's dependency claims drift out of sync with the manifests, silently. This has now happened twice:

- Backlog entry #96 caught "Only 3 runtime dependencies" when `package.json` had 5 (`graphology`, `graphology-communities-louvain`). Resolved by #545 removing those deps.
- It went stale again when `smol-toml` was added in c50a9919 (#2376) — README still said 3, actual is 4. Fixed manually in the current README pass.

There is no automated check, so the only thing catching it is a human reading the README closely.

## Scope

Three numbers in `README.md` are derived from manifests and should be asserted, not hand-maintained:

1. **npm runtime deps** — "Only **N** npm runtime dependencies" + the table rows below it, vs `package.json` `dependencies`.
2. **Cargo direct crates** — the "native engine is not dependency-free either" paragraph (46), vs `[dependencies]` in `crates/codegraph-core/Cargo.toml`.
3. **Competitive table row** — "Runtime dependencies (direct, all ecosystems)" for our own column (50 = 4 + 46).

## Suggested approach

A small test (or `scripts/` check wired into CI) that parses `package.json` + `crates/codegraph-core/Cargo.toml`, extracts the counts from `README.md`, and fails on mismatch. Cheapest version: assert the counts and the dependency-table row names match `Object.keys(pkg.dependencies)` exactly, so an added or removed dep breaks the build until the README is updated.

Competitor numbers in the comparison table stay manual — they can't be derived locally.

carlos-alm:
## Summary

Closes #2417

`README.md`'s runtime-dependency claims are hand-maintained numbers/table rows that silently drift out of sync with `package.json` — this has happened twice already (backlog #96; again when `smol-toml` was added in c50a9919/#2376, only caught by a human reading the README closely).

Adds `tests/integration/docs-readme-dependency-drift.test.ts`, asserting three things against `Object.keys(package.json.dependencies)`:
- the "Only **N** runtime dependencies" prose count
- the "Lightweight Footprint" dependency table's row names (exact set match, not just a count — a renamed or swapped dependency is caught, not just an added/removed one)
- the competitive-comparison table's own-column count

The issue also scoped a third number (a Cargo direct-crate count previously mentioned in a "native engine is not dependency-free either" paragraph). That paragraph no longer exists in the current README — the dependency section was restructured since the issue was filed to track npm deps only — so there's nothing left to guard there. Verified by reading the current README before writing the test rather than assuming the issue's original scope still applied unchanged.

This is a plain vitest test file, so it's "wired into CI" via the existing `npm test` job already required on every PR — no separate script or workflow needed, matching the issue's own "cheapest version" suggestion.

## Test plan
- [x] New test passes against current README.md/package.json
- [x] Revert-verify: manually corrupted the README's prose count, a table row's dependency name, and the competitive-table count (one at a time) and confirmed each specific assertion fails; restored and confirmed all 4 pass again
- [x] Full suite green: `npm test` (5359 passed), `npm run lint`, `codegraph diff-impact --staged -T` (no production code changed)

### DIFF
--- a/tests/integration/docs-readme-dependency-drift.test.ts
+++ b/tests/integration/docs-readme-dependency-drift.test.ts
@@ -1,0 +1,70 @@
+/**
+ * Regression test for #2417: README.md's runtime-dependency claims are
+ * hand-maintained numbers/table rows that silently drift out of sync with
+ * `package.json` — this has happened twice already (backlog #96, and again
+ * when `smol-toml` was added in c50a9919/#2376). Nothing catches it except a
+ * human reading the README closely.
+ *
+ * Guards the two places `package.json`'s `dependencies` are echoed in
+ * README.md: the "Lightweight Footprint" section's prose count + dependency
+ * table, and the competitive-comparison table's own-column count. Both must
+ * track `Object.keys(pkg.dependencies)` exactly — an added, removed, or
+ * renamed dependency breaks this test until the README is updated.
+ */
+
+import fs from 'node:fs';
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+const REPO_ROOT = path.join(__dirname, '../..');
+const README_PATH = path.join(REPO_ROOT, 'README.md');
+
+const pkg = JSON.parse(fs.readFileSync(path.join(REPO_ROOT, 'package.json'), 'utf-8'));
+const actualDeps: string[] = Object.keys(pkg.dependencies ?? {}).sort();
+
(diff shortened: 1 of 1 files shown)


carlos-alm:
@greptileai

Issue #2540, opened 2026-08-16
bug(build): tryNativeOrchestrator's catch swallows an unreadable-build-state error from the Rust pipeline


carlos-alm:
## Summary

Found while fixing #2418 (native `load_file_hashes` collapsed a read failure into "no prior build state", silently wiping a healthy graph on a transient `SQLITE_BUSY`/schema fault). #2418 fixes the Rust loader itself to correctly return an error instead of `None` in that case, and that error now propagates through `NativeDatabase::build_graph`'s `#[napi]` wrapper as a thrown `napi::Error` — up through `session.runBuildGraph()` and `tryNativeOrchestrator(ctx)` in `src/domain/graph/builder/stages/native-orchestrator.ts`, neither of which catch anything.

But `pipeline.ts`'s own call site does:

```ts
try {
  const nativeResult = await tryNativeOrchestrator(ctx);
  if (nativeResult === 'early-exit') return;
  if (nativeResult) return nativeResult;
} catch (err) {
  warn(`Native build orchestrator failed, falling back to JS pipeline: ${toErrorMessage(err)}`);
  // ...falls through to the JS pipeline, does NOT re-throw...
}
```

This unconditionally swallows *any* error from the native orchestrator — including the one #2418 makes it throw for unreadable prior state — and falls back to the JS pipeline instead of surfacing it. Contrast with the **earlier** JS-side fast-skip pre-flight a few lines above in the same function, which explicitly checks `isUnreadableBuildStateError(err)` before deciding whether to re-throw or fall through, specifically because "falling through here would... produce exactly the wipe this error prevents."

## Why this isn't just #2418 again

`isUnreadableBuildStateError` checks `e instanceof DbError` — a TS class. A Rust-thrown error surfaces at this boundary as a plain napi-generated `Error`, never a `DbError` instance, so the existing check can't recognize it even if added to this catch block unchanged. Correctly closing this gap needs either:
- matching on the error message content (fragile — ties TS logic to Rust's exact wording), or
- a structured signal from the Rust side (e.g. a distinguishable error code/prefix `NativeDatabase::build_graph` could set, that TS parses), or
- some other explicit contract between the two error-reporting layers.

That's a real design decision, not a one-line fix, which is why it's being filed separately rather than folded into #2418's precise, narrow ask (fix the Rust loader itself).

## Impact

[...]

PR #2541, opened 2026-08-16
fix(native): propagate a file_hashes read failure instead of reporting no prior state

### LINKED ISSUE #2418: fix(native): load_file_hashes collapses read failures into "no prior build state"
Deferred from PR #2414 review.

**Original reviewer comment:** https://github.com/optave/ops-codegraph-tool/pull/2414#discussion_r3746424339 (Greptile, P1 — "Native fallback still wipes state")

## The bug

`load_file_hashes` in `crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs:88` treats a failed row query as absent prior state:

```rust
let mut stmt = match conn.prepare("SELECT file, hash, mtime, size FROM file_hashes") {
    Ok(s) => s,
    Err(_) => return None,   // corruption, lock, I/O, schema drift → "no prior build"
};
```

`None` means a from-scratch build, and a full build `DELETE`s every graph table — `nodes`, `edges`, `file_hashes`, and `embeddings` among them. So a transient `SQLITE_BUSY` (a concurrent build, or the MCP server holding the DB) or a schema fault can silently discard a healthy graph and its expensive embeddings instead of failing and letting the operator retry.

This is the same defect PR #2414 fixed on the JS side, where `loadFileHashes` now reads table existence from `sqlite_master` and throws `DbError` (`DB_STATE_UNREADABLE`) when the table exists but cannot be read.

## Why this is still open

PR #2414 closed the reachable path: the native fast-skip pre-flight in `src/domain/graph/builder/pipeline.ts` used to swallow that error and fall through to the Rust orchestrator, and now re-throws it. So an ordinary incremental native build is protected.

The Rust loader itself is still wrong, which matters because:
- it is the source of truth whenever the pre-flight is bypassed;
- it violates the dual-engine rule in `CLAUDE.md` — the two engines must produce identical results, and the less-accurate one has the bug;
- direct consumers of the Rust API get no protection at all.

## The fix

Mirror the TS semantics in `load_file_hashes`. It already probes `sqlite_master` for existence, so the three cases are cleanly separable:

- table missing → no prior state (unchanged)
- table present, zero rows → no prior state (unchanged)
- table present, read fails → propagate the error instead of `return None`

That means changing the signature from `Option<HashMap<..>>` to `Result<Option<HashMap<..>>, _>` and handling it at the caller (`detect_changes.rs:1246`). The `.ok()?` and `filter_map(|r| r.ok())` on the `query_map` below have the same swallowing problem and should be reviewed in the same pass.

## Note

**Correction (was wrong when filed):** the original note said this was deferred because "this environment has no Rust toolchain (`cargo` is unavailable)". That is false. `cargo` (1.97.1) is installed at `~/.cargo/bin/cargo`; it is simply not on `PATH`, so a bare `which cargo` returns nothing and I wrongly read that as absent. The Rust change is verifiable locally and this issue is actionable now, not blocked.

To build the addon from source:

```
export PATH="$HOME/.cargo/bin:$PATH"
npm install --no-save @napi-rs/cli@3
cd crates/codegraph-core
../../node_modules/.bin/napi build --release
cp codegraph-core.node codegraph-core.darwin-arm64.node
```

The final `cp` matters: napi emits the unsuffixed `codegraph-core.node`, while `loadNative` (`src/infrastructure/native.ts`) resolves the platform-suffixed name — without it the stale npm-installed addon is still what loads.

**The remaining reason to keep this separate from #2414** stands on its own: `crates/codegraph-core/.../detect_changes.rs` is a different engine module, absent from that PR's diff, and #2414 was already review-approved and green when this was found. That is a scope/`one PR = one concern` argument, not a tooling limitation — so treat this as ordinary queued work rather than something gated on environment setup.

carlos-alm:
## Summary

Closes #2418

Deferred from PR #2414 review ([Greptile, P1](https://github.com/optave/ops-codegraph-tool/pull/2414#discussion_r3746424339)).

Native's `load_file_hashes` (`crates/codegraph-core/.../detect_changes.rs`) treated a failed row query the same as a genuinely absent/empty table — both returned `None`. `None` means "no prior build state," and a full build `DELETE`s every graph table (`nodes`, `edges`, `file_hashes`, `embeddings`), so a transient `SQLITE_BUSY` (a concurrent build, or the MCP server holding the DB) or a schema fault could silently discard a healthy graph and its expensive embeddings instead of failing loudly and letting the operator retry.

This mirrors the TS `loadFileHashes` semantics PR #2414 already established for the JS side:
- table missing → no prior state
- table present, zero rows → no prior state
- table present, read fails → propagate the error

- Changed `load_file_hashes`'s signature from `Option<..>` to `Result<Option<..>, String>`, and `detect_changes`'s from `ChangeResult` to `Result<ChangeResult, String>`, propagated through the single `pipeline.rs` call site via `?` (already the established error-propagation idiom for this function — the enclosing `run_pipeline` already returns `Result<_, String>`).
- Also stopped silently dropping a row that fails to deserialize mid-query (`collect::<Result<Vec<_>, _>>()` instead of `filter_map(|r| r.ok())`) — the same swallowing problem the issue asked to review in the same pass.
- Updated a stale comment in `pipeline.ts`'s JS-side pre-flight that described this Rust bug in the present tense ("which *still* collapses a read failure...") — now accurate about both layers.

## Note on scope

While verifying this end-to-end, found that `pipeline.ts`'s `tryNativeOrchestrator` catch swallows *any* native-thrown error (including this one) and falls back to the JS pipeline without checking for unreadable state the way the earlier pre-flight does. Verified empirically this doesn't currently cause data loss via the full `buildGraph()` flow — the JS fallback happens to independently re-detect the same corruption (via #2414's already-fixed `loadFileHashes`) and throws correctly. But it's a real, distinct design gap (an `instanceof DbError` check can't recognize a napi-thrown error) that matters for direct native-API consumers bypassing `pipeline.ts` entirely. Filed as #2540 rather than folding into this PR.

## Test plan
- [x] New Rust unit tests: table missing → `Ok(None)`, table empty → `Ok(None)`, rows present → `Ok(Some(..))`, schema-incompatible table → `Err`, and `detect_changes` itself propagates the error rather than reporting a full build
- [x] Revert-verify: temporarily restored the pre-fix "swallow as None" behavior behind a stub, confirmed both new tests fail, restored and confirmed they pass
- [x] Manually verified end-to-end: a real `buildGraph({engine:'native', incremental:true})` call against a corrupted `file_hashes` schema throws with a clear message and leaves the graph's node count unchanged, rather than silently wiping it
- [x] Full suite green: `cargo test` (1061 passed), `cargo clippy -- -D warnings`, `cargo fmt --check`, `npm test` (5359 passed), `npm run lint`, `codegraph diff-impact --staged -T`

### DIFF
--- a/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/pipeline.rs
@@ -647,7 +647,7 @@ pub fn run_pipeline(
         incremental,
         force_full_rebuild,
         scoped_rel_paths.as_ref(),
-    );
+    )?;
     timing.detect_ms = t0.elapsed().as_secs_f64() * 1000.0;
 
     // Filter out metadata-only changes
--- a/crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs
+++ b/crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs
@@ -101,6 +101,12 @@ fn relative_path(root_dir: &str, abs_path: &str) -> String {
 /// So an unreadable table (or a failed existence probe) returns `Err` and lets
 /// the caller decide; `--no-incremental` remains the explicit way to ask for
 /// a wipe-and-rebuild. Mirrors `loadFileHashes` in `detect-changes.ts`.
+///
+/// The `Err` message always embeds the literal `DB_STATE_UNREADABLE` marker
+/// (matching TS `UNREADABLE_BUILD_STATE`) — `tryNativeOrchestrator`'s catch in
+/// `pipeline.ts` greps for it via `isUnreadableBuildStateError`, since this
+/// error crosses the napi boundary as a plain `Error`, never a `DbError`
+/// instance carrying a `.code` (#2418 Greptile review).
 fn load_file_hashes(conn: &Connection) -> Result<Option<HashMap<String, FileHashRow>>, String> {
     // Read from `sqlite_master` rather than inferring existence from a failed
     // `SELECT` below, so a genuinely-absent table stays distinguishable from
@@ -115,10 +121,10 @@ fn load_file_hashes(conn: &Connection) -> Result<Option<HashMap<String, FileHash
         .map(|c| c > 0)
         .map_err(|e| {
             format!(
-                "Could not determine whether the file_hashes table exists: {e}. Refusing to \
-                 fall back to a from-scratch build, which would delete the existing graph and \
-                 its embeddings. Retry once the database is readable, or pass --no-incremental \
-                 to rebuild deliberately."
+                "DB_STATE_UNREADABLE: Could not determine whether the file_hashes table exists: \
+                 {e}. Refusing to fall back to a from-scratch build, which would delete the \
+                 existing graph and its embeddings. Retry once the database is readable, or \
+                 pass --no-incremental to rebuild deliberately."
             )
         })?;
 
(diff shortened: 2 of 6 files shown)


carlos-alm:
@greptileai

PR #2554, opened 2026-08-17
fix(extractors): stop crediting a fallback read a prior write already killed

### LINKED ISSUE #2438: follow-up: fallback liveness scan credits a read that a prior write already killed
Deferred from PR #2432 review.

**Context:** the `#2257` value-ref liveness scan (`hasLaterReferenceInEnclosingBlock` /
`blockContainsIdentifierExcluding` in `src/extractors/javascript.ts`, mirrored in
`crates/codegraph-core/src/extractors/javascript.rs`) decides whether a logical-or /
nullish-coalescing / ternary fallback is actually consumed by looking for a later
*read* of the declared variable. It correctly ignores plain writes (an assignment
target is a write, not a read), but it does **not** model a write as a *kill* — so a
read that occurs after the value has already been overwritten is still credited.

**Repro:**

```js
function outer() {
  var fn = options.custom || fetchLatestVersion;
  fn = other;   // kills the fallback value
  fn();         // reads `other`, NOT the fallback
}
```

A value-ref `calls` edge to `fetchLatestVersion` is emitted in both engines, even
though the fallback value is never consumed. Verified against both the WASM and
native extractors at PR #2432's head.

PR #2432 handles the *loop-head* form of this (a `for` / `for-in` / `for-of` head that
binds the name kills the value before the body runs, so the body is not scanned), but
the general statement-level case above is out of scope there: doing it properly needs
ordered kill / reaching-definition analysis across the enclosing block rather than the
current "is the name mentioned anywhere at or after this statement" scan.

**Impact:** false-positive `calls` edges, which in turn keep a fallback function and
its transitive callees out of `roles --role dead` when they are genuinely unreachable.
Same severity class as the shadowing false positives already fixed in #2432, but a
different mechanism.

**Suggested fix:** walk the enclosing block's statements in order, tracking whether
`name` has been unconditionally overwritten; stop crediting reads after the first such
write. Conditional writes (inside an `if`/loop) must not kill, since the original
value can still reach a later read.

carlos-alm:
## Summary

The `#2257` value-ref liveness scan (`hasLaterReferenceInEnclosingBlock` / `blockContainsIdentifierExcluding` in `src/extractors/javascript.ts`, mirrored in `crates/codegraph-core/src/extractors/javascript.rs`) decides whether a logical-or/nullish-coalescing/ternary fallback is actually consumed by looking for a later *read* of the declared variable. It correctly ignores plain writes (an assignment target is a write, not a read), but did not model a write as a *kill* — so a read occurring after the value had already been overwritten was still credited.

```js
function outer() {
  var fn = options.custom || fetchLatestVersion;
  fn = other;   // kills the fallback value
  fn();         // reads `other`, NOT the fallback
}
```

A value-ref `calls` edge to `fetchLatestVersion` was emitted in both engines, even though the fallback value is never consumed — a false positive that kept a fallback function (and its transitive callees) out of `roles --role dead`.

## Fix

Added `killsBinding`/`kills_binding`: true when a statement — a *direct* child of the enclosing block, the exact granularity the per-statement scan already iterates — unconditionally overwrites the name (a top-level `name = value;` assignment, or a `var name = value;` redeclaration). `hasLaterReferenceInEnclosingBlock`/`has_later_reference_in_enclosing_block` now checks each statement for a read first (unchanged), then stops the scan entirely once a kill is found.

- A write nested inside an `if`/loop/`switch`/`try` never matches — it surfaces as a single wrapping statement (`if_statement`, etc.), not as the assignment itself — so a conditional write correctly never kills, since the branch might not run.
- The killing statement's own right-hand side is still scanned for a genuine read *before* the kill takes effect (`fn = fn || other;` still credits the pre-existing read), since the read-check always runs before the kill-check for the same statement.
- Compound assignments (`+=`, a distinct `augmented_assignment_expression` grammar node) are unaffected — they already fall through to the generic identifier scan and are credited as a genuine read at that statement, independent of this change.

## Tests

- Extended `tests/integration/issue-2257-logical-or-ternary-value-ref.test.ts` (both WASM and native engines) with 4 new scenarios: kill via plain reassignment, kill via `var` redeclaration, conditional write does NOT kill, and a genuine same-statement read survives its own killing statement.
- Added 4 matching Rust unit tests in `crates/codegraph-core/src/extractors/javascript.rs`'s existing test module for the same scenarios.
- Revert-verified: temporarily disabled the kill-check in both engines and confirmed the exact 2 targeted assertions fail (all others unaffected), confirming the tests exercise the fix rather than passing trivially.

Closes #2438

### DIFF
--- a/crates/codegraph-core/src/extractors/javascript.rs
+++ b/crates/codegraph-core/src/extractors/javascript.rs
@@ -5439,6 +5439,59 @@ fn block_contains_identifier_excluding(
     false
 }
 
+/// True when `statement` — a DIRECT child of the enclosing block, exactly the
+/// granularity `has_later_reference_in_enclosing_block` iterates —
+/// unconditionally overwrites `name`: a top-level `name = value;` assignment
+/// (any operator; `pattern_binds_name` also covers destructuring targets
+/// like `[name] = arr`) or a `var name = value;` redeclaration sitting
+/// directly in the block. A write nested inside an `if`/loop/`switch`/`try`
+/// never matches here — it surfaces as a single
+/// `if_statement`/`for_statement`/etc. child, not as the assignment itself —
+/// so a conditional write correctly never kills (issue #2438's own
+/// requirement: the original value can still reach a later read when the
+/// write didn't actually run).
+///
+/// `exclude_id` skips the declarator this liveness check is FOR, so the
+/// declaration statement that introduces `name` (which trivially "binds"
+/// name via its own declarator) is never mistaken for a kill of its own
+/// freshly-assigned value.
+///
+/// Mirrors `killsBinding` in `src/extractors/javascript.ts`.
+fn kills_binding(statement: &Node, name: &str, source: &[u8], exclude_id: usize) -> bool {
+    let node: Node = if statement.kind() == "expression_statement" {
+        match statement.child(0) {
+            Some(child) => child,
+            None => return false,
+        }
+    } else {
+        *statement
+    };
+    if node.kind() == "assignment_expression" {
+        return match node.child_by_field_name("left") {
+            Some(left) => pattern_binds_name(&left, name, source, 0),
+            None => false,
+        };
+    }
+    if node.kind() == "variable_declaration" || node.kind() == "lexical_declaration" {
+        for i in 0..node.child_count() {
+            let Some(declarator) = node.child(i) else {
+                continue;
+            };
+            if declarator.kind() != "variable_declarator" || declarator.id() == exclude_id {
+                continue;
+            }
+            let decl_name = declarator.child_by_field_name("name");
+            let value = declarator.child_by_field_name("value");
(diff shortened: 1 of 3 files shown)


carlos-alm:
@greptileai please re-review — addressed both findings: kills_binding now recurses through parenthesized/sequence expressions, and block_contains_identifier_excluding now stops crediting reads from later declarators in the same statement once an earlier one kills the name.

carlos-alm:
@greptileai please re-review — addressed the sequence-expression ordering finding: block_contains_identifier_excluding now walks a sequence_expression's parts in order, checking each for a read before checking whether it kills the name, mirroring the same pattern used for top-level block statements and multi-declarator statements.

carlos-alm:
@greptileai please re-review — addressed the same-declaration finding: the hasExcludedDeclarator branch now checks declaratorKillsName after each later sibling declarator's read-check, matching the pattern used everywhere else in this function.

PR #2557, opened 2026-08-17
fix(hooks): normalize $IFS/${IFS} before any guard-git.sh verb-detection check

### LINKED ISSUE #2451: guard-git.sh: unquoted ${IFS}/whitespace-expansion between git/gh and any subcommand bypasses every check
## Summary

While fixing #2280 (the `checkout --` false-positive/bypass narrowing), Greptile's review surfaced that `${IFS}` (bash's own whitespace-field-separator variable) can supply a token boundary that the command TEXT itself has no literal whitespace for — `git checkout --${IFS}file.txt` looks, as static text, like a single token, but bash expands the unquoted `${IFS}` to whitespace before Git ever runs, so Git actually receives `checkout -- file.txt`.

That specific instance (the `checkout --`/`--ours`/`--theirs` boundaries) is fixed in #2280 / PR #2450. But the same class of bypass exists much more broadly, and predates that PR:

1. **The hook's own top-level fast-path filter** (`guard-git.sh` line ~32):
   ```
   if ! echo "$COMMAND" | grep -qE '(^|[[:space:]]|&&[[:space:]]*|["'"'"'`(])(git|gh)[[:space:]]+'; then
     exit 0
   fi
   ```
   This requires **literal** whitespace directly after `git`/`gh`. A command like `git${IFS}checkout${IFS}--${IFS}file.txt` has no literal whitespace anywhere in it, so this filter doesn't recognize it as a git command at all, and the hook exits 0 immediately — **skipping every single check in the file** (`add -A`, `reset`, `checkout --`, `restore`, `clean -f`, `stash`, all of it), not just the checkout one.

2. **Every individual subcommand check** in the file (`git[[:space:]]+add[[:space:]]+`, `git[[:space:]]+reset`, `git[[:space:]]+restore`, `git[[:space:]]+clean`, `git[[:space:]]+stash`, etc.) has the same literal-whitespace assumption between the verb and its own subcommand/flags — e.g. `git add${IFS}-A` still gets recognized as *a* git command by the fast-path (since `git ` itself has a real space), but the specific `add -A` block regex (`add[[:space:]]+(\.[[:space:]]*$|-A|--all)`) won't match it, since it requires literal whitespace between `add` and `-A` too.

## Repro

```bash
# Bypasses the fast-path filter entirely — hook exits 0, no check runs:
git${IFS}checkout${IFS}--${IFS}some-file.txt

# Recognized as *a* git command, but slips past the add -A-specific check:
git add${IFS}-A
```

## Suggested fix

This needs a more systemic fix than a single regex tweak — likely normalizing runs of shell metacharacters/variables that are KNOWN to commonly expand to whitespace (`$IFS`/`${IFS}` at minimum) into a literal space before any of the verb-detection regexes run, similar in spirit to how `mask-quoted-text.mjs` already normalizes quoted-string contents before verb detection (#2099). Given how many regexes in this file share the `[[:space:]]+`-after-keyword assumption, a shared normalization pass (rather than patching each regex individually, which is how PR #2450 had to iterate three times just for one check) is probably the right architecture — see CLAUDE.md's "prioritize the best architecture" guidance.

Out of scope for #2280 / PR #2450, which is narrowly about the `checkout --` false-positive on legitimate flags (`--detach`, `--track`, etc.) plus the two IFS-specific instances Greptile found on that same check.

carlos-alm:
## Summary

While fixing #2280 (the `checkout --` false-positive/bypass narrowing), Greptile's review surfaced that `${IFS}` (bash's own whitespace-field-separator variable) can supply a token boundary that the command TEXT itself has no literal whitespace for — `git checkout --${IFS}file.txt` looks, as static text, like a single token, but bash expands the unquoted `${IFS}` to whitespace before Git ever runs, so Git actually receives `checkout -- file.txt`.

That specific instance (the `checkout --`/`--ours`/`--theirs` boundaries) was fixed in #2280 / PR #2450. This issue is the broader, systemic version, predating that PR:

1. **The hook's own top-level fast-path filter** requires literal whitespace directly after `git`/`gh`. A command like `git${IFS}checkout${IFS}--${IFS}file.txt` has no literal whitespace anywhere, so the filter didn't recognize it as a git command at all, and exited 0 immediately — **skipping every single check in the file**, not just the checkout one.
2. **Every individual subcommand check** has the same literal-whitespace assumption between the verb and its own subcommand/flags — e.g. `git add${IFS}-A` is recognized as *a* git command (the space after `git` is real), but slips past the `add -A`-specific check, which requires literal whitespace between `add` and `-A` too.

## Fix

Rather than patching each regex individually (how PR #2450 had to iterate three times just for one check), this adds a shared normalization pass — `normalize-ifs.mjs` — that replaces every `$IFS`/`${IFS}` reference with a literal space **upstream of both** the fast-path filter and the existing quote-masking pipeline (`mask-quoted-text.mjs`), so no individual regex needs its own IFS-awareness.

Deliberately not quote-aware on its own: it runs *before* masking, so:
- An IFS reference inside a quoted, inert string gets replaced with a space here, then blanked out (along with everything else in that span) by masking anyway.
- An IFS reference inside a command-substitution span (`$(...)`/backticks) is left exactly as normalized, since masking copies those spans through verbatim — they execute unconditionally in real bash, so IFS-driven splitting genuinely applies there too.

The bare `$IFS` form (as opposed to the unambiguous braced `${IFS}`) only safely references the `IFS` variable when the next character isn't part of a longer identifier — `$IFScheckout` is bash's own distinct (and here undefined, empty-expanding) variable `IFScheckout`, not `$IFS` + literal `checkout`. The normalization uses a negative-lookahead boundary to avoid over-matching this.

## Tests

- `tests/unit/hook-guard-git-ifs-bypass.test.ts`: exercises both bug classes from the issue's repro (fast-path bypass via `git${IFS}checkout${IFS}--${IFS}file.txt`, and the `add -A`-specific bypass via both braced and bare IFS forms), plus the negative cases (doesn't over-match a longer variable name, doesn't create a false token boundary inside quoted text, still allows legitimate `--detach`-style flags).
- Revert-verified: ran the new tests against the pre-fix hook script and confirmed the targeted ones fail; restored and confirmed all pass.
- All pre-existing `guard-git.sh` tests (including PR #2450's own checkout-- IFS tests) still pass — no regression to that prior fix.
- Updated `docs/examples/claude-code-hooks/` (README + byte-identical copies of `guard-git.sh` and the new `normalize-ifs.mjs`) and added the new file to the existing docs-sync test's tracked list, per that test's own established pattern.
- Full suite: 339 test files / 5410 tests pass.

Closes #2451

### DIFF
--- a/.claude/hooks/guard-git.sh
+++ b/.claude/hooks/guard-git.sh
@@ -21,6 +21,8 @@ if [ -z "$COMMAND" ]; then
   exit 0
 fi
 
+HOOK_DIR="$(cd "$(dirname "$0")" && pwd)"
+
 # The actual directory the Bash tool will run this command in — a top-level
 # field on every PreToolUse payload, independent of whatever cd/-C text
 # happens to appear in $COMMAND. This is what a bare `git push` (relying on
@@ -35,6 +37,25 @@ HOOK_CWD=$(echo "$INPUT" | node -e "
   });
 " 2>/dev/null) || true
 
+# Replace every $IFS/${IFS} reference with a literal space (#2451) BEFORE
+# even the fast-path filter below runs. `$IFS`/`${IFS}` (bash's own
+# whitespace field-separator variable) supplies a token boundary bash
+# expands at execution time that the command TEXT itself has no literal
+# whitespace for — `git${IFS}checkout` looks like a single token to every
+# regex in this file, including the fast-path filter's own
+# `(git|gh)[[:space:]]+` check, but bash expands the unquoted `${IFS}` to
+# whitespace before Git ever runs, so Git actually receives `checkout` as a
+# separate argument. Without this, the fast-path filter doesn't recognize
+# such a command as git/gh at all and exits 0 immediately, skipping EVERY
+# check below, not just whichever one the attacker was targeting. See
+# normalize-ifs.mjs for why this can run without any quote-awareness of its
+# own. Extraction logic below (detect_work_dir, MSG_FILE, the AI-attribution
+# scan) deliberately keeps reading the raw, unnormalized $COMMAND.
+IFS_NORM_COMMAND=$(echo "$COMMAND" | node "$HOOK_DIR/normalize-ifs.mjs" 2>/dev/null) || true
+if [ -z "$IFS_NORM_COMMAND" ]; then
+  IFS_NORM_COMMAND="$COMMAND"
+fi
+
 # Act on git and gh commands (may appear after cd "..." &&, inside a quoted
 # nested-shell invocation like `bash -c "git clean -fd"`, or inside a command
 # substitution like `"message $(git clean -fd)"` — #2099 Greptile review).
@@ -43,7 +64,7 @@ HOOK_CWD=$(echo "$INPUT" | node -e "
 # rather than exact: a false "yes" here just means the rest of the script
 # runs its checks anyway, while a false "no" would exit before ever reaching
 # them.
-if ! echo "$COMMAND" | grep -qE '(^|[[:space:]]|&&[[:space:]]*|["'"'"'`(])(git|gh)[[:space:]]+'; then
+if ! echo "$IFS_NORM_COMMAND" | grep -qE '(^|[[:space:]]|&&[[:space:]]*|["'"'"'`(])(git|gh)[[:space:]]+'; then
   exit 0
 fi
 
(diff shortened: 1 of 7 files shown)


carlos-alm:
@greptileai please re-review — addressed both findings: normalize-ifs.mjs now generically matches any ${IFS<operator>...} form (covering substring expansion like ${IFS:0:1}, not just the whole-variable reference), and Quick setup's cp commands now also copy *.mjs files (fixing the gap for normalize-ifs.mjs as well as the pre-existing mask-quoted-text.mjs/check-git-clean-force.mjs).

carlos-alm:
@greptileai please re-review — addressed the false-positive finding: normalize-ifs.mjs is now narrowly scoped back to just the whole-variable ($IFS/${IFS}) and substring (${IFS:offset}/${IFS:offset:length}) forms, dropping the over-broad any-operator generalization that wrongly normalized ${IFS:+word}, ${IFS/pattern/replacement}, and ${IFS@Q}-style transformation operators (which substitute arbitrary text, not IFS's own whitespace characters).

carlos-alm:
@greptileai please re-review — addressed both findings: the substring regex's length portion now requires a non-zero digit sequence (rejecting ${IFS:0:0}, which extracts an empty string, not whitespace), and a new, narrowly-scoped regex recognizes ${IFS:+ }/${IFS+ } specifically when the substituted content between the operator and closing brace is all whitespace (nothing else), leaving ${IFS:+x} and ${IFS:+} (empty) untouched. Filed #2558 for the related, broader problem (this trick isn't unique to the IFS name) as an explicitly out-of-scope follow-up.

carlos-alm:
@greptileai please re-review — addressed the out-of-range offset finding: the substring regex now constrains OFFSET to exactly the values provably non-empty against IFS's 3-character default value (0/1/2 from the start, -1/-2/-3 from the end), rejecting ${IFS:3} and ${IFS:5:1} which both return empty in real bash regardless of an otherwise-valid length.

carlos-alm:
@greptileai please re-review — addressed the arithmetic-spelling finding: OFFSET (0*[0-2] / -0*[1-3]) and LENGTH (0*[1-9]\d*) now tolerate leading zeros, so ${IFS:00} and ${IFS:0:01} normalize identically to ${IFS:0} and ${IFS:0:1}. This intentionally does not attempt full arithmetic-expression support (hex, operators, nested expansions) — that's the same open-ended class already tracked in #2558, distinct from the bounded, concretely-named leading-zero gap.

carlos-alm:
@greptileai please re-review — addressed the negative-zero finding: the offset alternation now includes -0+ (a literal minus followed by one or more zeros and nothing else), correctly treating ${IFS: -0}/${IFS: -00} the same as offset 0 (integers have no signed zero, so bash evaluates -0 identically to 0).

Issue #2299, opened 2026-08-06
Import resolution doesn't remap .mjs/.cjs specifiers to .mts/.cts sources


carlos-alm:
## Summary

Follow-up to #2073 (which fixed `.mts`/`.cts` file recognition — both engines now walk and parse these extensions). While verifying that fix end-to-end, I found a related but distinct gap: **import resolution never remaps a `.mjs`/`.cjs` specifier to an actual `.mts`/`.cts` source file**, unlike the existing `.js` → `.ts`/`.tsx` remap.

This matters because TypeScript's `NodeNext`/`Node16` module resolution requires the *emitted* extension in relative import specifiers, not the source extension: a `.mts` file is imported via `import './foo.mjs'` (not `'./foo.mts'`), and a `.cts` file's CJS output is `.cjs`. This is the exact same "source extension differs from the specifier's extension" situation the existing `.js`→`.ts`/`.tsx` remap already solves for plain `.ts` files — it just isn't extended to the ESM/CJS-specific pair.

## Evidence

Reproduced by building a 3-file fixture (`util.mts` exporting `greet`, `index.mts` importing it via `import { greet } from './util.mjs'`, and a `legacy.cts`) with `codegraph build --engine wasm`:

- The `calls` edge from `main` → `greet` resolves with **confidence 0.7** (same-directory proximity fallback), not the 1.0 an import-aware match gets.
- A control fixture with plain `.ts`/`.js` (`import { greet } from './util.js'` where the source is `util.ts`) resolves the identical pattern at **confidence 1.0**, because `remapJsToTs` in `src/domain/graph/resolve.ts:345` explicitly retries `.ts`/`.tsx` when a `.js`-suffixed specifier doesn't exist on disk.

`remapJsToTs` only triggers on `resolved.endsWith('.js')` — `.mjs`/`.cjs` specifiers never match that check, so they never get remapped to a `.mts`/`.cts` counterpart. The mirrored Rust function `probe_js_to_ts_remap` in `crates/codegraph-core/src/domain/graph/resolve.rs:819` has the identical gap (`if !resolved_str.ends_with(".js") { return None; }`, then only tries `[".ts", ".tsx"]`).

## Suggested fix

Mirror the existing `.js`→`.ts`/`.tsx` remap for the ESM/CJS pair, in both engines:
[...]

Issue #2073, opened 2026-07-09
TypeScript .mts/.cts files are not recognized by either engine (not walked, not parsed)


carlos-alm:
## Summary

While investigating a Greptile finding on PR #1978 (which widens `TYPESCRIPT_EXTENSIONS` scoping for `isTypeErasedImportTarget`), I found that TypeScript's `.mts` (ESM) and `.cts` (CJS) module extensions are not recognized as TypeScript — or as any supported language — anywhere in the pipeline, on either engine. Files with these extensions are silently skipped at the file-collection stage and never enter the graph at all.

This is a real, user-facing gap: `.mts`/`.cts` are increasingly common in real-world TypeScript packages that ship dual ESM/CJS builds.

## Evidence

**WASM/TS side** — file collection gates on `EXTENSIONS` (derived from `LANGUAGE_REGISTRY`), which never includes `.mts`/`.cts`:
- `src/domain/parser.ts:771-792` — `LANGUAGE_REGISTRY` entries for `typescript` (`extensions: ['.ts']`) and `tsx` (`extensions: ['.tsx']`) — no `.mts`/`.cts`.
- `src/domain/parser.ts:1033` / `src/shared/constants.ts:46` — `SUPPORTED_EXTENSIONS`/`EXTENSIONS` derived from the registry above, so also excludes them.
- `src/domain/graph/builder/helpers.ts:180` — `isCollectableSourceFile` gates the file walker on `EXTENSIONS.has(path.extname(entry.name))`. A `.mts`/`.cts` file fails this check and is never added to the walked file list — it never reaches the parser at all.
- `src/domain/parser.ts:490-496` — `NATIVE_SUPPORTED_EXTENSIONS` (used to classify native-drop fallbacks) also lacks `.mts`/`.cts`.

**Native/Rust side** — same gap, independently:
- `crates/codegraph-core/src/domain/parser.rs:89-140` — `LanguageKind::from_extension` explicitly checks `.tsx` and `.ts`/`.d.ts`, then falls through a `match ext { ... _ => None }` that has no `"mts"`/`"cts"` arm.
- `crates/codegraph-core/src/domain/graph/builder/stages/collect_files.rs:49-57` — the Rust-side `SUPPORTED_EXTENSIONS` list (used as a fallback in `is_supported_extension`) also has no `"mts"`/`"cts"` entry.
[...]

PR #2564, opened 2026-08-18
fix(resolve): probe .mts/.cts extensions for extension-less imports

### LINKED ISSUE #2464: Extension-less import resolution never probes .mts/.cts (only .ts/.tsx/.js/.jsx/.mjs)
## Summary

Discovered while fixing #2299 (the `.mjs`→`.mts`/`.cjs`→`.cts` emitted-extension remap). A separate, distinct code path — resolving an extension-less relative specifier like `import './foo'` by probing a fixed list of candidate extensions — never tries `.mts`/`.cts` in either engine, even though both engines otherwise recognize and parse `.mts`/`.cts` files (#2073).

- TS: the extension-probing loop in `resolveImportPathJS` (`src/domain/graph/resolve.ts`, ~line 1142) tries `['.ts', '.tsx', '.js', '.jsx', '.mjs', '.py', '.pyi', '/index.ts', '/index.tsx', '/index.js', '/__init__.py']` — no `.mts`/`.cts`.
- Rust: `probe_known_extensions`'s `EXTENSIONS` constant (`crates/codegraph-core/src/domain/graph/resolve.rs`, ~line 1382) has the identical list, same gap.

So `import './foo'` resolving to a source file `foo.mts`/`foo.cts` (as opposed to the emitted-specifier form `import './foo.mjs'`, which #2299 covers) falls through to proximity-based confidence instead of an import-aware match.

## Suggested fix

Add `.mts`/`.cts` to both extension lists, in both engines, keeping the two lists in sync per the dual-engine parity requirement.

## Scope note

Not addressed in #2299 itself — that issue is specifically about the emitted-extension remap (`.mjs`→`.mts`, a specifier that already carries an extension), not extension-less resolution. Distinct code path, filed separately to keep each PR to one concern.

carlos-alm:
## Summary

Discovered while fixing #2299 (the `.mjs`→`.mts`/`.cjs`→`.cts` emitted-extension remap). A separate, distinct code path — resolving an extension-less relative specifier like `import './foo'` by probing a fixed list of candidate extensions — never tries `.mts`/`.cts` in either engine, even though both engines otherwise recognize and parse `.mts`/`.cts` files (#2073).

- TS: the extension-probing loop in `resolveImportPathJS` (`src/domain/graph/resolve.ts`) tried `['.ts', '.tsx', '.js', '.jsx', '.mjs', '.py', '.pyi', ...index variants]` — no `.mts`/`.cts`.
- Rust: `probe_known_extensions`'s `EXTENSIONS` constant (`crates/codegraph-core/src/domain/graph/resolve.rs`) had the identical list, same gap.

So `import './foo'` resolving to a source file `foo.mts`/`foo.cts` (as opposed to the emitted-specifier form `import './foo.mjs'`, which #2299 covers) fell through to proximity-based confidence instead of an import-aware match.

## Fix

Added `.mts`/`.cts` to both extension lists, in both engines, placed alongside `.ts`/`.tsx` to keep TypeScript-family extensions grouped (matching the existing ordering convention where TS extensions precede JS extensions in the list).

## Test plan

- [x] New tests in `tests/unit/resolve.test.ts`: extension-less specifier resolves to `.mts`, extension-less specifier resolves to `.cts` — reusing the exact fixture files (`src/esm/util.mts`, `src/cjs/legacy.cts`) #2299's own remap tests already set up, but exercising the extension-probing loop directly (no extension on the specifier at all) rather than `EMIT_EXTENSION_REMAPS`
- [x] New tests in `crates/codegraph-core/src/domain/graph/resolve.rs` mirroring both cases
- [x] Revert-verified on both engines: temporarily removed `.mts`/`.cts` from each list and confirmed the corresponding new tests fail with the exact pre-fix output (resolves to the bare specifier, no extension); restored and confirmed green
- [x] `cargo test --lib` (full suite) — 1096 passed (up from 1094, +2 new)
- [x] `cargo clippy --lib -- -D warnings` — clean
- [x] `npm test` (full suite) — 339 files, 5449 passed
- [x] `npm run lint` — clean

Closes #2464

### DIFF
--- a/crates/codegraph-core/src/domain/graph/resolve.rs
+++ b/crates/codegraph-core/src/domain/graph/resolve.rs
@@ -1512,6 +1512,8 @@ fn probe_known_extensions(
         ".pyi",
         "/index.ts",
         "/index.tsx",
+        "/index.mts",
+        "/index.cts",
         "/index.js",
         "/__init__.py",
     ];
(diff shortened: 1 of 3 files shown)


carlos-alm:
@greptileai

Issue #2478, opened 2026-08-13
dart: local variable (not just parameter) shadowing a class field still mis-resolves receiver type


carlos-alm:
## Background

Follow-up to the Greptile finding on PR #2477 (itself a follow-up to #2319): Dart's receiver extraction (`findDartSelectorReceiver` in `src/extractors/dart.ts`, `find_dart_selector_receiver` + `handle_dart_call_expression` in `crates/codegraph-core/src/extractors/dart.rs`) now detects when a bare-identifier receiver is shadowed by a same-named **parameter** of the enclosing function/method, and skips the `this.`-prefix normalization that would otherwise incorrectly trigger the resolver's class-scoped field lookup for that receiver.

That fix's scope was deliberately limited to **parameters only**. Dart also allows a **local variable declaration** to shadow a class field:

```dart
class Service {
  final Repository _repo;   // field, type Repository
  Service(this._repo);

  void run() {
    // ... some code ...
    var _repo = MockRepository();   // LOCAL VARIABLE shadows the field, different type!
    _repo.mockOnlyMethod();         // means the LOCAL here, not the field
  }
}
```

This case is **not** handled by the parameter-shadowing fix — the extractor has no way to know a local variable declaration exists between the start of the enclosing block and the call site, so a bare `_repo.mockOnlyMethod()` after such a local declaration would still be treated as (potentially) a field access and could resolve incorrectly, or not fall through as safely as the parameter case does.

## Why this was scoped out

Full lexical-scope-correct local-variable shadowing detection requires tracking declaration order within nested blocks (a local variable declared AFTER the call site in the same block does NOT shadow it there; one declared BEFORE does, for the rest of the enclosing block) — a materially bigger undertaking than the parameter case, which only needs a single formal-parameter-list lookup with no ordering/control-flow reasoning. Per this file's own "conservative, don't guess, document the residual gap" convention (see `findDartSelectorReceiver`'s own doc comment for other deliberately-unhandled cases, e.g. chained-call intermediate receivers, subscript-indexed receivers, explicit `this.field.method()` chains), this gap is being tracked here rather than silently left unstated.

## Scope

- `src/extractors/dart.ts` (`findDartSelectorReceiver`, `findEnclosingDartParamListForCall` and friends)
- `crates/codegraph-core/src/extractors/dart.rs` (`find_dart_selector_receiver`, `handle_dart_call_expression`)

[...]

PR #2567, opened 2026-08-18
fix: seed typeMap for Dart local var initialized by a constructor call

### LINKED ISSUE #2474: Dart: local-variable constructor-call typing not seeded (svc.method() unresolved)

carlos-alm:
## Summary

`var svc = UserService(repo);` never seeded a typeMap entry for `svc`, unlike every other language extractor's identical "assign a constructor call to a local variable" convention (e.g. JS/TS's `handleVarDeclaratorTypeMap`). A later call through that local (`svc.createUser()`) could therefore never resolve via the typeMap, and the call edge was silently dropped.

- Adds `handleDartLocalVarTypeMap` (TS, `src/extractors/dart.ts`) and its mirror `handle_dart_local_var_type_map` (Rust, `crates/codegraph-core/src/extractors/dart.rs`), wired to `initialized_variable_definition`. Seeds a function-scoped typeMap entry (confidence 1.0) when the initializer is a bare constructor call.
- Both engines' `tree-sitter-dart` grammars needed separate handling:
  - Native (crates.io tree-sitter-dart 0.2): a clean `value: call_expression`.
  - WASM (npm tree-sitter-dart 1.x): confirmed by direct parse dump that `value:` is a field marker on **two** different children (the bare callee identifier *and* the trailing call `selector`) — `childForFieldName('value')` only returns the first match, so an earlier version of this fix wrongly took the "native" branch and bailed out before reaching the correct sibling-based lookup. Fixed by gating on the value field's type (`=== 'call_expression'`), not merely its presence.
- Extracted a shared `findEnclosingDartSignatureFromBody` / `find_enclosing_dart_signature_from_body` helper (refactored out of `findEnclosingDartParamListForCall` / `find_enclosing_dart_param_list_for_call`) since a Dart function's signature and body are sibling nodes, not nested — a body descendant (like a local variable declaration) can't reach its own signature via a simple ancestor walk.

Deliberately does not attempt to resolve a local variable shadowing a same-named class field (tracked separately as #2478 — already noted in this file's existing doc comments as out of scope for this fix).

## Test plan
- [x] New Rust unit tests in `dart.rs` (`local_var_constructor_call_typing` module) — seeding, scoping across functions, non-constructor-call no-op, class-method scoping, issue repro
- [x] New TS unit tests in `tests/parsers/dart.test.ts` (`#2474` describe block) — same coverage, WASM engine
- [x] New dual-engine integration test `tests/integration/issue-2474-dart-local-var-constructor-call-typing.test.ts` — asserts the actual call edge resolves via `buildGraph`, both `wasm` and `native`
- [x] Revert-verified: disabling the dispatch arm on each side reproduces the pre-fix failures for the exact new tests added
- [x] `npx tsc --noEmit -p .`, `npm run lint`, full `npm test` (5464 passed)
- [x] `cargo fmt -- --check`, `cargo clippy --lib -- -D warnings`, `cargo test --lib` (1107 passed)

Closes #2474

### DIFF
--- a/crates/codegraph-core/src/extractors/dart.rs
+++ b/crates/codegraph-core/src/extractors/dart.rs
@@ -896,10 +896,8 @@ fn find_enclosing_dart_function_qualifier_for_body(node: &Node, source: &[u8]) -
     })
 }
 
-/// Seed a function-scoped typeMap entry (confidence 1.0 — a constructor call
-/// determines its assigned local's type with full certainty, matching every
-/// other language extractor's identical "assign a constructor call to a
-/// local variable" convention) for `var svc = UserService(repo);` (#2474).
+/// Seed a function-scoped typeMap entry for `var svc = UserService(repo);`
+/// (#2474).
 ///
 /// tree-sitter-dart's two grammar versions structure this differently, same
 /// class of divergence `handle_dart_constructor_call` already documents:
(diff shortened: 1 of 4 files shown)


carlos-alm:
@greptileai

carlos-alm:
Addressed both findings in the latest commit (fd438c63):

**Finding 1 (bare lowercase call → constructor type)**: fixed by gating the seeding on the callee being capitalized, matching Dart's own type-naming convention.

**Finding 2 (uppercase ordinary function → constructor type)**: this one is real and I verified it empirically rather than arguing it away — added a dual-engine integration test with a capitalized top-level function (`OrderService MakeOrderService()`), and confirmed the resulting call edge (`order.placeOrder()`) is genuinely dropped on both engines. Root cause: `resolveByReceiver` (TS) and `resolve_call_targets_core` (Rust) both skip the untyped direct-qualified-method fallback whenever *any* typeMap entry exists for the receiver, right or wrong — this is a pre-existing, language-agnostic property of the shared call resolver, not something this PR introduces.

Fully closing this needs either (a) a shared-resolver change so a failed type-aware lookup falls through to the untyped fallback, which has blast radius across every language using this cascade and needs its own precision/recall validation, or (b) a same-file "is this name already a known ordinary function?" cross-check in the Dart extractor, which requires first refactoring `dart.ts`'s single-pass `walkDartNode` into the two-pass design `dart.rs`/`javascript.ts` already use (a single-pass check would be declaration-order-dependent and diverge from the native engine). Both are out of scope for this fix — filed as #2568.

For this PR, I lowered the heuristic's confidence from 1.0 to 0.7, matching the identical capitalization-based uncertainty tier this codebase already accepts for JS/TS's `Foo.create()` factory heuristic, and added a test locking in the one guarantee that does hold: the wrong guess never fabricates an edge to a nonexistent node (it drops the edge, same as the pre-#2474 baseline for this specific ambiguous shape — not a regression, just not a full fix of an unrelated, harder problem).

carlos-alm:
@greptileai

Issue #2572, opened 2026-08-18
bug(parity): native engine tags inherited-method call edges as ts-native instead of cha (jelly-micro classes/classes2)


carlos-alm:
## Context

Discovered while investigating #2482 (whose own described divergence — bind.js/fun.js self-referential dynamic call edges — no longer reproduces on current `origin/main`; see that issue for the resolution). Re-running the exact same repro surfaced a different, unrelated divergence in the same fixture.

## Divergence

```
node scripts/parity-compare.mjs --langs jelly-micro
```

```
=== jelly-micro: wasm vs native DIVERGED (0 node diffs, 6 edge diffs)
  [edge] [calls] classes/classes.js:classes/classes.js(file) -> classes/classes.js:A.s2(method) conf=0.8 dyn=0 technique=cha  wasm=1 native=0
  [edge] [calls] classes/classes.js:classes/classes.js(file) -> classes/classes.js:A.s2(method) conf=0.8 dyn=0 technique=ts-native  wasm=0 native=1
  [edge] [calls] classes/classes.js:classes/classes.js(file) -> classes/classes.js:C11.f16(method) conf=0.8 dyn=0 technique=cha  wasm=1 native=0
  [edge] [calls] classes/classes.js:classes/classes.js(file) -> classes/classes.js:C11.f16(method) conf=0.8 dyn=0 technique=ts-native  wasm=0 native=1
  [edge] [calls] classes2/classes2.js:classes2/classes2.js(file) -> classes2/classes2.js:E.m1(method) conf=0.8 dyn=0 technique=cha  wasm=1 native=0
  [edge] [calls] classes2/classes2.js:classes2/classes2.js(file) -> classes2/classes2.js:E.m1(method) conf=0.8 dyn=0 technique=ts-native  wasm=0 native=1
```

Note this is NOT a missing or wrong-target edge — source, target, confidence (0.8), and dynamic flag (0) all match between engines. The only difference is the `technique` metadata tag: WASM tags these edges `cha` (class-hierarchy-analysis-based dispatch), native tags the identical edge `ts-native` (the generic direct-resolution fallback label — per `build_edges.rs`'s own doc comment, `'ts-native'` "is not an engine marker — it is the resolution-technique... backfill" applied when nothing more specific tagged the edge).

## Likely shape

Both affected call sites are an inherited-method dispatch through a subclass instance that does not override the method:

```js
// classes/classes.js
function A(x) { this.f44 = x; }
[...]

PR #2573, opened 2026-08-18
fix(native): stop scoping full-build dataflow-vertex extraction to files with edges

### LINKED ISSUE #2483: bug(parity): native engine emits far fewer dataflow vertices than WASM for JS/TS/pts-javascript fixtures
## Found while working on

#2326 (cargo clippy CI gate). Not caused by that change — reproduces identically on a clean `origin/main` checkout with no modifications, confirmed by building the native addon from `origin/main` in an isolated worktree and re-running the same check.

## Divergence

`node scripts/parity-compare.mjs --dataflow --langs javascript,typescript,pts-javascript`:

```
PARITY FAILED — 3/3 fixture(s) diverged: javascript, pts-javascript, typescript
```

- `pts-javascript`: 6 df-vertex diffs (all `wasm=1 native=0`), e.g. `array-from.js:mapCallback|param|item`, `dispatch-table.js:runDispatch|param|key`, `spread.js:consumer1|param|x`, etc.
- `typescript`: 20 df-vertex diffs (all `wasm=1 native=0`) — params, locals, and returns across `barrel-consumer.ts`, `hierarchy.ts`, `index.ts`, `serializer.ts`.
- `javascript` fixture also diverges (see full run output; truncated here for brevity).

In every case the WASM engine records the dataflow vertex and the native engine does not — this isn't a handful of edge-case misses, it's most params/locals/returns for the affected functions, suggesting the native dataflow-vertex path may not be wired up for these node shapes (or the fixtures use a code pattern the native `DataflowRules` for JS/TS don't cover, distinct from the already-tracked Dart-specific gaps in #2356/#2357/#2359).

## Repro

```
npm run build
cd crates/codegraph-core && npx napi build --platform --release && codesign --sign - --force codegraph-core.darwin-arm64.node && cp codegraph-core.darwin-arm64.node ../../node_modules/@optave/codegraph-darwin-arm64/codegraph-core.node
cd ../..
node scripts/parity-compare.mjs --dataflow --langs javascript,typescript,pts-javascript
```

## Note

The script's own usage comment notes `--dataflow` "requires migration v18" — worth checking whether the test DB in these fixture builds actually has that migration applied, or whether this is a genuine extraction-logic gap in the native `DataflowRules` for JS/TS.

carlos-alm:
## Summary

`runDataflowVertexPass`'s full-build file selection (the P6 pass that backfills `dataflow_vertices` after the Rust orchestrator's edge-only dataflow pass) only included native-language files that already had `dataflow` EDGE rows (`flows_to`/`returns`/`mutates`) — on the theory that a file with no edges would produce zero vertices anyway.

That theory is wrong: a plain leaf function with params and a return but no calls to or from any other function has zero *inter-procedural* edges, yet still has vertex-worthy params/returns — `extractDataflowAnalysis`'s vertex output isn't gated on `argFlows`/`assignments`/`mutations` being non-empty. In practice this meant the native engine silently dropped `dataflow_vertices` for most files in a typical codebase (confirmed empirically: `bind-call-apply.js`, `class-scope.js`, `define-property.js`, `define-property-accessor.js` in the `javascript` fixture, plus most of `hierarchy.ts`/`index.ts`/`serializer.ts` in `typescript` and several files in `pts-javascript`, all had zero native-engine vertices while WASM recorded them correctly), while WASM has always recorded vertices unconditionally.

Fix: the filter now includes any native-language file with at least one function/method definition (checked via the `nodes` table, `kind IN CALLABLE_SYMBOL_KINDS`) — which is what actually determines whether a file can produce vertices, not edge presence.

## Verification

`node scripts/parity-compare.mjs --dataflow` across the full 34-language fixture suite:
- **Before**: `javascript`, `typescript`, `pts-javascript` all diverged on df-vertices (17/20/6 missing vertex rows respectively)
- **After**: 42/42 fixtures OK on df-vertices — the one remaining `jelly-micro` divergence is a pre-existing, unrelated `technique`-label mismatch on `calls` edges (0 node/edge/df-vertex diffs beyond that), filed separately as #2572

The originally-filed issue's own note ("`--dataflow` requires migration v18 — worth checking whether the test DB has that migration applied") turned out not to be the cause — this was a genuine extraction-scoping bug, not a migration issue.

## Test plan
- [x] New integration test `tests/integration/issue-2483-native-dataflow-vertices-leaf-functions.test.ts` — a real dual-function leaf-function fixture built end-to-end through `buildGraph` with the native engine, confirming both functions' param/return vertices are recorded
- [x] Revert-verified: disabling the new filter condition reproduces the pre-fix `[]` (zero vertices) for both new tests
- [x] Existing dataflow integration tests (`dataflow-vertices`, `dataflow-p4-native`, `dataflow-incremental`) still pass (28 passed)
- [x] `npx tsc --noEmit -p .`, `npm run lint`, full `npm test` (5487 passed)
- [x] Full `parity-compare.mjs --dataflow` run across all 34 language fixtures (42/42 OK on df-vertices)

Closes #2483

### DIFF
--- a/src/domain/graph/builder/stages/native-orchestrator.ts
+++ b/src/domain/graph/builder/stages/native-orchestrator.ts
@@ -28,6 +28,7 @@ import { semverCompare } from '../../../../infrastructure/update-check.js';
 import { getOrCreatePerDbChunkStmt } from '../../../../shared/chunked-stmt-cache.js';
 import { normalizePath, TS_NATIVE_CONFIDENCE_FLOOR } from '../../../../shared/constants.js';
 import { toErrorMessage } from '../../../../shared/errors.js';
+import { CALLABLE_SYMBOL_KINDS } from '../../../../shared/kinds.js';
 import { CODEGRAPH_VERSION } from '../../../../shared/version.js';
 import type {
   BetterSqlite3Database,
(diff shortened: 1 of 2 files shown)


carlos-alm:
@greptileai

Issue #2344, opened 2026-08-08
follow-up: lint-skill.sh flags a cross-fence $COUNT violation in fixer/SKILL.md (possible false positive)


carlos-alm:
Discovered while editing `.claude/skills/fixer/SKILL.md` for issue #2160.

Running `lint-skill.sh` against `.claude/skills/fixer/SKILL.md` (also reproduces identically on `origin/main`, unrelated to my change) reports:

```
ERROR: Cross-fence variable: $COUNT assigned in bash block 1, referenced in block 36 without file persistence (Pattern 1)
```

This looks like a false positive: `COUNT` IS correctly persisted via `.codegraph/fixer/count` — block 1 writes `printf '%s\n' "$COUNT" > .codegraph/fixer/count`, and block 36 re-derives it fresh via `COUNT=$(cat .codegraph/fixer/count)` before using it, which is exactly the file-persistence pattern the check is supposed to allow. Possible root cause: the check's per-block "is this var reassigned in this same block" tracking may not correctly recognize `VAR=$(cat ...)` as a valid same-block reassignment when checking whether a LATER line in that same block referencing `$VAR` should be exempted — worth someone tracing through `lint-skill.sh`'s "Check 1: Cross-fence variable usage" logic directly against this exact repro (`bash .claude/skills/create-skill/scripts/lint-skill.sh .claude/skills/fixer/SKILL.md`).

If it turns out to be a genuine cross-fence bug in `fixer/SKILL.md` rather than a linter false positive, that also needs fixing — either way, someone should confirm which it is.

Not fixed here since it's unrelated to #2160's scope (G2 semantic-verification gate) and touches shared lint tooling used by many skills.

PR #2574, opened 2026-08-18
fix(skills): lint-skill.sh no longer masks cross-fence leaks fed via read here-strings

### LINKED ISSUE #2491: lint-skill.sh: blanket 'read ' substring exemption masks genuine cross-fence leaks on read lines
Discovered while addressing Greptile's review on PR #2490 (issue #2344).

`lint-skill.sh`'s Check 1 (cross-fence variable usage) has a reference-checking exemption (currently around the `[ ! "$line" == *'read '* ]`-style condition guarding the `error(...)` call) meant to allow patterns like `VAR=$(cat file)` — genuinely re-deriving a value from a persisted file rather than relying on stale shell state from an earlier bash block.

`read` is included in that same exemption list, but as a **blanket substring check on the whole line**, not tied to whether the specific variable being referenced was actually the destination of that line's `read`. So a line that both (a) binds some unrelated variable via `read` and (b) separately references a different, genuinely-stale variable from an earlier block — e.g. a here-string interpolation — gets fully exempted, silently accepting a real cross-fence bug.

Repro (pre-existing on `origin/main`, independent of PR #2490's fix):

Block 1:
```bash
FOO=hello
```

Block 2:
```bash
read -r BAR <<< "$FOO"
```

`$FOO` here is a genuine cross-fence leak (block 2 relies on block 1's shell state, which does not persist across separate ` ```bash ` fences) but `lint-skill.sh` reports 0 errors, because the line contains the substring `read ` (from `read -r BAR`), which trips the blanket exemption meant for `cat`/`read`-from-file patterns — even though this `read`'s own destination is `BAR`, not `FOO`.

Root cause: the exemption is a per-line substring match (`*'read '*`) rather than checking whether the *specific referenced variable* is the one actually bound by that line's `read`/`cat`.

Not fixed as part of #2344/PR #2490 — that PR's scope was the assignment-registration side (recognizing `read` as a binding mechanism so genuine same-block rebindings aren't flagged); this is a separate, pre-existing false-negative in the reference-checking side, discovered via Greptile's review while verifying the #2344 fix didn't introduce it.

carlos-alm:
## Summary
- `lint-skill.sh`'s Pattern-1 cross-fence variable check exempted any line containing the substring `read ` from being flagged, without checking whether the variable in question was actually the destination `read` bound — so a stale variable fed as the *input* to a `read` here-string (`read -r BAR <<< "$FOO"`) slipped through undetected.
- A second, independent blanket exemption (`< ` substring match, meant to detect genuine file redirection) made this worse: `<<<`'s own text contains a trailing `< `, so even narrowing the `read` exemption alone wouldn't have caught the exact repro.
- Adds `extract_read_dest_vars`/`is_read_dest_of_line` (checks whether `$var` is genuinely the destination a `read` line binds) and `has_file_redirect_in` (a regex distinguishing a real single-`<` file redirect from `<<`/`<<<`), replacing both blanket substring checks.

## Test plan
- [x] New regression test `tests/unit/lint-skill-here-string-2491.test.ts`: exact issue repro + here-string-no-read case fail before the fix and pass after (revert-verified against the pre-fix script).
- [x] Existing `read`-binding exemption (#2344) and genuine file-redirection exemption still pass — no regressions.
- [x] Ran the updated linter against all 21 real `SKILL.md` files in the repo — output identical to before the fix (zero new false positives).
- [x] `npm test` (full suite, 92 files / 1929 tests) and `npm run lint` both pass.

Closes #2491

### DIFF
--- a/.claude/skills/create-skill/scripts/lint-skill.sh
+++ b/.claude/skills/create-skill/scripts/lint-skill.sh
@@ -78,24 +78,36 @@ collect_assignments() {
 }
 # Strips a trailing, unquoted `# comment` from a read command's argument
 # text. A `#` is only a genuine shell comment when it isn't inside an open
-# double-quoted string, so this walks the string quote-span by quote-span
-# (checking for `#` only in the text BETWEEN quotes) rather than scanning
-# for the first `#` outright (#2491 Greptile round 3: `read -r NUM < f
-# # MAX_LIMIT` left `# MAX_LIMIT` for the destination grep to pick up
-# `MAX_LIMIT` as a spurious destination).
+# quoted string (double OR single — a literal `#` inside a single-quoted
+# redirect target, e.g. `<<< '#tag value'`, is data, not a comment marker,
+# and must not truncate away a genuine destination that follows it (#2491
+# Greptile round 4)), so this walks the string quote-span by quote-span —
+# whichever quote character opens first — checking for `#` only in the
+# text BETWEEN quotes, rather than scanning for the first `#` outright
+# (#2491 Greptile round 3: `read -r NUM < f # MAX_LIMIT` left
+# `# MAX_LIMIT` for the destination grep to pick up `MAX_LIMIT` as a
+# spurious destination).
 strip_trailing_comment() {
-  local s="$1" out="" seg
-  while [[ "$s" == *'"'* ]]; do
-    seg="${s%%\"*}"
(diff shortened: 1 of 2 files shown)


carlos-alm:
@greptileai

carlos-alm:
@greptileai

carlos-alm:
@greptileai

carlos-alm:
@greptileai

Issue #2358, opened 2026-08-08
follow-up: extractParams gives every name from one grouped-param wrapper the same paramIndex


carlos-alm:
## Problem

Discovered while fixing #2182 and adding Dart's `optional_formal_parameters` handling (`extractDartParamName` in `src/ast-analysis/rules/b2.ts`) — a grammar-level "one child node expands to multiple logical parameter names" case, e.g. Dart's `{int times, bool loud}` named-parameter group.

`extractParams`'s generic loop (`src/ast-analysis/visitor-utils.ts`) increments `index` once per CHILD of the parameter list, not once per NAME extracted:

```ts
for (const child of paramsNode.namedChildren) {
  const names = extractParamNames(child, rules);
  for (const name of names) {
    result.push({ name, index });   // every name from this ONE child shares `index`
  }
  index++;
}
```

For `int greet(String name, {int times = 1, bool loud = false})`, this currently produces `paramIndex: 0` for `name`, then `paramIndex: 1` for BOTH `times` and `loud` (should be `1` and `2`). This mechanism is generic (not Dart-specific) — it would affect any language whose `extractParamName`/wrapper-unwrapping logic legitimately expands one wrapper child into multiple sibling names sharing one nominal "slot."

## Impact

Low severity in practice for Dart specifically, since named/optional parameters are almost always referenced by name at call sites, not position — but the recorded `paramIndex` data is still incorrect, and any consumer that does rely on positional index for these params (or for another language hitting the same pattern) would get wrong results.

## Fix direction

Change `extractParams` to increment `index` per extracted NAME rather than per child node — i.e. `index++` inside the inner `for (const name of names)` loop instead of after it. Needs verification against every other language currently using `extractParamName`/wrapper types to confirm none of them rely on the current "one index per child" behavior (e.g. destructuring patterns where multiple bound names conceptually share one argument position).

Filed per this repo's scope-discipline convention — out of scope for #2182's sibling-body architecture fix.

PR #2575, opened 2026-08-18
fix(go): give each name in a grouped param declaration its own index

### LINKED ISSUE #2501: follow-up: Go's grouped `a, b int` parameter declarations share one paramIndex in both engines
## Problem

Discovered while fixing #2358 (Dart's `optional_formal_parameters` sharing one `paramIndex` across multiple named-parameter-group children).

Go's grammar allows a single `parameter_declaration` node to declare multiple comma-separated names against one shared type: `func f(a, b int, c string)`. Per `node_modules/tree-sitter-go/src/node-types.json`, `parameter_declaration` has a `name` field with `multiple: true` (both `a` and `b` are `name`-field children) plus a single `type` field.

Both engines extract all names from one such node in a single pass, and the outer per-child loop increments `index` only once per node (the exact bug pattern #2358 fixes for Dart) — so `a` and `b` above both get `paramIndex: 0`, and `c` gets `paramIndex: 1`, when the correct indices are `0`, `1`, `2` respectively (each is a genuinely independent, individually-positioned parameter, not a destructured binding of one argument slot):

- TS: `extractParamName` hook in `src/ast-analysis/rules/go.ts` (`parameter_declaration` branch) returns `names.length > 0 ? names : null` — multiple names from one child in the generic `extractParams` loop (`src/ast-analysis/visitor-utils.ts`).
- Rust: `extract_params_go` in `crates/codegraph-core/src/ast_analysis/dataflow.rs` has the identical shape — collects every `identifier` child into one `Vec<String>` returned for one `parameter_declaration` node, consumed by `extract_params`'s per-child increment.

## Why this isn't fixed by #2358's mechanism directly

#2358 introduces `groupedParamTypes` (TS-only, since Dart dataflow doesn't exist in the Rust engine yet — tracked separately by #2359): when a child's type is in this set, the outer loop iterates that child's OWN named children as separate slots instead of calling `extractParamNames` on the whole node.

Go's shape doesn't fit that directly — `a` and `b` are not separate child NODES, they're two `identifier`s under the SAME `parameter_declaration`'s `name` field (`multiple: true`), sitting alongside a `type` field that must NOT be treated as its own slot. Naively marking `parameter_declaration` as a `groupedParamTypes` entry and iterating all of its `namedChildren` would incorrectly also try to extract a slot from the `type` node.

Fixing this needs either:
- A field-aware variant (e.g. read `childrenForFieldName('name')` specifically, one slot per name, skip `type`), or
- Restructuring `go.ts`'s hook so each identifier is walked as an independent grandchild before `extractParamName` currently intercepts the whole `parameter_declaration` node.

Needs the equivalent fix mirrored in `extract_params_go` (Rust) for dual-engine parity, since — unlike Dart — Go already ships in both engines today.

Filed per this repo's scope-discipline convention — out of scope for #2358's Dart-specific `groupedParamTypes` fix.

carlos-alm:
## Summary
- tree-sitter-go's `parameter_declaration` node shares one type across multiple comma-separated names (`func f(a, b int)` — both `a` and `b` are `name`-field children of the same node, per `node-types.json`).
- Both engines extracted all names from one such node in a single pass, while the outer per-child loop incremented its index only once per node — so `a` and `b` both got `paramIndex: 0` instead of `0` and `1` respectively.
- Mirrors the `groupedParamTypes`/`grouped_param_types` mechanism from #2358 (originally built for Dart's `optional_formal_parameters`): the outer loop now iterates a `parameter_declaration`'s own children as separate slots — each `identifier` resolves via the existing generic `paramIdentifier`/`param_identifier` handler and gets its own index, while the `type_identifier` yields no name and consumes no index.
- Verified the native and WASM grammars produce the identical AST shape for this construct before implementing (both: `parameter_declaration name: (identifier) name: (identifier) type: (type_identifier)`).

## Test plan
- [x] New Rust tests (`crates/codegraph-core/src/ast_analysis/dataflow.rs`, `go_tests` module): grouped multi-name indices, single-name declarations still correct, variadic parameter still indexes correctly after a grouped group.
- [x] Strengthened an existing TS test (`tests/parsers/dataflow-go.test.ts`, "extracts multi-name parameters") that exercised this exact scenario but never asserted `paramIndex` — added two new tests that do.
- [x] Revert-verified on both engines: reverting just the production fix (keeping the new tests) makes them fail with the exact bug (`b` getting `paramIndex: 0` instead of `1`); restoring the fix makes them pass.
- [x] `cargo test -p codegraph-core` (1125 tests), `cargo fmt --check`, `cargo clippy` all clean.
- [x] Full `npx vitest run` (5504 tests) and `npm run lint` clean.
- [x] `node scripts/parity-compare.mjs --langs go --dataflow` — native (freshly rebuilt) vs WASM identical, including the real fixture's own grouped-parameter function (`CreateUser(id, name, email string)`).

Closes #2501

### DIFF
--- a/crates/codegraph-core/src/ast_analysis/dataflow.rs
+++ b/crates/codegraph-core/src/ast_analysis/dataflow.rs
@@ -258,7 +258,16 @@ static GO_DATAFLOW: DataflowRules = DataflowRules {
     param_list_field: "parameters",
     param_identifier: "identifier",
     param_wrapper_types: &[],
-    grouped_param_types: &[],
+    // A `parameter_declaration` can share one type across multiple
+    // comma-separated names (`func f(a, b int)` — both `a` and `b` are
+    // `name`-field children of the SAME node, per tree-sitter-go's
+    // node-types.json). This makes `extract_params` iterate this node's own
+    // named children as separate slots instead of calling
+    // `extract_params_go` on the whole node — each `identifier` name
+    // resolves via the generic `param_identifier` handler and gets its own
+    // index, while the trailing `type_identifier` yields no name and
+    // consumes no index (issue #2501).
+    grouped_param_types: &["parameter_declaration"],
     default_param_type: None,
     rest_param_type: None,
     object_destruct_type: None,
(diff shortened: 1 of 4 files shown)


carlos-alm:
@greptileai

PR #2585, opened 2026-08-19
fix(hooks): normalize-ifs.mjs generalizes :+/+ whitespace bypass beyond IFS

### LINKED ISSUE #2558: guard-git.sh: any normally-set variable (not just IFS) can inject a literal whitespace token-boundary via ${VAR:+ } / ${VAR+ }
## Summary

Found while fixing #2451 (Greptile review, round 3 on PR #2557).

`normalize-ifs.mjs` now correctly recognizes `${IFS:+ }`/`${IFS+ }` (bash's alternate-value expansion, restricted to all-whitespace substituted content) as producing a literal space, the same as the whole-variable `${IFS}`/`$IFS` form. But this operator's behavior has nothing to do with the variable being named `IFS` specifically — bash's `${VAR:+word}` substitutes `word` whenever `VAR` is set and non-null (the normally-true case for almost any commonly-set variable), regardless of what `VAR`'s own value actually is.

This means `${HOME:+ }`, `${PWD:+ }`, `${PATH:+ }`, or any other variable that's normally set in a shell session works **identically** to `${IFS:+ }` for this specific purpose — none of them reference IFS's value at all, they just use an arbitrary set variable as the condition and inject a literal space as the substituted "alternate" text.

## Why this is a separate problem from #2451

`normalize-ifs.mjs` is (by design and by name) scoped to recognizing IFS-specific obfuscation. A per-variable-name normalizer like this one can never fully close the "any normally-set variable + `:+`/`+` with a whitespace alternate value" class, since the variable name in the bypass is not fixed — an attacker (or a sufficiently obfuscation-aware agent) could pick any variable known to be set in the target shell.

## Suggested fix

This needs a fundamentally different detection strategy than name-based normalization, e.g.:
- A regex matching the general SHAPE `${<any-identifier>:+<all-whitespace>}` / `${<any-identifier>+<all-whitespace>}` regardless of the specific variable name (broader coverage, but still name-shape-based rather than semantically evaluating whether that variable is actually set at execution time — a variable that's NOT normally set would make this a false positive, though for guard-purposes a false positive here is the safe direction to err in).
- Or, a more fundamental rethink of guard-git.sh's approach: instead of static regex scanning of command TEXT, actually invoke a shell dry-run/AST parse to see the real resulting argv bash would produce, sidestepping the entire class of "does this obfuscation technique produce whitespace" heuristics.

The first option is a bounded, incremental improvement; the second is a much larger architectural change (matching CLAUDE.md's "prioritize best architecture" guidance if pursued) that would also close the `${IFS/pattern/replacement}` and `${IFS@Q}`-style gaps noted in `normalize-ifs.mjs`'s own comments, and any other future bash expansion trick, without needing per-case regex patches.

## Related

- #2451 (the original issue this was found while fixing)
- `normalize-ifs.mjs`'s own doc comment documents this as a known, accepted gap pending this follow-up.

carlos-alm:
## Summary
- Found while fixing #2451 (Greptile review round 3 on PR #2557): `normalize-ifs.mjs` correctly recognizes `\${IFS:+ }`/`\${IFS+ }` as producing a literal space, but this operator's behavior has nothing to do with the variable being named `IFS` specifically — bash's `\${VAR:+word}` substitutes `word` whenever `VAR` is set and non-null, regardless of `VAR`'s own value. `\${HOME:+ }`, `\${PWD:+ }`, `\${PATH:+ }`, or any other normally-set variable works identically.
- A per-variable-name normalizer can never fully close this class, since the variable name in the bypass isn't fixed — this issue's own text lays out two options: a bounded regex generalizing the shape regardless of name (what this PR does), or a much larger architectural rethink (actual shell parsing instead of static regex heuristics, tracked separately as a known, accepted broader limitation).
- The other three replacements in this file (`\${IFS}`, `\$IFS`, and the substring-expansion form) stay `IFS`-specific, since they extract from IFS's own known default value (space/tab/newline) — that reasoning doesn't generalize to other variables, whose actual values aren't known statically. Only the `:+`/`+` alternate-value replacement — whose substituted text is entirely independent of the named variable's value — now matches any bash identifier shape (`[A-Za-z_][A-Za-z0-9_]*`) in that position.
- Kept `docs/examples/claude-code-hooks/normalize-ifs.mjs` byte-identical to `.claude/hooks/normalize-ifs.mjs`, as enforced by `tests/unit/hook-guard-git-clean.test.ts`.

Closes #2558

## Test plan
- [x] New tests in `tests/unit/hook-guard-git-ifs-bypass.test.ts`: blocks `git\${HOME:+ }reset` and `git\${PWD+ }reset` (generalized bypass via non-IFS variables), does not flag `\${SOME_VAR:+x}` (non-whitespace substituted text, any name).
- [x] Revert-verify: temporarily reverted the regex to the IFS-only form and confirmed both new tests fail with exactly the pre-fix symptom (`expected false to be true`); restored the fix and all pass again.
- [x] Full guard-git test suite (`hook-guard-git-ifs-bypass`, `hook-guard-git-clean`, `hook-guard-git-branch-validation`, `hook-guard-git-commit-cwd-fallback`) — 82/82 pass, confirming the byte-identical-copies check still holds.
- [x] `npm run lint` clean.
- [x] Full `npx vitest run`: 5528/5528 tests pass (344 test files, +3 new).

### DIFF
--- a/.claude/hooks/normalize-ifs.mjs
+++ b/.claude/hooks/normalize-ifs.mjs
@@ -111,6 +111,18 @@
 // be set in a given shell, rather than trying to track which variables are
 // actually set.
 //
+// Also matches bash's SPECIAL parameters in that position — `?` (exit
+// status), `$` (PID), `#` (positional-parameter count), `-` (current shell
+// option flags), `!` (last background PID), and a bare digit sequence
+// (positional parameters, `${10:+ }` etc.) — verified directly against real
+// bash (Greptile review): `${?:+ }`/`${$:+ }`/`${#:+ }`/`${-:+ }` all
+// substitute the whitespace word exactly like an ordinary variable would,
+// and `?`/`$`/`#`/`-` are always set in any shell (unlike `!`/digit
+// parameters, which depend on whether a job has been backgrounded or
+// positional arguments are present — matched anyway, erring toward
+// normalizing). None of these characters overlap with the identifier
+// alternative above, so a single alternation covers both without ambiguity.
+//
 // Deliberately does NOT generalize the other three replacements below to
 // "any `${VAR<operator>...}`" — only `:+`/`+` generalizes across variable
 // names, because only its substituted text is entirely independent of the
@@ -150,7 +162,7 @@ process.stdin.on('end', () => {
   const normalized = input
     .replace(/\$\{IFS\}/g, ' ')
     .replace(/\$\{IFS: *(?:0*[0-2]|-0*[1-3]|-0+)(?::0*[1-9]\d*)?\}/g, ' ')
-    .replace(/\$\{[A-Za-z_][A-Za-z0-9_]*:?\+[ \t]+\}/g, ' ')
+    .replace(/\$\{(?:[A-Za-z_][A-Za-z0-9_]*|[0-9]+|[?$!#@*-]):?\+[ \t]+\}/g, ' ')
     .replace(/\$IFS(?![A-Za-z0-9_])/g, ' ');
   process.stdout.write(normalized);
 });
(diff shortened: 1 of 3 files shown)


carlos-alm:
@greptileai

Issue #2072, opened 2026-07-09
edges table has no content-level uniqueness constraint despite comments claiming otherwise


carlos-alm:
Found while investigating a Greptile P2 comment on PR #1972.

**The issue:** The `edges` table schema (`crates/codegraph-core/src/db/connection.rs:39-48`, `src/db/migrations.ts:26-35`) has no `UNIQUE` constraint on `(source_id, target_id, kind)` or any other content combination — only the autoincrement `id` primary key. Every edge-insertion call site uses `INSERT OR IGNORE`, and several comments assert this deduplicates edge content:

- `crates/codegraph-core/src/domain/graph/builder/stages/detect_changes.rs:608-609`: "INSERT OR IGNORE silently swallows duplicate-row constraint errors" (for edges specifically)
- `crates/codegraph-core/src/domain/graph/builder/stages/insert_nodes.rs:137`: groups edges together with nodes' real UNIQUE constraint in the same comment

These comments are incorrect for edges — `nodes` has a real `UNIQUE(name, kind, file, line)` constraint (`connection.rs:37` / `migrations.ts:24`), so `OR IGNORE` genuinely dedupes there. For `edges`, `OR IGNORE` has nothing to suppress on a duplicate content row; it would only fire on a (now essentially unreachable, since ids are freshly resolved against `nodes` moments earlier) FK violation.

**Why this hasn't caused visible bugs:** every current call site that could theoretically produce duplicate edges is protected by a purge-before-insert pattern elsewhere (full builds wipe the whole `edges` table first; incremental builds delete-then-reinsert per changed file). So `OR IGNORE` is currently dead/misleading code, not an active source of duplicate rows — but a future change that skips or narrows the purge step would silently start producing duplicate/mixed edge rows, since there is no actual constraint to catch it.

**Suggested fix (either one):**
1. Add a real `UNIQUE(source_id, target_id, kind)` index (or similar) so `OR IGNORE` does what its surrounding comments already claim, or
2. Fix the comments to stop claiming deduplication happens, so a future reader doesn't rely on protection that isn't there.

Out of scope for PR #1972 (which only threads `Result` through existing insert paths, per #1827) — filed as a follow-up per CLAUDE.md's scope-discipline rule.

Issue #1936, opened 2026-07-07
Extend native Leiden port to cover full LeidenOptions surface (directed, CPM, weights, candidate strategies)


carlos-alm:
> **Status: closed as won't-do for the remainder.** Per-edge weights shipped (#2593); the guard that made the whole thing safe shipped in #2592; directed / CPM / confidence-weighting were measured and declined. The numbers and the replacement trigger are below. See the [closing comment](https://github.com/optave/ops-codegraph-tool/issues/1936#issuecomment-5362350651) for the full analysis.
>
> **Do not re-verify this by grepping call sites.** The original gate ("only take this on if/when a real caller needs one of these knobs") was unsatisfiable by construction — the knobs are unreachable because `LouvainOptions` does not expose them, and exposing them *is* the work being gated. Four consecutive `/fixer` passes re-derived the same "not met" conclusion from the same grep before this was spotted.

## Background

#1804 ported Leiden community detection to Rust (`crates/codegraph-core/src/graph/algorithms/leiden.rs`) so the native and JS/WASM engines run the same algorithm instead of native running classic Louvain and JS running Leiden. That port was deliberately scoped to the option surface reachable through `louvainCommunities`/`LouvainOptions` (`src/graph/algorithms/louvain.ts`).

This issue tracked extending it to the rest of the TS reference implementation's surface (`src/graph/algorithms/leiden/*`).

## What shipped

**#2592 — the actual bug.** `louvainCommunities` handed native `graph.toEdgeArray()`, which drops every edge attribute at the FFI boundary, and passed no node attributes at all, while the JS reference resolved per-edge `weight` and per-node `size` from attributes. Any graph carrying either would have had the two engines optimize *different* weighted graphs and return different partitions — silently. `nativeParityBlocker` now detects that up front and defers to the JS reference. Latent rather than live at the time (no in-tree builder set those attrs), but silent, so it was closed first.

[...]

PR #2593, opened 2026-08-20
feat(communities): native Leiden reads per-edge weights

carlos-alm:
> **Stacked on #2592** — review that first; the base will retarget to `main` automatically once it merges.

## What this does

Adds `leidenWeightedCommunities` to the native binding, so per-edge weights reach native Leiden instead of being flattened to 1.0 at the FFI boundary. This is item 3 of #1936's plan, and it removes edge weights from `nativeParityBlocker`'s reasons to fall back — weighted graphs now run on whichever engine is available and produce byte-identical results.

## Design notes

**A new napi struct, not a field on `GraphEdge`.** `GraphEdge` is also the input to `detectCycles`, `shortestPath`, `fanInOut` and `bfs`, none of which read a weight. An inert field there would let a caller set a weight on a `shortestPath` edge and have it silently ignored — the same class of quiet-wrong-answer bug this change exists to close.

**One implementation of the attribute-to-number rule.** `resolveEdgeWeight` (exported from the Leiden adapter, and used by the adapter's own call sites) resolves weights for both engines, so native and JS cannot drift apart on what a graph weighs. Rust owns only the two mappings that cannot cross the FFI boundary: absent means 1.0, NaN means 0.0.

**Version skew is handled in both directions.** `leidenCommunities` stays exported, defined as exactly the all-weights-1.0 case, for a newer addon loaded by older JS; a Rust test pins the two entry points to identical output and the TS suite keeps it covered so it cannot rot. In the reverse case — older addon, newer JS — the weighted export is missing and `nativeParityBlocker` falls back to JS for weighted graphs rather than sending weights to a binding that cannot read them.

## Also fixed: `buildTemporalGraph` was not actually weighted

It advertises a graph "weighted by Jaccard similarity" but stored the score only under a `jaccard` key that **neither engine reads** — so every weight-aware consumer saw a uniformly weighted graph and the entire co-change signal was silently discarded. It now sets `weight` too (`jaccard` kept for existing readers). Measured on this repo: 628 co-change pairs spanning Jaccard 0.02–1.0.

## What is deliberately NOT wired up, having measured it

**Call-site multiplicity — the premise was false.** Repeated call sites are not recoverable from the DB: `idx_edges_content_unique` (#2072) makes duplicate `(source, target, kind, confidence, …)` rows impossible in both engines. This repo's graph has **10712 `calls` rows for 10711 distinct pairs** (max multiplicity 2, from two resolution techniques on one pair — not multiplicity). Recording it needs a schema column, not a weight attribute.

**Weighting by resolution `confidence` — measured, no value.** Already available on `CallEdgeRow` for 22% of call edges, so it was the cheapest candidate. It does not improve community quality on this repo:

| graph | modularity | directory purity |
|---|---|---|
| file-level, unweighted | 0.5071 | 0.7771 |
| file-level, confidence-weighted | 0.5077 | **0.7725** |
| function-level, unweighted | 0.5879 | 0.9647 |
| function-level, confidence-weighted | 0.5912 | **0.9641** |

Confidence clusters at 0.9–1.0, so it perturbs the graph without restructuring it. Not enabling a behavior change by default on a measurement that says "no better, slightly worse alignment". (Modularity is not comparable across weightings — it is a different objective function — which is why directory purity is the readout.)

Node `size` remains native-blind and still routes to the JS reference: nothing in-tree sets it, so there is no caller to verify a port against, and it is not inert — the JS reference sorts communities by total size when assigning final IDs.

## Verification

- **5518** vitest tests pass (0 failures); **1132 + 3 new** Rust tests pass
- `cargo fmt --check` and `cargo clippy --workspace --all-targets -D warnings` clean; `tsc --noEmit` and `biome check` clean
- Native/JS **byte-for-byte** parity asserted across weighted bridges, weighted reciprocal edges (the undirected averaging path), weighted self-loops, zero/NaN weights, and a weighted multi-level graph across 3 resolutions and 3 seeds
- Verified against a locally built addon, not the published one — note that `napi build --release` writes `codegraph-core.node` while the loader looks for `codegraph-core.darwin-arm64.node`, so a local build is silently ignored unless renamed (filed separately)
- Local runs are on Node 24 while CI pins 22

Refs #1936

### DIFF
--- a/crates/codegraph-core/src/graph/algorithms/leiden.rs
+++ b/crates/codegraph-core/src/graph/algorithms/leiden.rs
@@ -10,22 +10,35 @@
 //!
 //! ## Scope
 //!
-//! This port covers exactly the option surface reachable through
+//! This port covers the option surface reachable through
 //! `louvainCommunities`/`LouvainOptions`
 //! (`src/graph/algorithms/louvain.ts`): undirected graphs, modularity
 //! quality (not CPM), the default "neighbors" candidate strategy,
-//! `refine: true` (always), uniform node size (1.0) and edge weight (1.0
-//! per edge — `GraphEdge` carries no weight field), no
-//! `maxCommunitySize`/`fixedNodes`/`preserveLabels` overrides. These are the
-//! *only* knobs `louvainCommunities` ever threads through to either engine
-//! (see `LouvainOptions` and `louvainJS()`'s call into `detectClusters`).
+//! `refine: true` (always), per-edge weights, uniform node size (1.0), no
+//! `maxCommunitySize`/`fixedNodes`/`preserveLabels` overrides.
 //!
-//! The TS `leiden/` directory's directed-graph mode, CPM quality function,
-//! alternate candidate strategies (all/random/random-neighbor),
-//! `allowNewCommunity`, `fixedNodes`, and `preserveLabels` knobs are **not**
-//! ported — they are unreachable from this binding today. Issue #1936
-//! tracks porting that remaining surface if a caller ever needs to drive
-//! native Leiden with it.
+//! Per-edge weights arrive through `leiden_weighted_communities` and are
+//! resolved *on the TypeScript side* (`resolveEdgeWeight` in
+//! `src/graph/algorithms/leiden/adapter.ts`, shared with the JS reference's
+//! own adapter) so that the rule mapping an edge attribute to a number has
+//! exactly one implementation across both engines. This binding therefore
+//! takes already-resolved numbers and does not reimplement that rule — see
+//! `resolve_weight` below. The older `leiden_communities` entry point is
+//! retained for version skew (newer addon, older JS) and is exactly the
+//! all-weights-1.0 case.
+//!
+//! Node size is still uniform 1.0: nothing in-tree attaches a `size`
+//! attribute, so there is no caller to verify a port against. Sizes are not
+//! inert under modularity — `compact_community_ids` sorts communities by
+//! total size to assign final IDs — so `nativeParityBlocker` in `louvain.ts`
+//! routes any size-bearing graph to the JS reference rather than letting the
+//! two engines label communities differently.
+//!
+//! The TS `leiden/` directory's CPM quality function, alternate candidate
+//! strategies (all/random/random-neighbor), `allowNewCommunity`,
+//! `fixedNodes`, and `preserveLabels` knobs are **not** ported — they are
+//! unreachable from this binding today. Issue #1936 tracks porting that
+//! remaining surface if a caller ever needs to drive native Leiden with it.
 //!
 //! ## Determinism
 //!
@@ -61,6 +74,24 @@ use crate::types::GraphEdge;
 // napi-facing types + entry point
 // ════════════════════════════════════════════════════════════════════════
 
+/// A graph edge carrying an already-resolved weight.
+///
+/// Distinct from the shared `GraphEdge` on purpose: `GraphEdge` is also the
+/// input to `detect_cycles`, `shortest_path`, `fan_in_out` and `bfs`, none of
+/// which read a weight. Adding an inert field there would let a caller set a
+/// weight on, say, a `shortest_path` edge and have it silently ignored —
+/// which is the same class of quiet-wrong-answer bug that per-edge weight
+/// support exists to close.
+#[napi(object)]
+#[derive(Debug, Clone)]
+pub struct LeidenWeightedEdge {
+    pub source: String,
+    pub target: String,
+    /// Resolved edge weight; `None` means 1.0 (see the module doc — the
+    /// attribute-to-number rule lives on the TypeScript side).
+    pub weight: Option<f64>,
+}
+
 #[napi(object)]
 #[derive(Debug, Clone)]
 pub struct LeidenCommunityAssignment {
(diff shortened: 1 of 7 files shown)


carlos-alm:
Fixed in 388e9ae7 — good catch, and it turned out to be more than a wording slip.

You're right that `+w || 0` preserves `Infinity`/`-Infinity` (they're truthy); only `NaN`, zero and non-numeric values collapse. Verified:

```
+Infinity || 0  // Infinity
+NaN      || 0  // 0
```

Rather than just correct the sentence, I checked what a surviving infinite weight actually does: it flows into the adapter's `strength`/`totalWeight` accumulators and poisons the modularity sums via `Infinity - Infinity` — the exact failure the NaN branch exists to prevent, reached by a different route. So the comment was hiding a real gap, not only misdescribing one.

I did not clamp it in this PR: both engines share the behavior (the native `resolve_weight` mirrors the rule deliberately, and this PR asserts native/JS parity on zero and NaN), nothing in-tree can produce an infinite weight, and clamping would change the vendored ngraph.leiden reference's numeric semantics and need a matching native change. Filed as #2597 with the suggested fix, and the comment now points there so it reads as a tracked gap rather than intended design.

@greptileai

carlos-alm:
**CI note — the benchmark-gate failure is not from this PR.**

`Pre-publish benchmark gate` failed with:

```
diffImpact latency: 10.1 → 25.3 (+150%, threshold 25%)
```

That flipped between two commits on this branch whose only difference is **one doc comment**:

| commit | diff | CI |
|---|---|---|
| `289b5e27` | all the Rust + TS + test changes | success |
| `388e9ae7` | one doc comment, zero executable lines | failure |

```
git diff 289b5e27 388e9ae7 -U0 | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' \
  | grep -vE '^[+-]\s*(\*|/\*\*|\*/|//)'      # → no output
```

Nothing in this PR touches `diffImpact` — the changes are confined to Leiden community detection, the shared weight resolver, and the co-change builder.

Root cause looks like measurement noise rather than a real regression, and I filed #2598 with the mechanism: `benchDiffImpact` times `diffImpactData(…, { staged: true })`, which spawns a `git` subprocess via `runGitDiff` **inside the timed loop**, and unlike `benchDepths` it applies no `WARMUP_RUNS` — so it is median-of-5 samples each dominated by process spawn, gated at 25% on a ~10ms baseline.

`CI Testing Pipeline` is just the aggregate gate failing on the above, not a second independent failure. I have re-run the failed jobs.

Everything else is green, including engine parity on all three OSes and the Node 22 test matrix.

carlos-alm:
**Correcting my previous comment.** I called this measurement noise; that diagnosis was wrong. The measurement is reproducible — the re-run gave 24.4 vs 25.3. The real cause is a **stale baseline**, and my PR is still exonerated, but by better evidence than I first gave.

`diffImpact` is elevated against its recorded 10.1ms baseline on **every branch I checked**, including three unrelated PRs:

| branch | measured | vs baseline | delta | gate |
|---|---|---|---|---|
| `fix/issue-2521` | 15.1 | +50% | 5.0ms | warn |
| `fix/issue-2524` | 18.7 | +85% | 8.6ms | warn |
| `fix/issue-2512` | 19.8 | +96% | 9.7ms | warn |
| #2592 (guard only) | 17.5 | +73% | 7.4ms | warn |
| this PR @ `289b5e27` | 15.8 | +56% | 5.7ms | warn |
| this PR @ `388e9ae7` | 25.3 | +150% | 15.2ms | **FAIL** |
| `388e9ae7` re-run | 24.4 | +142% | 14.3ms | **FAIL** |

Nothing measured it below +50%. Note #2592 — the guard-only PR, which shares no code path with this one — shows the same elevation, and neither PR touches `diffImpactData`.

**Why it fails only sometimes:** `MIN_ABSOLUTE_DELTA = 10` (`regression-guard.test.ts:193`) requires a ≥10ms absolute delta before flagging. Against a 10.1ms baseline with a true value in the 15–25ms band, the delta straddles that floor — 5.7ms passes, 14.3ms fails. So the verdict is decided by where jitter lands, which is why identical code landed on both sides of it here.

I have updated #2598 with the corrected diagnosis; the primary fix is to re-record the baseline, not just to add warmup. The 15% warning has apparently been firing on every PR for some time.

I am not going to keep re-running this — it would pass eventually, which is exactly the wrong reason for a gate to go green. Flagging it for a maintainer call on whether to merge past it or fix #2598 first.

@greptileai

carlos-alm:
**Blocked on the benchmark gate, fix is in #2599.**

This PR's only failing check is `diffImpact latency: 10.1 → 24.4` (`CI Testing Pipeline` is just the aggregate gate reporting it). I traced that to the benchmark rather than to this change, and fixed it in #2599:

- the `git` subprocess was inside the timed region (~8ms of a ~9ms call), and
- the probe edit landed outside every function, so `diffImpactData` short-circuited at `affectedFunctions: 0` before doing the work the metric guards.

Together the analysis was ~0.5ms of a ~9ms measurement. After #2599 the metric reads **0.6ms (wasm) / 0.7ms (native)** with `affectedFunctions: 1` — and the two engines agree, which is what confirms the 2.4× native/wasm gap here was spawn cost, not a query regression in this PR.

Evidence that it is not this PR: the same elevation shows on `fix/issue-2512` (9.9/19.8), `fix/issue-2521` (9.5/15.1), `fix/issue-2524` (15.2/18.7) and on #2592 — none of which touch community detection. Details in #2598.

**Suggested sequencing:** merge #2599 first, then I will merge `main` into this branch and the gate will pass on its own. I have deliberately not re-run this PR's failed job — it would eventually go green by chance, which is the wrong reason for a gate to pass and would bury the fact that this affects everyone's PRs.

Happy to merge `main` in as soon as #2599 lands — just say the word.

carlos-alm:
@greptileai

PR #2599, opened 2026-08-21
fix(bench): make the diffImpact metric measure diff-impact analysis

### LINKED ISSUE #2598: diffImpact benchmark measures git subprocess spawn, not graph work — verdict decided by the 10ms floor
> **Two earlier diagnoses in this issue were wrong and are corrected below.** First I called it measurement noise (the re-run reproduced, so no). Then I called it a stale baseline and recommended re-recording it (that would enshrine the defect — see below). The evidence now points somewhere more specific, and it changes the fix.

## Summary

`diffImpact` is ~90% `git` subprocess spawn, so the metric barely measures codegraph. The native variant pays systematically more spawn cost than the wasm variant on CI, and the resulting gap straddles `MIN_ABSOLUTE_DELTA`, so identical code passes or fails at random.

## What the gate reported

On PR #2593:

```
AssertionError: Benchmark regressions exceed threshold:
  diffImpact latency: 10.1 → 24.4 (+142%, threshold 25%)
```

## It is not a product regression — ruled out point by point

- **The code path is untouched.** No commit in the 32 since `v3.17.0` modifies `src/domain/analysis/diff-impact.ts`, `src/domain/queries.ts`, or `src/domain/analysis/impact.ts`.
- **Reproduced locally at baseline.** Replicating the benchmark exactly (append a comment to the hub file `src/domain/graph/builder/pipeline.ts`, stage it, `{ staged: true, depth: 3, noTests: true }`, median of 12 after 4 warmups): **native 9.1ms, wasm 9.2ms**, both against the recorded 10.1ms baseline, both with `affectedFunctions: 0`. With a real diff (`ref: HEAD~1`, 11 affected functions): native 14.2ms, wasm 14.3ms. No engine gap locally, on either code path.
- **The engines are not producing different graphs.** CI artifact for the failing run: nodes 20812 vs 20812; edges 45499 (native) vs 45789 (wasm) — native slightly *fewer*; DB bytes 47.4MB vs 48.3MB. Both report `affectedFunctions: 0, affectedFiles: 0`, i.e. identical (zero) query work.
- **Not indexes or planner statistics.** Both engines' DBs carry the same 67 indexes (byte-identical name list) and neither has `sqlite_stat1`.
- **Not a leftover WAL** inflating each `openReadonlyOrFail`. Neither engine leaves a `-wal`/`-shm` behind after a build.
- **Every other native query is faster than wasm** in the same artifact: fnDeps 11.4 vs 13.8, queryTime 10.6 vs 15, fnImpact/path/roles comparable. Only `diffImpact` inverts.

## What it actually is

**1. The metric is dominated by process spawn.** `benchDiffImpact` times `diffImpactData(…, { staged: true })`, which calls `runGitDiff` → `execFileSync('git', …)` (`src/domain/analysis/diff-impact.ts:50`). Measured locally, `git diff --staged --unified=0` alone is **8.0–8.6ms** of the ~9.1ms total. The graph work this metric is supposed to guard is roughly **1ms of a 9ms measurement**.

**2. The native/wasm gap is systemic across every branch**, so it is not attributable to any commit:

| branch | wasm | native | ratio | gate |
|---|---|---|---|---|
| `fix/issue-2524` | 15.2 | 18.7 | 1.23x | warn |
| `fix/issue-2521` | 9.5 | 15.1 | 1.59x | warn |
| `fix/issue-2512` | 9.9 | 19.8 | 2.00x | warn |
| #2592 (guard only) | — | 17.5 | — | warn |
| #2593 `289b5e27` | — | 15.8 | — | warn |
| #2593 `388e9ae7` | 10.3 | 24.4 | 2.37x | **FAIL** |
| #2593 re-run | — | 24.4 | — | **FAIL** |

wasm consistently lands near the 10.1ms baseline; native consistently runs 1.2–2.4× higher. Since both engines execute the *same* `diffImpactData` code over an equivalent graph, and the measurement is ~90% spawn, the gap lives in spawn cost under whatever process state the preceding native build leaves on a shared runner — not in codegraph's query path. (Loading the addon does not by itself slow spawn: locally, git-spawn median was 8.0ms with the addon loaded vs 8.6ms without, and RSS grew only 3MB since the binary is mmap'd lazily.)

**3. The 10ms floor turns that gap into a coin flip.** `MIN_ABSOLUTE_DELTA = 10` (`tests/benchmarks/regression-guard.test.ts:193`) requires a ≥10ms absolute delta before flagging. Against a 10.1ms baseline with native landing in a 15–25ms band, the delta straddles the floor: 15.8 measured → 5.7ms delta → warn; 24.4 measured → 14.3ms delta → **fail**. On PR #2593 the *same code* landed on both sides — the two commits differ by one doc comment:

```
git diff 289b5e27 388e9ae7 -U0 | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' \
  | grep -vE '^[+-]\s*(\*|/\*\*|\*/|//)'      # → no output
```

## Why re-recording the baseline is the wrong fix

It was my first instinct and it is a trap. A baseline re-recorded from a native CI run captures that run's spawn cost, so the gate keeps measuring runner conditions and keeps flipping. Worse, it would make the gate *permanently blind to the thing it exists to catch*: since graph work is ~1ms of ~9ms, a genuine **2× regression in diff-impact's graph work would move the metric by ~1ms** — an order of magnitude below the 10ms floor. The gate cannot currently fail on a real diff-impact regression at all.

## Suggested fix

1. **Take `git` out of the timed region** so the metric measures graph work. The clean seam is an option on `diffImpactData` accepting a precomputed diff (changed ranges / changed-file list); `benchDiffImpact` then captures the diff once and times only the graph traversal. Useful beyond benchmarking — `src/mcp/tools/diff-impact.ts` could reuse the same seam.
2. **Re-record this metric's baseline once (1) lands**, because it will then measure a different (and much smaller) quantity. Legitimate: new metric, new baseline — as opposed to re-recording today's spawn-dominated number.
3. **Reconsider `MIN_ABSOLUTE_DELTA` for sub-10ms metrics.** A metric whose true value is ~1ms can never exceed a 10ms absolute floor, so it is unguarded no matter the percentage. Either scale the floor per metric or gate small metrics on percentage alone with more samples.
4. **Give `benchDiffImpact` the `WARMUP_RUNS` treatment** `benchDepths` already has (line 124) — currently a bare `timeMedian(…, RUNS)` with `RUNS = 5`. Still worth doing, but note this alone would not have prevented the failure: the gap is systematic, not cold-start.
5. **Consider not failing the aggregate `CI Testing Pipeline`** on this metric until (1)–(3) land, since it currently blocks unrelated PRs and trains reviewers to re-run until green.

## Impact

Today the gate blocks arbitrary PRs (it blocked #2593) while being structurally incapable of catching the regression it was built for.

carlos-alm:
Closes #2598

## Why

`Pre-publish benchmark gate` failed on #2593 with `diffImpact latency: 10.1 → 24.4 (+142%)`. Investigating it turned up two defects in the *measurement*, not in the product — and together they meant the gate could not fail on the regression it exists to catch, while failing on unrelated PRs instead.

**It was not a product regression.** Ruled out point by point:

- No commit in the 32 since `v3.17.0` touches `diff-impact.ts`, `queries.ts`, or `impact.ts`.
- Reproduced locally at baseline: replicating the benchmark exactly gave **native 9.1ms / wasm 9.2ms** against the 10.1ms baseline.
- The engines are not building different graphs: nodes 20812 vs 20812, edges 45499 vs 45789 (native slightly *fewer*), same 67 indexes, no `sqlite_stat1`, no leftover `-wal`.
- Every other native query is *faster* than wasm in the same run (fnDeps 11.4 vs 13.8). Only `diffImpact` inverted.
- The native/wasm split is systemic across unrelated branches — `fix/issue-2512` 9.9/19.8, `fix/issue-2521` 9.5/15.1, `fix/issue-2524` 15.2/18.7 — so it is attributable to no commit.

## The two defects

**1. The `git` subprocess sat inside the timed region.** `benchDiffImpact` timed `diffImpactData(…, { staged: true })`, which shells out via `runGitDiff` → `execFileSync('git', …)`. Measured here, `git diff --cached --unified=0` alone is **~8ms of a ~9ms call** — the metric was ~90% process spawn.

**2. The probe edit landed outside every function.** It appended `// benchmark-probe` after the last line, so the changed range intersected no definition: `findAffectedFunctions` returned empty and the call short-circuited before the BFS/co-change/ownership work. Every recorded run shows `affectedFunctions: 0`.

Net effect: the analysis was **~0.5ms inside a ~9ms measurement**, so a real 2× regression in diff-impact's graph work would move the metric by well under a millisecond — far below the gate's 10ms `MIN_ABSOLUTE_DELTA`.

## What this changes

- `diffImpactData` gains a `diffText` option accepting pre-captured diff output, and `runGitDiff` is exported so a caller can capture exactly the diff the function would have produced itself. `benchDiffImpact` captures once, outside the timed region.
- The probe is inserted at the midpoint of the resolved hub function's **body**, using a line range now returned by `selectHubTargets` — the single source of truth for target selection, whose doc comment already directed probe-writing callers there.
- Both degrade gracefully: an older benchmark source without `runGitDiff`/`diffText` falls back to timing the whole call (this script benchmarks published versions too, via `resolveBenchmarkSource`), and a missing hub line range falls back to appending at EOF. `excludesGitSpawn` is recorded so the two modes are distinguishable rather than reading as a huge improvement.

Stacked on #2591, which landed the `WARMUP_RUNS` fix while this was in progress. I kept that commit's warmup rationale and corrected the one clause it left stale — it credited warming the `.git` data read by the staged-diff subprocess, which no longer runs inside the timed region.

## Verification

```
diffImpact  before: ~9ms measured, affectedFunctions 0   (≈0.5ms of analysis)
            after:  0.6ms wasm / 0.7ms native, affectedFunctions 1, affectedFiles 2
```

The two engines now agree, confirming the 2.4× CI gap was spawn cost rather than a query regression.

- `npx tsc --noEmit` clean; `npx biome check src/ tests/` clean; `npm run build` clean
- `RUN_REGRESSION_GUARD=1 npm run test:regression-guard` — **36/36 pass** against a regenerated report (17 of those are skipped without a `dev` entry, so they had not been exercised locally before)
- `npx vitest run tests/graph tests/integration` — 171 files, 1838 pass / 2 skipped / 2 todo
- Benchmark run emits no fallback warnings, so the probe lands inside the hub body as intended

## Deliberately not in this PR

- **Re-recording the baseline.** The metric now measures a different, much smaller quantity, so 0.7ms vs the recorded 10.1ms reads as an improvement and cannot fail the gate. It should be re-recorded on the next release rather than hand-edited here.
- **Scaling `MIN_ABSOLUTE_DELTA` for sub-10ms metrics.** A ~0.7ms metric can never exceed a 10ms absolute floor, so `diffImpact` remains effectively unguarded even after this fix. That is a policy call on a constant shared with other metrics, and it needs its own decision — see #2598's remaining item.

One caveat worth stating: `scripts/` is excluded from `tsconfig.json`, so these script changes are not covered by `npm run typecheck`. They are validated by running them, which is what the verification above does.

### DIFF
--- a/scripts/lib/hub-selection.ts
+++ b/scripts/lib/hub-selection.ts
@@ -52,11 +52,26 @@ export interface HubTargets {
 	 */
 	mid: string;
 	leaf: string;
+	/**
+	 * 1-based line range of the resolved hub definition, when the graph
+	 * recorded one.
+	 *
+	 * `benchDiffImpact` needs it to place its probe edit *inside* the hub
+	 * function's body. Appending the probe after the last line instead (what it
+	 * did before #2598) puts the changed range outside every function, so
+	 * `diffImpactData` finds zero affected functions and short-circuits before
+	 * the BFS/co-change/ownership work the metric exists to guard — measuring
+	 * the short-circuit rather than the analysis.
+	 */
+	hubLine: number | null;
+	hubEndLine: number | null;
 }
 
 interface NodeRow {
 	name: string;
 	file: string;
+	line: number | null;
+	end_line: number | null;
 }
 
 interface RankedNodeRow extends NodeRow {
@@ -84,10 +99,12 @@ export function selectHubTargets(dbPath: string, pinnedCandidates: readonly stri
 
 		let hub: string | null = null;
 		let hubFile: string | null = null;
+		let hubLine: number | null = null;
+		let hubEndLine: number | null = null;
 		for (const candidate of pinnedCandidates) {
 			const row = db
 				.prepare(
-					`SELECT n.name, n.file FROM nodes n
+					`SELECT n.name, n.file, n.line, n.end_line FROM nodes n
            JOIN edges e ON e.source_id = n.id OR e.target_id = n.id
            WHERE n.name = ? AND n.kind IN (${kindPlaceholders})
              AND n.file NOT LIKE '%test%' AND n.file NOT LIKE '%spec%'
@@ -98,13 +115,15 @@ export function selectHubTargets(dbPath: string, pinnedCandidates: readonly stri
 			if (row) {
 				hub = row.name;
 				hubFile = row.file;
+				hubLine = row.line;
+				hubEndLine = row.end_line;
 				break;
 			}
 		}
 
 		const rows = db
 			.prepare(
-				`SELECT n.id, n.name, n.file, COUNT(e.id) AS cnt
+				`SELECT n.id, n.name, n.file, n.line, n.end_line, COUNT(e.id) AS cnt
          FROM nodes n
          JOIN edges e ON e.source_id = n.id OR e.target_id = n.id
          WHERE n.kind IN (${kindPlaceholders})
@@ -121,12 +140,14 @@ export function selectHubTargets(dbPath: string, pinnedCandidates: readonly stri
 		if (!hub) {
 			hub = rows[0].name;
 			hubFile = rows[0].file;
+			hubLine = rows[0].line;
+			hubEndLine = rows[0].end_line;
 		}
 
 		const mid = rows[Math.floor(rows.length / 2)].name;
 		const leaf = rows[rows.length - 1].name;
 
-		return { hub, hubFile: hubFile as string, mid, leaf };
+		return { hub, hubFile: hubFile as string, mid, leaf, hubLine, hubEndLine };
 	} finally {
 		db.close();
 	}
--- a/scripts/query-benchmark.ts
+++ b/scripts/query-benchmark.ts
@@ -92,9 +92,16 @@ const { srcDir, cleanup } = await resolveBenchmarkSource();
 const dbPath = path.join(root, '.codegraph', 'graph.db');
 
 const { buildGraph } = await import(srcImport(srcDir, 'domain/graph/builder.js'));
-const { fnDepsData, fnImpactData, diffImpactData } = await import(
-	srcImport(srcDir, 'domain/queries.js')
-);
+const queries = await import(srcImport(srcDir, 'domain/queries.js'));
+const { fnDepsData, fnImpactData, diffImpactData } = queries;
+// #2598: `runGitDiff` + `diffImpactData`'s `diffText` option let this
+// benchmark hoist the `git` subprocess out of the timed region, so the metric
+// measures graph work instead of process spawn (spawn was ~8ms of a ~9ms
+// call). Older releases -- which this script also benchmarks, via
+// resolveBenchmarkSource -- lack both, so fall back to timing the whole call
+// there rather than failing. Numbers from the two modes are not comparable;
+// see benchDiffImpact.
+const runGitDiff = typeof queries.runGitDiff === 'function' ? queries.runGitDiff : null;
 // v3.9.5+ parses WASM in a worker_thread that keeps the event loop alive until
 // disposed. Older releases don't export disposeParsers — fall back to a no-op.
 let disposeParsers = async () => {};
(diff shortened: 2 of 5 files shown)


carlos-alm:
**Confirmed on CI — the native/wasm gap disappeared entirely.**

This PR's own benchmark run (all 26 checks green, including `Pre-publish benchmark gate`):

| | wasm | native | gap | affectedFunctions |
|---|---|---|---|---|
| before (#2593's run) | 10.3 | 24.4 | **2.37×** | 0 |
| after (this PR) | **1.7** | **1.7** | **none** | **1** |

That is the proof the investigation was after. The 2.4× native/wasm split — which is what tripped the gate and blocked #2593 — vanished the moment the `git` subprocess left the timed region. It was process spawn cost, not a regression in the query path.

Also worth noting: `affectedFunctions` went 0 → 1, so the metric now actually executes the BFS/co-change/ownership work it exists to guard, instead of timing a short-circuit.

CI measures 1.7ms against 0.6/0.7ms locally (~2.5× slower hardware), which is the expected direction and magnitude.

Caveat unchanged from the description: at 1.7ms this metric still sits far below `MIN_ABSOLUTE_DELTA`'s 10ms floor, so it cannot fail the gate on a real regression. That is the remaining open item on #2598 and needs a policy call on a constant shared with other metrics.

Issue #2647, opened 2026-08-25
issue-2088 plan: for-of enclosing-scope name (funcStack/enclosing_func_context) diverges from findCaller for TS class methods, class-field arrows, and object-literal arrow props


carlos-alm:
## What

While verifying WU-5(a)'s `resolveReceiverSites` (PR #2612, plan for issue #2088, round 28) against the real pipeline, found that the enclosing-scope name the JS/TS extractor writes for a for-of binding (`ForOfBinding.enclosingFunc`, consumed by `buildForOfConstraints` to key its points-to constraint) can diverge from the name `findCaller`/`find_enclosing_caller` derives for a call at the same source position — for the same call, in the same file. Both engines are affected identically, and the divergence is pre-existing (not introduced by #2088).

Two independent root causes, both symmetric across the TS and Rust extractors:

**(a) TS class names parse as `type_identifier`, not `identifier`.** The context-collector that builds the qualified enclosing-scope name only qualifies a class method when the class-name node's grammar kind is `identifier`:
- TS: `src/extractors/javascript.ts`'s `computeClassNameContext` (`classNameIsIdentifier: nameNode?.type === 'identifier'`, ~line 3395) gates `pushClassContext` (~line 3253-3263), consumed by `pushMethodDefContext` (~line 3321-3334).
- Rust: `crates/codegraph-core/src/extractors/javascript.rs`'s `enclosing_func_context` (~line 7196-7210) filters `.filter(|name| name.kind() == "identifier")` — the doc comment on the function (~line 7180-7184) already states this caveat explicitly.

Since a TS class's name node is `type_identifier`, both engines fail to qualify the method, and a for-of directly inside a TS instance/static/getter/async method gets `enclosingFunc`/`enclosing_func_context` = the bare method name (e.g. `run`) instead of `ClassName.run`.

Meanwhile, the SEPARATE mechanism that produces the `Definition`/`Def` entry `findCaller`/`find_enclosing_caller` reads from — TS's `findParentClass`→`findParentNode` (`javascript.ts:6739-6741`), Rust's `find_parent_class`→`find_enclosing_type_name` (`javascript.rs:6684-6686`) — reads the class name node's `.text` regardless of its grammar kind, so it correctly returns `ClassName.run`. The two mechanisms disagree only for TS classes; a JS class (name node `identifier`) gets the same qualified name from both.

[...]

Issue #2645, opened 2026-08-25
issue-2088 fixture matrix: function-shape containers need an explicit invocation mechanism for the runtime oracle


carlos-alm:
Follow-up from #2612 (docs/plans/issue-2088.md), round 25's tightening of the round-17 coverage-check recipe after Greptile's P1 review comment.

## Context

The mechanically-generated fixture matrix planned in WU-10 (docs/plans/issue-2088.md, "Fixture matrix" section) computes a three-term contract for each generated combination: `invoked === true && escapes === 0 && !hasT1Evidence`. `invoked` is obtained by executing the generated file under real Node and checking whether the tested property's handler actually runs.

## The gap

For the round-17 coverage-check recipe (a `var`-kind for-of head decoy, target = content axis's own identifier), executing the generated file with `container` set to any of the nine function-shape values (`function_declaration`, arrow function, IIFE, `async function`, generator function, object method, class method, callback) never causes `invoked` to read `true`, because nothing in the axis definitions ensures the container itself gets called. The container's body — and the decoy/correlated-call machinery inside it — never runs unless something invokes it.

Executed check: the recipe flips `escapes` correctly (0 unfixed, 1 fixed) in all fifteen containers, since that's a purely static computation. But `invoked` only reads `true` for `container = module` and the five block-shaped containers (bare block, `if` block, `try` block, `switch` case, loop body), which execute unconditionally as part of ordinary top-level control flow. The nine function-shape containers need the matrix generator to also emit a call to the container itself (or otherwise guarantee its body executes) before the three-term contract becomes observable for those values.

## Ask

When WU-10 is implemented, the matrix generator should ensure every `container` axis value's body is guaranteed to execute at runtime — e.g. by appending a trailing call/instantiation appropriate to the container kind (`containerName()` for a function declaration, invoking a returned/stored arrow function, `new` for a class, etc.) — so the `invoked` oracle is observable for every combination, not only six of the fifteen container values.

This does not affect the correctness of any existing fix; `escapes` is computed statically and is unaffected by whether the container is ever called. This is purely a completeness gap in the runtime-oracle side of the matrix's own test methodology.

Issue #2643, opened 2026-08-25
functionScopeDeclaresVar cannot see a for (var x of y) loop-head binding


carlos-alm:
Found during the `/oversee` review of [#2088](https://github.com/optave/ops-codegraph-tool/issues/2088)'s plan (PR #2612, round 21).

## What

`functionScopeDeclaresVar` (`src/extractors/javascript.ts:4660-4672`) detects a hoisted `var` binding by looking for a `variable_declaration` node:

```ts
if (node.type === 'variable_declaration' && declarationDeclaresName(node, name)) return true;
```

A `for (var x of y)` / `for (var x in y)` loop head does **not** produce one. Verified against `tree-sitter-javascript@0.25.0`: the `var` token and the `left` identifier sit **directly under `for_in_statement`**, with `kind` carrying `var` — there is no nested `variable_declaration` node anywhere in that subtree for the walk to match.

So a `var`-kind for-of/for-in loop variable is invisible to `functionScopeDeclaresVar`, and therefore to `introducesShadowedBinding`'s function-shape arm, even though `var` is function-scoped and the binding genuinely is hoisted to the enclosing function.

## Why it matters

This is the same shape already fixed **twice** in `#2088`'s own dedicated helpers — round 17 taught both `countHoistedVarScopeDeclarations` and `subtreeContainsReassignmentOf` to test `for_in_statement` with `kind === 'var'` for exactly this reason. The shared primitives never got the equivalent treatment.

**Direction: OVER-escape (recall only) for the consumers that remain.** `#2088`'s own resolution path works around it — the plan's for-of recursion computes its widened boundary directly rather than through `findDeclaringScopeNode`, precisely because of this gap. For the remaining consumers (`blockContainsIdentifierExcluding`, `findDeclaringScopeNode`), a *missed* shadow means over-inclusion in a conjunction, which yields more escapes, not fewer. No live-code-reported-dead path is known through it today.

## Why it wasn't fixed inline

Widening `functionScopeDeclaresVar` or `introducesShadowedBinding` changes behaviour for every verified consumer of those primitives — `findResolvingScopeNode`, `findDeclaringScopeNode`, `blockContainsIdentifierExcluding`, `killsBinding`. Every round of the #2088 review has declined to touch them for that reason, fixing locally in condition 4's own helpers instead. Doing it here would need its own review of each consumer.

## Suggested fix shape (not binding)

[...]

Issue #2638, opened 2026-08-24
issue-2088 plan: a get-flavoured method_definition can smuggle a this-using function through its return value


carlos-alm:
## What

Found in round 18 of the plan (PR #2612, `docs/plans/issue-2088.md`).

`literalHasUnmodeledThisReference`'s `method_definition` arm proved safety only via `subtreeContainsThisKeyword` — sound for an ordinary method, not for a GETTER. A getter's own body can be entirely free of the `this` keyword while still RETURNING a value that, once the property is accessed and the result called (`T.k()`), gets `this` bound to the receiver — per ordinary member-expression-callee semantics, independent of anything the getter's own body does:

```js
function fnA(){ return 41; }
function runImpl(){ return this.alpha(); }
const T = { alpha: fnA, get run(){ return runImpl; } };
T.run();            // → 41: fnA invoked via this.alpha(), this === T
```

Closed (round 18) by treating a `get`-flavoured `method_definition` as escaping unconditionally, regardless of what its own body contains. A plain method and a setter are unaffected — a setter's return value is never called (assignment expressions evaluate to the assigned value, never to what the setter returns), so `subtreeContainsThisKeyword` remains sound for it.

## Direction

`OVER-escape`. This is a new, deliberate narrowing, not a safety gap: an entirely harmless getter (`get value(){ return 42; }`) now also makes its literal escape, because resolving what a getter returns — to any bounded number of hops — would still be incomplete (a returned call expression, inline function expression, or further property read could each independently be `this`-using, unlike the identifier-valued `pair`/`shorthand` arms' own single, positively-enumerable resolution target). This costs recall only, matching the same "detect the shape, fail safe outright" trade-off round 9's `spread_element` exclusion and round 10's/round 13's shadow/reassignment exclusions already accept elsewhere in this same design.

## Suggested next step

Not required for issue #2088 to ship soundly (the exclusion is disclosed and conservative). If pursued: resolve a getter's own `return` expression the same way `resolveIdentifierValueThisReference` resolves an identifier-valued `pair`, when the getter's body is exactly `return <identifier>;` — but this still leaves every other return-expression shape (call, inline function, member access) unresolved, so the getter arm would remain partially unconditional regardless. Would need its own focused round and its own review, per this plan's own precedent for #2625/#2627/#2631.

## Where

- `src/extractors/javascript.ts` (`literalHasUnmodeledThisReference`'s `method_definition` arm — not yet implemented; this is a plan-doc-only follow-up)
[...]

PR #2654, opened 2026-09-03
feat(roles): receiver-correlated invoked-property evidence for object-literal value-refs

### LINKED ISSUE #2088: follow-up: receiver-type-aware invoked-property matching to reduce dead-code false negatives
Deferred from PR #2034 review.

**Original reviewer comment:** inline comment — https://github.com/optave/ops-codegraph-tool/pull/2034#discussion_r3543999310

**Context:**

`collectInvokedPropertyNames` (`src/domain/graph/builder/call-resolver.ts`, mirrored as `collect_invoked_property_names` in `crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs`) adds a property name to its evidence set whenever ANY call in the processed files has a non-empty `receiver` and that `name` — with no check that the receiver is plausibly the same dispatch-table object referenced by the object-literal being evaluated. A completely unrelated `logger.resolve()` or `promise.resolve()` call anywhere in the codebase will add `resolve` to the set and prevent `{ resolve: neverCalled }` from being flagged dead, even if the actual dispatch table is never read.

This is the conservative error direction (false negative for dead-code — it can never misclassify genuinely live code as dead), so it was accepted as-is for #2034: that PR's stated scope was "require *any* invocation evidence" as a deliberately coarse "one hop further" heuristic (see the doc comment on `collectInvokedPropertyNames`), not precise receiver-type correlation. Tightening it now would require:

- Extending the existing points-to/type-map machinery to track each receiver call site's inferred type/binding.
- Correlating that inferred type back to the specific object-literal (or dispatch-table variable) that defines the property being checked.
- Implementing this identically in both the WASM/TS and native Rust engines, per this repo's dual-engine parity requirement.

That's a nontrivial dataflow precision improvement, not a small tweak, so it's being tracked separately rather than folded into #2034's narrower "any evidence at all" fix.

**What the fix entails:** teach `collectInvokedPropertyNames`/`resolveFallbackTargets` (and their Rust mirrors) to only count a `x.name(...)` call as evidence for a given object-literal's `{ name: fn }` property when `x`'s resolved type/binding plausibly refers to that same object (e.g. via the existing `typeMap`/points-to infrastructure), rather than matching on bare property name alone.

carlos-alm:
Closes #2088
Part of #2612

## Summary

Receiver-correlated invoked-property evidence for object-literal `value-ref`s. A property `{ resolve: neverCalled }` is no longer kept live by an unrelated `x.resolve(...)` elsewhere in the build, **unless** the owning object literal escapes what the solver models.

This is the object-literal slice of ROADMAP §8.3 allocation-site abstraction. No new subsystem: new constraint rows in the existing Andersen solver (ADR-002).

## Success criteria

- Local-closed object literals use T1 exclusively (`site|key` correlated evidence). Escaping sites keep today's T2 bare-name predicate. T3 (`computedDispatchTableEvidence`) stays name-keyed and is ORed in.
- Escape analysis fail-safe is `escapes: true`. Parameter-passing, `this`-using methods, spreads, exports, factory returns, and array-container `.forEach` all fall back to T2.
- `analysis.correlatedPropertyEvidence` is the only new `DEFAULTS` key. Setting it `false` restores pre-#2088 T2 (verified: decoy fixture keeps `neverCalled` live).
- DB migration v32 adds `object_literal_sites` and `invoked_property_sites`, persisted and purged per-file like `invoked_property_names`.
- Dual-engine: TS WASM extractor/solver/edges + Rust mirrors + NAPI `objectLiteralSites` / `extraInvokedPropertySites` / `correlationEnabled`.
- WASM worker protocol threads top-level `ExtractorOutput.objectLiteralSites`.
- Nine existing regression tests pass unedited (`issue-1895`, `issue-2260`, `issue-2087`).
- `pts-javascript` fixture `objlit-site.js` is 100%/100%; `javascript` precision 1.0 floor still holds.

## Soundness

This change must never convert a conservative false-negative (live property not flagged dead) into a false-positive (live property flagged dead). Every incomplete path sets `escapes: true` and uses T2.

## Dogfood measurement

`codegraph roles --role dead -T` on this repo: **241** dead symbols with the feature on, **241** with `analysis.correlatedPropertyEvidence: false`, wasm ≡ native (byte-identical 241-symbol set). This repo's own dispatch tables are not local-closed, so the recall win is in the WU-10 fixtures rather than the dogfood count.

## Tests

- `tests/integration/issue-2088-correlated-property-evidence.test.ts` — decoy, handler-array/for-of, alias, two-file pass-ordering, function/block-scoped tables, mixed data/handler, TS wrappers (paren/`as const`/`satisfies`/`!`), class method / class-field arrow / object-literal arrow (#2647), collision cost (shape 36), config-off restore, producer-only watch rebuild.
- `tests/integration/issue-2088-escape-fallback.test.ts` — exported, param-flow, `this.k()`, spread, `.forEach`, factory return.
- Parser extraction tests for site id, exported escape, `this`/spread escape, mixed-data non-escape.
- Nightly matrix `tests/integration/issue-2088-matrix.test.ts` is intentionally not in this PR (plan: not per-PR).

The plan enumerates 36 correlation shapes and 86 escape-fallback cases accumulated across planning rounds. This PR lands the headline, scope-trio, B1/B2, incremental, config-off, and soundness-gate cases under both engines. Remaining lettered round-N guard snippets are the same mechanisms those cases already exercise.

## Out of scope (already filed, not implemented, not closed)

#2610, #2611, #2617–#2625, #2627, #2629–#2636, #2638–#2643, #2645–#2647.

## Verification

- `npx tsc --noEmit`
- `npx biome check` on the PR files
- `cargo test -p codegraph-core` (1135 passed)
- `cargo clippy -p codegraph-core --all-targets -- -D warnings`
- `npm run build`
- `npm run doctor` (healthy; optional WASM grammar warn)
- Plan-named vitest files, including unedited #1895/#2260/#2087
- `pts-javascript` 100% P/R; `javascript` precision floor
- `codegraph diff-impact --staged -T`
- `node dist/cli.js build --engine wasm` then `--engine native` on the same DB (plan-required order)

`npm test` (full suite) fails in this worktree on optional-language parser/parity tests because 12 optional WASM grammars are missing (`npm run doctor` already warns). Not a #2088 regression.

`npm run benchmark` was not re-run here; no solver-iteration change (`MAX_SOLVER_ITERATIONS` / `buildCallSiteTypeMap` untouched).


src/domain/graph/builder/stages/build-edges.ts
```diff
@@ -2941,6 +3147,7 @@ function buildCallEdgesPhase(
   // #2087: persist once per pass regardless of which sub-path below resolves
   // this build's edges — see persistInvokedPropertyNames doc comment.
   persistInvokedPropertyNames(ctx);
+  persistObjectLiteralSites(ctx);
   // #2138: same rationale, for cross-file return-type propagation.
   persistReturnTypes(ctx);
```
carlos-alm, line 3183: **Blocking — persist gap.** `buildCallEdgesPhase` always persists names / object-literal sites / return types, but `persistInvokedPropertySites` is only called from `buildCallEdgesJS` (line 1600). Full builds with a native addon take `useNativeCallEdges` below and skip it.

Success Criteria: `invoked_property_sites` must be persisted per-file like `invoked_property_names`. Move the persist (and the correlated-key collection it needs) here so it runs on both sub-paths. Add a test that reads the table after a full native / native-call-edges build *before* `rebuildFile`.

crates/codegraph-core/src/domain/graph/builder/pipeline.rs
```diff
@@ -2185,6 +2190,8 @@ fn build_and_insert_call_edges(
     // view for repos built with the native engine too.
     import_edges::persist_invoked_property_names(conn, &file_entries)
         .map_err(|e| format!("invoked property name persistence failed: {e}"))?;
+    import_edges::persist_object_literal_sites(conn, &file_entries)
+        .map_err(|e| format!("object literal site persistence failed: {e}"))?;
```
carlos-alm, line 2194: **Blocking — WU-8 incomplete.** This persists `invoked_property_names` and `object_literal_sites`, then SELECTs `invoked_property_sites` as extras (lines 2208-2214) — but `import_edges.rs` has no `persist_invoked_property_sites`, so the extra-SELECT is always empty on the native orchestrator path.

Add `persist_invoked_property_sites` next to `persist_object_literal_sites` and call it after `collect_invoked_property_sites` has this pass's keys.

src/extractors/javascript.ts
```diff
@@ -4554,21 +4563,728 @@ function findEnclosingTableName(node: TreeSitterNode): string | undefined {
...
+    if (node.id !== scope.id && scopeShadowsName(node, bindingName)) return;
+
+    if (
+      (node.type === 'identifier' || node.type === 'shorthand_property_identifier') &&
+      node.text === bindingName &&
+      !isBindingOccurrence(node) &&
+      node.id !== skipNode?.id
+    ) {
+      refs.push(node);
+    }
```
carlos-alm, line 5173: **Blocking — live code as dead (B5 / #2640).** This walk only matches `identifier` / `shorthand_property_identifier`. `globalThis.T` is a `property_identifier`; `globalThis['T']` is a string index. `isGlobalObjectQualifiedWrite` exists at line 4920 but is only used from the write scan.

Executed on this PR's extractor: `var T = { resolve: liveFn }; globalThis.T.resolve();` and `f(globalThis.T)` both emit `escapes: false`. T1 then exclusive-misses `resolve` (receiver text is `globalThis.T`, not a pts key) and reports `liveFn` dead.

Reuse `isGlobalObjectQualifiedWrite` verbatim as a third matching case, always untracked. Mirror in `crates/codegraph-core/src/extractors/javascript.rs` (~5292). Land WU-10 case (bh) under both engines asserting `escapes === 1` and `liveFn` live. Fail-safe is `escapes: true`.

crates/codegraph-core/src/extractors/javascript.rs
```diff
@@ -4450,10 +4451,949 @@ fn find_enclosing_table_name(node: &Node, source: &[u8]) -> Option<String> {
...
+        if node.id() != scope_id && scope_shadows_name(&node, binding_name, source) {
+            return;
+        }
+        if (node.kind() == "identifier" || node.kind() == "shorthand_property_identifier")
+            && node_text(&node, source) == binding_name
+            && !is_binding_occurrence(&node, source)
+            && skip_node != Some(node.id())
+        {
+            refs.push(node);
+        }
```
carlos-alm, line 5332: **Blocking — Rust mirror of B5.** Same hole as TS `allReferencesTracked`: no `is_global_object_qualified_write` case, so a script-scope `var T` read via `globalThis.T` / `window.T` / `global.T` / `self.T` is treated as having no references (`escapes` stays false). Dual-engine: this must change with the TS walk, not be documented as a parity gap.

carlos-alm (commented):
## Request changes — PR #2654 vs issue-2088 Success Criteria

Independent verification against `docs/plans/issue-2088.md` Success Criteria + Verification Commands. I did not push fixes.

### Verification actually run

| Command | Result |
|---|---|
| `cargo fmt -- --check` | **FAIL** (same as CI "Rust compile check") |
| `npx tsc --noEmit` | pass |
| `npx biome check` on PR TS files | pass |
| `cargo test -p codegraph-core` | 1135 passed |
| `cargo clippy -p codegraph-core --all-targets -- -D warnings` | pass |
| Plan-named vitest files | WASM #2088 + #1895/#2260/#2087 + parser + config: pass. Native #2088: failed in this worktree because `loadNative()` used the published `@optave/codegraph-win32-x64-msvc` addon, not a just-built napi binary from this PR. CI "Engine parity (ubuntu-latest)" already passed, so I am **not** treating that native-suite failure as a source bug. |
| `npm test` / `npm run benchmark` / dogfood `roles --role dead` wasm→native | not completed here (`npm install` prepare was killed mid-WASM build; `dist/` was produced by vitest globalSetup). Reported, not silently skipped. |

---

### Blocking 1 — `cargo fmt` is red (CI already failed)

`cargo fmt -- --check` diffs `crates/codegraph-core/src/domain/graph/builder/stages/build_edges.rs` (every `build_call_edges(..., None, None, None)` call site plus the `pts.entry(...)` seed) and `crates/codegraph-core/src/extractors/javascript.rs` (new helpers). CI job "Rust compile check" failed on `cargo fmt -- --check` for the same files.

**Must change:** run `cargo fmt` on those two files and push. Do not merge while that check is red.

---

### Blocking 2 — `invoked_property_sites` is never persisted on the production paths

Success Criteria: *"DB migration v32 adds `object_literal_sites` and `invoked_property_sites`, **persisted and purged per-file like `invoked_property_names`**"* and *"WASM and native engines produce identical `object_literal_sites`, identical `invoked_property_sites`"*. WU-8 Implementation: *"`import_edges.rs` — `persist_invoked_property_sites` + `persist_object_literal_sites` beside the existing `persist_invoked_property_names`"*.

What shipped:

- `src/domain/graph/builder/stages/build-edges.ts:3147-3152` (`buildCallEdgesPhase`) always persists names, object-literal sites, and return types — **not** invoked-property sites.
- `persistInvokedPropertySites` is only invoked from `buildCallEdgesJS` at `src/domain/graph/builder/stages/build-edges.ts:1600`. Full builds with a native addon take `useNativeCallEdges` (`:3156-3160`) and **skip** that function.
- `crates/codegraph-core/src/domain/graph/builder/stages/import_edges.rs` has `persist_invoked_property_names` (`:319`) and `persist_object_literal_sites` (`:359`) and **no** `persist_invoked_property_sites`.
- `crates/codegraph-core/src/domain/graph/builder/pipeline.rs:2191-2222` persists names + sites, then `SELECT`s `invoked_property_sites` as `extra_invoked_property_sites`, then calls `build_call_edges`. Nothing ever `INSERT`s into that table on the native orchestrator path.

`object_literal_sites` persistence is in place on both engines. `invoked_property_sites` is schema + purge + read-side only. After a native (or JS-orchestrator + native call-edges) full build the table is empty; a later `codegraph watch` / `rebuildFile` cannot recover correlated evidence from untouched files. The incremental test does not catch this: it rebuilds the producer and then asserts `COUNT(*) > 0` on rows **that `rebuildFile` itself just wrote**.

**Must change:**

1. Add `persist_invoked_property_sites` in `import_edges.rs` next to `persist_object_literal_sites`, keyed per `file` like `persist_invoked_property_names`. Call it from `pipeline.rs` **after** `collect_invoked_property_sites` has keys (this pass's rows), so the extra-SELECT on the next incremental pass is not vacuously empty.
2. Call `persistInvokedPropertySites` from `buildCallEdgesPhase` for **both** the native-call-edges and JS branches (same "once per pass regardless of sub-path" contract as `persistInvokedPropertyNames` at `:3147-3148`). That means computing the correlated keys in JS even when `buildCallEdgesNative` resolves edges, or returning those keys from the napi call and persisting them in JS.
3. Add a test that queries `invoked_property_sites` **after a full native build and after a full wasm build with native call-edges**, *before* any `rebuildFile`. It must fail if either path leaves the table empty while `object_literal_sites` is populated.

---

### Blocking 3 — `allReferencesTracked` fail-opens on `globalThis`/`window`/`global`/`self` qualified reads (live code as dead)

Success Criteria round 20 B5 / #2640: *"`allReferencesTracked`'s candidate-matching walk never treats a `globalThis`-qualified read of a script-scope binding as if no reference exists at all … enforced by a third, structural node-matching case reusing `isGlobalObjectQualifiedWrite` verbatim, always classified untracked."* Hard invariant: this change must **never** report live code as dead; fail-safe is `escapes: true`.

What shipped: `isGlobalObjectQualifiedWrite` exists at `src/extractors/javascript.ts:4920-4944` (and `is_global_object_qualified_write` in `crates/codegraph-core/src/extractors/javascript.rs:5027`) but is only consulted from `subtreeContainsReassignmentOf` / the write scan. The `allReferencesTracked` walk at `src/extractors/javascript.ts:5157-5164` (Rust mirror `javascript.rs:5292-5298`) only pushes `identifier` / `shorthand_property_identifier` nodes. `globalThis.T` is a `property_identifier`; `globalThis['T']` is a string index. Both are invisible to the walk.

Executed against this PR's extractor (`createParsers` + `extractSymbols`):

```js
function liveFn() { return 1; }
var T = { resolve: liveFn };
globalThis.T.resolve();
// objectLiteralSites: [{ owner: 'T', escapes: false }]  // must be true

function liveFn() { return 1; }
var T = { resolve: liveFn };
f(globalThis.T);
// objectLiteralSites: [{ owner: 'T', escapes: false }]  // must be true
```

With `escapes === false`, T1 is exclusive. `collectInvokedPropertySites` resolves `call.receiver` via pts; `extractReceiverName` for `globalThis.T.resolve()` returns the text `"globalThis.T"`, which is not a pts key, so `resolve` is not credited. `liveFn` is classified dead while it is live. Same hole for `window`/`global`/`self` (`GLOBAL_OBJECT_NAMES`).

This is the forbidden error direction. The PR body's "remaining lettered round-N guard snippets are the same mechanisms" does not apply: the third node-matching case is **absent**, not merely unfixtured.

**Must change (both engines, identical):**

1. In `allReferencesTracked` / `all_references_tracked`, add the plan's third matching case: if `isGlobalObjectQualifiedWrite(node, bindingName)` (reuse verbatim, including the subscript + `unwrapParens` arms), treat it as a reference that is **not** a tracked position (`return false` / `escapes = true`).
2. Land WU-10 escape-fallback case (bh) (classic-script `globalThis.T.resolve()` / `f(globalThis.T)`, no `"type": "module"`, no `"use strict"`) under **both** engines, asserting `escapes === 1` **and** `liveFn` not dead. A test that only checks `isDead` under T2 will not catch this.
3. Do not document the hole as a "classic-script limitation."

---

### Not blocking, but do not paper over

- Nightly matrix `issue-2088-matrix.test.ts` is correctly out of this PR per WU-10. That does **not** license dropping B5 or persist.
- `analysis.correlatedPropertyEvidence` is the only new `DEFAULTS` key and is wired through `pipeline.ts:79`, `build-edges.ts:1607`, and Rust `AnalysisConfig`. Config discipline is fine.
- No new runtime dependency, no `LANGUAGE_REGISTRY` change, no dual-engine "parity gap" comments framing wrong output as expected.
- ROADMAP §8.3 records the 241/241 dogfood number.

Sweep agent: fix 1–3 in this PR. Do not add unrelated drive-by work.

src/domain/graph/resolver/points-to.ts
```diff
@@ -449,13 +478,51 @@ export function buildPointsToMap(
...
+  relPath: string,
+  objectLiteralSites?: readonly ObjectLiteralSite[],
+  callAssignments?: readonly CallAssignment[],
+): void {
+  if (!objectLiteralSites?.length) return;
+
+  for (const { site, owner } of objectLiteralSites) {
+    const siteKey = objectLiteralSiteKey(relPath, site);
+    pts.set(siteKey, new Set([siteKey]));
+    if (owner) constraints.push({ lhs: owner, rhsKey: siteKey });
```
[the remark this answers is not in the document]
carlos-alm: Not changing pts owner keys.

This is T1 over-credit (false-negative for dead), not live-as-dead. The approved plan's round-28 collision write-up already bounds that family by T2's admit set, and WU-10 shape 36 fixtures the over-credited-but-still-live outcome. The plan did not specify scoped owner keys; Success Criteria round 28 records this as an accepted cost of the B1-completion probe set, not a defect to redesign the lattice for.

src/domain/graph/builder/stages/build-edges.ts
carlos-alm: Fixed. `persistInvokedPropertySites` now runs from `buildCallEdgesPhase` for both the native-call-edges and JS sub-paths (same once-per-pass contract as `persistInvokedPropertyNames`). A WU-8 test reads `invoked_property_sites` after a full build and before any `rebuildFile`, under both engines.

crates/codegraph-core/src/domain/graph/builder/pipeline.rs
carlos-alm: Fixed. `import_edges.rs` now has `persist_invoked_property_sites` next to `persist_object_literal_sites`. `pipeline.rs` collects this pass's keys via `collect_invoked_property_sites_by_file` and persists them before the extra-SELECT, so a later incremental pass is not vacuously empty.

src/extractors/javascript.ts
carlos-alm: Fixed. `allReferencesTracked`'s walk now treats `isGlobalObjectQualifiedWrite(node, bindingName)` as a third matching case, always untracked (`escapes: true`). WU-10 case (bh) lands under both engines: classic-script `globalThis.T.resolve()` / `take(globalThis.T)`, asserting `escapes === 1` and `liveFn` not dead.

crates/codegraph-core/src/extractors/javascript.rs
carlos-alm: Fixed in lockstep with the TS walk. `all_references_tracked` now matches `is_global_object_qualified_write` as an unconditionally untracked reference. Same WU-10 (bh) fixture runs under the native engine.

carlos-alm:
Intentional — this PR must not close #2647. #2647 is the independently-scoped extractor `enclosingFunc` / context-collector fix (TS class `type_identifier`, class-field arrows, object-literal arrow props). #2088 only tolerates that divergence via `candidateScopesFor`; the PR body lists #2647 under "Out of scope (already filed, not implemented, not closed)."

carlos-alm:
Addressed the three blocking findings from the verification review:

1. `cargo fmt` — landed on `build_edges.rs` and `javascript.rs`.
2. `invoked_property_sites` persist — `persistInvokedPropertySites` now runs from `buildCallEdgesPhase` for both sub-paths; Rust `persist_invoked_property_sites` runs from the native orchestrator after this pass's keys are collected. Full-build test reads the table before `rebuildFile`.
3. B5 / #2640 — both engines treat `isGlobalObjectQualifiedWrite` as an untracked reference in `allReferencesTracked`. WU-10 (bh) asserts `escapes === 1` and `liveFn` live.

Greptile P1 (bare owner-key collision) is T1 over-credit bounded by T2; the plan's collision write-up + shape 36 already accepted that family. Not redesigning pts keys.

#2647 is intentionally not closed.

carlos-alm:
@greptileai

